PR #40, opened 2020-12-18
fix: Send the correct timestamp to CloudWatch when uploading logs.

nikkhilmuthye:
**Issue #, if available:**

**Description of changes:**
If the log message is in JSON format, then use the timestamp from the the JSON object.
If the log message is in TEXT format, then get the timestamp using a regex. The timestamp should be at the very beginning of the log line.

Currently we are sending the timestamp of the CW log events as the time of the upload which is different from the logs timestamp.

Added unit tests.

**Checklist:**
 - [X] Updated or added new unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
+++ b/src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
@@ -31,6 +31,9 @@ import java.io.RandomAccessFile;
 import java.nio.charset.StandardCharsets;
 import java.text.SimpleDateFormat;
 import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.time.temporal.TemporalAccessor;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.Locale;
@@ -41,6 +44,8 @@ import java.util.StringJoiner;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.inject.Inject;
 
 import static com.aws.greengrass.deployment.converter.DeploymentDocumentConverter.LOCAL_DEPLOYMENT_GROUP_NAME;
@@ -61,6 +66,7 @@ public class CloudWatchAttemptLogsProcessor {
     private final Kernel kernel;
     private final DeviceConfiguration deviceConfiguration;
     private static final Logger logger = LogManager.getLogger(CloudWatchAttemptLogsProcessor.class);
+    private static Pattern textTimestampPattern = Pattern.compile("([\\w-:.+]+)");
 
     /**
      * Constructor.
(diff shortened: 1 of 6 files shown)


hui-yang (approved):

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -314,7 +328,7 @@ private boolean checkAndAddNewLogEvent(AtomicInteger totalBytesRead,
         if (currentLogLevel.toInt() < desiredLogLevel.toInt()) {
             return false;
         }
-        return addNewLogEvent(totalBytesRead, attemptLogInformation, data.toString());
+        return addNewLogEvent(totalBytesRead, attemptLogInformation, data.toString(), logMessage.getTimestamp());
```
MikeDombo, line 331: What if the timestamp isn't present? Is that possible?

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -61,6 +64,7 @@
     private final Kernel kernel;
     private final DeviceConfiguration deviceConfiguration;
     private static final Logger logger = LogManager.getLogger(CloudWatchAttemptLogsProcessor.class);
+    private static Pattern textTimestampPattern = Pattern.compile("([\\w-:.]+)Z");
```
jbutler, line 67: `Instant.parse()` expects ISO 8601, right? Rename this variable to convey that. Also, this regex doesn't support time zone offsets. Please add more test cases for this, I think we're missing a lot here

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -246,7 +250,17 @@ private boolean processLogLine(AtomicInteger totalBytesRead,
...
+            Instant logTimestamp = Instant.now();
+            if (matcher.find()) {
+                try {
+                    logTimestamp = Instant.parse(matcher.group());
+                } catch (DateTimeParseException e) {
+                    logger.atTrace().cause(e).log("Unable to parse timestamp for log line");
+                }
+            }
+            reachedMaxSize = addNewLogEvent(totalBytesRead, attemptLogInformation, data.toString(),
+                    logTimestamp.toEpochMilli());
```
jbutler, line 263: Nothing is logged if no timestamp is found. Is this expected? If we do find one but parsing fails, is trace really appropriate?

nikkhilmuthye: I think not logging is fine. It is not that important. If the parsing fails I kept it at trace since the log fill will be full of exceptions since it will log for all log lines. It is too noisy. Also, I feel it is not that important to log this error at error level.

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
nikkhilmuthye: It should not be possible since we don't set that the timestamp field is optional. So if it is not there, the deserialization should fail.

MikeDombo (approved):

hui-yang (approved):

src/test/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessorTest.java
```diff
@@ -221,17 +250,26 @@ public void GIVEN_one_component_one_file_less_than_max_WHEN_locate_throws_Servic
         assertTrue(attempt.getLogStreamsToLogEventsMap().containsKey(logStream));
         CloudWatchAttemptLogInformation logEventsForStream1 = attempt.getLogStreamsToLogEventsMap().get(logStream);
         assertNotNull(logEventsForStream1.getLogEvents());
-        assertEquals(7, logEventsForStream1.getLogEvents().size());
+        assertEquals(13, logEventsForStream1.getLogEvents().size());
         assertTrue(logEventsForStream1.getAttemptLogFileInformationMap().containsKey(file1.getAbsolutePath()));
         assertEquals(0, logEventsForStream1.getAttemptLogFileInformationMap().get(file1.getAbsolutePath()).getStartPosition());
-        assertEquals(13, logEventsForStream1.getAttemptLogFileInformationMap().get(file1.getAbsolutePath()).getBytesRead());
+        assertEquals(2943, logEventsForStream1.getAttemptLogFileInformationMap().get(file1.getAbsolutePath()).getBytesRead());
```
jbutler, line 256: Nit: consider refactoring these test cases so you don't need to make all these edits every time a test file changes

jbutler (approved):

PR #46, opened 2021-02-09
fix: initialize configuration in constructor

hui-yang:
**Issue #, if available:**

**Description of changes:**

**Checklist:**


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -137,9 +137,11 @@
             }
             processConfiguration(configTopicsPojo);
         });
+        // Process initial configuration
+        processConfiguration(configTopics.toPOJO());
```
MikeDombo, line 141: calling `subscribe` already does this. Have a look at the implementation of subscribe, it calls the callback synchronously the first time.

PR #54, opened 2021-04-07
fix(deps): update dependencies so that we don't duplicate dependencie…

MikeDombo:
…s in the jar

**Issue #, if available:**

**Description of changes:**
Decrease jar size from 10MB to 1MB.

**Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/logmanager/LogManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/logmanager/LogManagerTest.java
@@ -29,6 +29,7 @@ import org.mockito.Captor;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import org.slf4j.event.Level;
+import software.amazon.awssdk.crt.CrtRuntimeException;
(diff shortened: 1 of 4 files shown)


src/integrationtests/java/com/aws/greengrass/logmanager/LogManagerTest.java
```diff
@@ -136,6 +137,7 @@ void beforeEach(ExtensionContext context) {
         ignoreExceptionOfType(context, TLSAuthException.class);
         ignoreExceptionOfType(context, InterruptedException.class);
         ignoreExceptionOfType(context, DateTimeParseException.class);
+        ignoreExceptionOfType(context, CrtRuntimeException.class);
```
k-khatri, line 140: Why have we started ignoring a new exception in a test case?

MikeDombo: It wasn't happening before; I don't know why. It could be a new exception due to the new SDK version in the nucleus. But this exception doesn't matter, it is only complaining that it isn't configured with certificates, which is fine.

k-khatri (approved):

hui-yang (approved):

PR #56, opened 2021-04-14
ci: separate comments to allow for external PRs

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Separating commenting steps so that external PRs will trigger the testing workflow.
Also fixing the integration tests so that they run separate to the unit tests and generate their own jacoco report.

**Checklist:**


pom.xml
```diff
@@ -379,12 +379,12 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.88</minimum>
+                                            <minimum>0.71</minimum>
```
philcali, line 382: oh no!

pom.xml
```diff
@@ -379,12 +379,12 @@
...
                                         </limit>
                                         <limit>
                                             <counter>BRANCH</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.83</minimum>
+                                            <minimum>0.59</minimum>
```
philcali, line 387: Why was this lowered so much?

nikkhilmuthye (approved):

pom.xml
MikeDombo: It was never checking and enforcing it before. This coverage number is specific to the integ test; it doesn't include what is covered by unit tests.

nikkhilmuthye: I think the integration tests where in the wrong namespace which was why the coverage was high since we have unit tests also in the same namespace.

MikeDombo: _This_ coverage was never evaluated. The unit test coverage was including coverage from the integ tests.

philcali (approved):

PR #62, opened 2021-06-08
fix: respect 24 hour time span limits for CW upload

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Cloudwatch requires that any batch upload does not exceed a time span of 24 hours. This change splits out each 24 hour batch in order to respect that requirement.

**Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
@@ -50,6 +50,7 @@ import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.TimeZone;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
@@ -70,6 +71,7 @@ import static org.mockito.Mockito.atLeastOnce;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+@SuppressWarnings("PMD.UnsynchronizedStaticFormatter")
 @ExtendWith({GGExtension.class, MockitoExtension.class})
 class LogManagerTest extends BaseITCase {
     private static final String THING_NAME = "ThingName";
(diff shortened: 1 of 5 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -81,6 +83,10 @@
     private Path tempDirectoryPath;
     private final static ObjectMapper YAML_OBJECT_MAPPER = new ObjectMapper(new YAMLFactory());
 
+    static {
+        DATE_FORMATTER.setTimeZone(TimeZone.getTimeZone("UTC"));
```
philcali, line 87: These aren't thread safe. Not sure if that matters in this plugin:

> Date formats are not synchronized. It is recommended to create separate format instances for each thread. If multiple threads access a format concurrently, it must be synchronized externally.

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -68,6 +71,9 @@
     @Inject
     public CloudWatchAttemptLogsProcessor(DeviceConfiguration deviceConfiguration) {
         this.deviceConfiguration = deviceConfiguration;
+        // Set timezone of the formatter to UTC since we always pass in UTC dates. We don't want it to think
+        // that it should use the local timezone.
+        dateFormatter.setTimeZone(TimeZone.getTimeZone("UTC"));
```
philcali, line 76: Same with these,

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
MikeDombo: Yeah, it would matter, but this is a test, so it doesn't matter

philcali (approved):

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
MikeDombo: Fixed with a threadlocal

philcali (approved):

nikkhilmuthye (approved):

Issue #65, opened 2021-08-10
(aws.greengrass.LogManager): allow for multiple  log manager component configuration deployed to the same device to be merged


massi-ang:
**Feature Description**
Defines the `componentLogsConfiguration` field inside `logsUploaderConfiguration` as a map instead of an array, allowing for `componentLogsConfiguration` to be merged across deployments

**Use Case**
A device can be targeted by multiple independent deployments. Each deployment might add new components and user wish to configure aws.greengrass.LogManager for such components. Currently, the `componentLogsConfiguration` is specified as an array, which is not merged, meaning that the last deployment applied to the device will determine the effective component configuration.

**Proposed Solution**
1. Change the `componentLogsConfiguration` to a map of configuration objects
2. `componentName` is now specified as the key in the map
3. removed `componentName` from the allowed fields in the component log configuration object

**Other**
Add detailed explanation, stacktraces, related issues, links for us to have context, etc

* [x] :wave: I may be able to implement this feature request
* [x] :warning: This feature might incur a breaking change

MikeDombo:
Merged in 2.2.0

PR #66, opened 2021-08-10
Change `componentLogsConfiguration` type from array to map of configuration objects

massi-ang:
**Issue #, if available:**
#65 

**Description of changes:**
Changed the `componentLogsConfiguration` object to a map instead of an array
`componentName` is now specified as the key in the map and has been removed from the configuration object fields

To allow for `componentLogsConfiguration` to be merged

Existing tests have been used, with their setup code modified to adapt to the new configuration

This is a breaking change. LogManager configurations from previous versions are not applicable to this. 
No specific warning has been implemented. If the wrong configuration is specified, eg an array instead of a map, it will be ignored.

**Checklist:**
 - [x] Updated or added new end-to-end tests


leaf94:
Thank you very much @massi-ang for raising the PR to allow merge/reset logging configuration per component. This is super sweet!

massi-ang:
@leaf94 Hi Ethan, I see the PR is blocked. Is there something I need to do on my side to make it go through?

MikeDombo:
If you could make it backward compatible @massi-ang then we can merge it in. Otherwise we will need to update it to be compatible which will take a while since we're working on higher priority issues right now.

MikeDombo (approved):
LGTM!
Thank you very much for this contribution

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -144,7 +144,17 @@
     private synchronized void processConfiguration(Map<String, Object> configTopicsPojo) {
         Map<String, ComponentLogConfiguration> newComponentLogConfigurations = new ConcurrentHashMap<>();
         configTopicsPojo.computeIfPresent(COMPONENT_LOGS_CONFIG_TOPIC_NAME, (s, o) -> {
-            if (o instanceof ArrayList) {
+            if (o instanceof Map) {
```
MikeDombo, line 147: Just 1 small change to request.

Instead of switching on this single topic, let's create a new configuration topic for the map type specifically. `componentLogsConfigurationMap`

so 
```
configTopicsPojo.computeIfPresent("componentLogsConfigurationMap", (s, o) -> {
  if (o instanceof Map) {
    .... same logic you added ....
 }
});
```

rbattle: +1 on this - the same approach (new topic) was used for shadow: https://github.com/aws-greengrass/aws-greengrass-shadow-manager/pull/93

PR #68, opened 2021-10-04
fix: do not upload logs older than 14 days

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Add a date validation in CW to ignore log events which are 14 days or older as per the CW documentation.

CW does not allow logs older than 14 days -> https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html
Due to this, Log Manager needs to skip those log events.

CW putEvents API also returns a rejected event info field if any log events are too old or too new or expired. Log Manager needs to be able to handle that case as well. If all the logs in the file are rejected, then no next sequence token is returned. That caused an NPE when adding it to the next token map.

Added unit test.

**Checklist:**


src/test/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessorTest.java
```diff
@@ -270,7 +274,48 @@ void GIVEN_one_component_one_file_more_than_max_WHEN_merge_THEN_reads_partial_fi
...
+            String logStream2 = sdf.format(new Date(now.toEpochMilli()));
+            assertThat(attempt.getLogStreamsToLogEventsMap().get(logStream1).getLogEvents(), hasSize(2));
+            assertThat(attempt.getLogStreamsToLogEventsMap().get(logStream2).getLogEvents(), hasSize(2));
+        } finally {
+            assertTrue(file.delete());
+        }
+    }
+
+    @Test
+    void GIVEN_one_component_one_file_older_than_14_days_WHEN_skip_file()
```
rbattle, line 318: nit: `GIVEN_one_component_one_file_older_than_14_days_WHEN_skip_file` _THEN ??

Or is this supposed to be: `GIVEN_one_component_WHEN_file_older_than_14_days_THEN_skip_file`

rbattle: This also doesn't seem to fully test the change - you have it filter out log events that are > 14 days.

Is it possible to have a log file with events that are > 14 days old and some that are < 14 days old?

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -332,6 +339,10 @@ private boolean addNewLogEvent(AtomicInteger totalBytesRead, CloudWatchAttemptLo
                 .isBefore(timestamp)) {
             return true;
         }
+        // Cloudwatch does not allow uploading data older than 14 days.
+        if (timestamp.isBefore(Instant.now(clock).minus(14, ChronoUnit.DAYS))) {
+            return false;
+        }
```
rbattle, line 345: should this be the first check in the method? you could return `true` on line 330 even if the timestamp is 14 days before now.

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -330,6 +330,10 @@ private boolean addNewLogEvent(AtomicInteger totalBytesRead, CloudWatchAttemptLo
             return true;
         }
 
+        // Cloudwatch does not allow uploading data older than 14 days.
+        if (timestamp.isBefore(Instant.now(clock).minus(14, ChronoUnit.DAYS))) {
+            return false;
+        }
```
rbattle, line 336: It still seems like this can return true even if the timestamp is too early. I haven't gone through the log manager code so - does that matter that this can return that the maximum message size has been reached even if the message to be added is filtered out?

MikeDombo: Returning true/false here doesn't mean that the log event will be uploaded, it is actually signaling if the batch is full or not. Adding the event happens at the very end of this method. It is not possible that a log event which is too old would be added

rbattle: right - I can see that we don't add the event - we return.
I'm just curious about the logic of this flag - if `current + x > limit`, but we aren't going to use `x`, why are we even checking?

MikeDombo: You're talking about L329? We want to check if we were to add the event, would it be too big. If so, then the batch is too big and this current event needs to go into the next batch.

rbattle: ok - it seems a bit odd to say the batch would be too large when the message that would trigger it being too large isn't ever going to be added to the batch. 

This is a bit of an edge case though so I'm not too worried. I'm just trying to understand what this code is doing

MikeDombo: The current event cannot fit, therefore the batch cannot be added to anymore. We don't try any further log events to try and fill the batch 100% or anything like that.

rbattle (approved):

nikkhilmuthye (approved):

PR #76, opened 2022-01-24
fix: respect log event size limit set by cloudwatch and break into ch…

shaguptashaikh:
…unks if reached

**Issue #, if available:**

**Description of changes:**
There's a limit on individual event size from cloudwatch but currently we treat each log line as one event and log lines can technically exceed that size. 

New and existing tests pass

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
+++ b/src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
@@ -14,6 +14,7 @@ import com.aws.greengrass.logmanager.model.CloudWatchAttemptLogFileInformation;
 import com.aws.greengrass.logmanager.model.CloudWatchAttemptLogInformation;
 import com.aws.greengrass.logmanager.model.ComponentLogFileInformation;
 import com.aws.greengrass.util.Coerce;
+import com.aws.greengrass.util.Pair;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -35,6 +36,7 @@ import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.time.temporal.ChronoUnit;
 import java.time.temporal.TemporalAccessor;
+import java.util.Arrays;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.Locale;
@@ -61,6 +63,7 @@ public class CloudWatchAttemptLogsProcessor {
     private static final int EVENT_STORAGE_OVERHEAD = 26;
     private static final int TIMESTAMP_BYTES = 8;
     private static final int MAX_BATCH_SIZE = 1024 * 1024;
+    private static final int MAX_EVENT_LENGTH = 1024 * 256 - TIMESTAMP_BYTES - EVENT_STORAGE_OVERHEAD;
     private static final ObjectMapper DESERIALIZER = new ObjectMapper()
             .configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true);
     private static final ThreadLocal<SimpleDateFormat> DATE_FORMATTER =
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -213,6 +215,8 @@ private boolean processLogLine(AtomicInteger totalBytesRead,
                                    long currentPosition,
                                    long lastModified) {
         boolean reachedMaxSize;
+        int expectedToAddSize = data.toString().getBytes(StandardCharsets.UTF_8).length;
```
MikeDombo, line 218: can we do this any cheaper? This is an expensive call.

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -324,22 +330,18 @@ private boolean checkAndAddNewLogEvent(AtomicInteger totalBytesRead,
...
-        // and there is an overhead for each log event on the cloud watch side which needs to be added.
-        if (totalBytesRead.get() + dataSize + TIMESTAMP_BYTES + EVENT_STORAGE_OVERHEAD > MAX_BATCH_SIZE) {
-            return true;
-        }
+        AtomicInteger currBytesAdded = new AtomicInteger();
 
         // Cloudwatch does not allow uploading data older than 14 days.
         if (timestamp.isBefore(Instant.now(clock).minus(14, ChronoUnit.DAYS))) {
-            return false;
+            return currBytesAdded;
```
MikeDombo, line 344: do we skip over the event higher up when the added count is `0`? Because that's what needs to happen in this case.

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -348,15 +350,34 @@ private boolean addNewLogEvent(AtomicInteger totalBytesRead, CloudWatchAttemptLo
                 attemptLogInformation.getLogEvents().stream().min(Comparator.comparingLong(InputLogEvent::timestamp));
         if (earliestTime.isPresent() && Instant.ofEpochMilli(earliestTime.get().timestamp()).plus(23, ChronoUnit.HOURS)
                 .isBefore(timestamp)) {
-            return true;
+            return currBytesAdded;
```
MikeDombo, line 353: same deal here.

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
shaguptashaikh: I think so, does line 256 address this concern?

MikeDombo: probably, yes. I believe it is tested, so if the tests are passing we should be OK.

Definitely add a test for your new requirement.

shaguptashaikh: not tested yet, just want to get an early opinion while I test and implement tests

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
shaguptashaikh: Computing only once now as it was done earlier

MikeDombo (approved):

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -245,11 +250,14 @@ private boolean processLogLine(AtomicInteger totalBytesRead,
...
-            reachedMaxSize = addNewLogEvent(totalBytesRead, attemptLogInformation, data.toString(), logTimestamp);
+            addEventResult =
+                    addNewLogEvent(totalBytesRead, attemptLogInformation, dataStr, dataSize, logTimestamp);
         }
-        if (!reachedMaxSize) {
-            updateCloudWatchAttemptLogInformation(fileName, startPosition, currentPosition, attemptLogInformation,
-                    lastModified);
+        boolean reachedMaxSize = addEventResult.getLeft();
+        int actualAddedSize = addEventResult.getRight().get();
+        if (actualAddedSize > 0) {
```
nikkhilmuthye, line 258: nit: you can use `reachedMaxSize` here

nikkhilmuthye (approved):

rbattle (approved):

rightpeter (commented):
Will fix these issues in another PR

PR #77, opened 2022-01-27
fix: bugs resulting log events created incorrectly

rightpeter:
**Issue #, if available:**

**Description of changes:**
* fix total chunk calculating logic to not introduce an extra chunk
* fix log line copy logic to copy correct chunk of log line instead
  of always copy from the begining

There are 2 case for failure:
* If a log line is multiple times of `MAX_EVENT_LENGTH`, 1 empty log event will be created
* For over size log line, LogManager can only copy from the beginning of the line

This fix resolves the bugs

* Unit-tests covered

**Checklist:**
 - [X] Updated or added new unit tests

n

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
+++ b/src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
@@ -367,23 +367,25 @@ public class CloudWatchAttemptLogsProcessor {
             logger.atTrace().kv("log-line-size-bytes", dataSize)
                     .log("Log line larger than maximum event size 256KB, " + "will be split into multiple log events");
         }
-        int totalChunks = dataSize / MAX_EVENT_LENGTH + 1;
+        int totalChunks = (dataSize - 1) / MAX_EVENT_LENGTH + 1;
         int currChunk = 1;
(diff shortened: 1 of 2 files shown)


shaguptashaikh (approved):

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -367,23 +367,26 @@ private void updateCloudWatchAttemptLogInformation(String fileName,
             logger.atTrace().kv("log-line-size-bytes", dataSize)
                     .log("Log line larger than maximum event size 256KB, " + "will be split into multiple log events");
         }
-        int totalChunks = dataSize / MAX_EVENT_LENGTH + 1;
+        int totalChunks = (int) Math.ceil((float) dataSize / MAX_EVENT_LENGTH);
```
junfuchen99, line 370: I think you could use `(dataSize - 1)/ MAX_EVENT_LENGTH + 1` to avoid casting and float division, assuming dataSize nonzero. Although it's a little intuitive so you might also need to add comments.

junfuchen99 (approved):

shaguptashaikh (approved):

PR #80, opened 2022-02-10
fix: reset isCurrentlyUploading flag on shutdown to avoid log manager…

shaguptashaikh:
… upload getting stuck on restart

**Issue #, if available:**

**Description of changes:**
periodic log processing and upload is only attempted when isCurrentlyUploading is false, when log manager shuts down and this flag is true, when log manager restarts it will keep thinking there's an ongoing update

Integ test added, verified that without the fix once log manager is restarted due to an error it keeps waiting for the upload attempt to finish but there is no upload attempt ongoing because it failed and caused the error. The test verifies the ideal fixed behavior where log manager attempts uploading after restart correctly

**Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
@@ -63,13 +63,15 @@ import static com.aws.greengrass.logging.impl.config.LogConfig.newLogConfigFromR
 import static com.aws.greengrass.logmanager.CloudWatchAttemptLogsProcessor.DEFAULT_LOG_STREAM_NAME;
 import static com.aws.greengrass.logmanager.LogManagerService.DEFAULT_FILE_REGEX;
 import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector.ignoreExceptionOfType;
+import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector.ignoreExceptionWithMessage;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.atLeastOnce;
-import static org.mockito.Mockito.lenient;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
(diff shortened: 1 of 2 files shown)


MikeDombo (dismissed):

junfuchen99 (dismissed):

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -737,8 +736,8 @@ private void deleteFiles(long minimumBytesToBeDeleted, Path directoryPath, Patte
     @Override
     @SuppressWarnings("PMD.UselessOverridingMethod")
     public void shutdown() throws InterruptedException {
-        // Need to override the function for tests.
         super.shutdown();
+        isCurrentlyUploading.set(false);
```
nikkhilmuthye, line 740: What will happen if the logs have already been uploaded and the Log Manager is shutting down that time? The logs would have reached CW but we would not have updated the state in LogManager. So when LogManager starts up again, it will again reupload those logs which have already been uploaded. 
This is just an optimization, but should we wait until the current execution of logs has been completed before shutting the component down? We did something similar in Shadow Manager as well.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -142,10 +142,12 @@ void setupKernel(Path storeDirectory, String configFileName) throws InterruptedE
 
     @BeforeEach
     void beforeEach(ExtensionContext context) {
+        kernel = new Kernel();
         ignoreExceptionOfType(context, TLSAuthException.class);
         ignoreExceptionOfType(context, InterruptedException.class);
         ignoreExceptionOfType(context, DateTimeParseException.class);
         ignoreExceptionOfType(context, CrtRuntimeException.class);
+        ignoreExceptionOfType(context, SdkClientException.class);
```
MikeDombo, line 150: ignore this in _every_ test? Can we only ignore it when it will actually happen?

MikeDombo (approved):

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
shaguptashaikh: We could try that, but doesn't seem straightforward and will definitely need more testing. So, it'll be taken up separately as a follow up to push this out soon

MikeDombo (dismissed):

MikeDombo (approved):

junfuchen99 (approved):

PR #81, opened 2022-02-10
fix: component config reset bug

J-C-L:
**Issue #, if available:**
 
**Description of changes:**
Fixing bugs so that when logFileRegex or periodicUploadIntervalSec config key is removed, the value is reset to the correct default.
Also adding testing for correct reset values for other component config keys

**Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
@@ -44,6 +44,7 @@ import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.text.SimpleDateFormat;
+import java.time.Duration;
 import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.Date;
@@ -63,13 +64,16 @@ import static com.aws.greengrass.logging.impl.config.LogConfig.newLogConfigFromR
 import static com.aws.greengrass.logmanager.CloudWatchAttemptLogsProcessor.DEFAULT_LOG_STREAM_NAME;
 import static com.aws.greengrass.logmanager.LogManagerService.DEFAULT_FILE_REGEX;
 import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector.ignoreExceptionOfType;
+import static com.github.grantwest.eventually.EventuallyLambdaMatcher.eventuallyEval;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.atLeastOnce;
+import static org.mockito.Mockito.lenient;
 import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
 
 @SuppressWarnings("PMD.UnsynchronizedStaticFormatter")
 @ExtendWith({GGExtension.class, MockitoExtension.class})
@@ -83,6 +87,9 @@ class LogManagerTest extends BaseITCase {
     private Path tempDirectoryPath;
     private final static ObjectMapper YAML_OBJECT_MAPPER = new ObjectMapper(new YAMLFactory());
 
+    private static final Level MINIMUM_LOG_LEVEL_DEFAULT =  Level.INFO;
+    private static final boolean DELETE_LOG_FILE_AFTER_CLOUD_UPLOAD_DEFAULT = false;
+    private static final String FILE_NAME_REGEX_DEFAULT = "^\\QUserComponentA\\E\\w*.log";
     static {
         DATE_FORMATTER.setTimeZone(TimeZone.getTimeZone("UTC"));
     }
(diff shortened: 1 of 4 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -283,4 +291,60 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_interval_elapses_THEN
...
+        assertThat(()-> logManagerService.getPeriodicUpdateIntervalSec(), eventuallyEval(is(10), Duration.ofSeconds(30)));
+        logManagerService.getConfig().find("configuration", "periodicUploadIntervalSec").remove();
+        assertThat(()-> logManagerService.getPeriodicUpdateIntervalSec(),
+                eventuallyEval(is(logManagerService.DEFAULT_PERIODIC_UPDATE_INTERVAL_SEC), Duration.ofSeconds(30)));
+
+        // Verify correct reset for fileNameRegex
+        assertThat(()-> logManagerService.getComponentLogConfigurations().get("UserComponentA").getFileNameRegex().pattern(),
+                eventuallyEval(is("^integTestRandomLogFiles.log\\w*"), Duration.ofSeconds(30)));
+        logManagerService.getConfig().find("configuration", "logsUploaderConfiguration",
+                "componentLogsConfigurationMap", "UserComponentA", "logFileRegex").remove();
```
tilo-chen, line 314: nit: can we reference all these config keys from existing string constants? e.g. "logsUploaderConfiguration" -> LOGS_UPLOADER_CONFIGURATION_TOPIC

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -283,4 +291,60 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_interval_elapses_THEN
         }
         assertEquals(1, allFiles.size());
     }
-}
+
+    @Test
+    void GIVEN_component_configs_WHEN_individual_configs_are_reset_THEN_correct_default_values_are_used()
+            throws Exception {
+        Path LOG_FILE_DIRECTORY_PATH_DEFAULT = tempRootDir.resolve("logs");
```
tilo-chen, line 298: nit: if this is only used once, it doesn't need to be an all-caps constant. You can also define it closer to where it's used

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -230,8 +239,8 @@ private void setUserComponentConfiguration(Map<String, Object> componentConfigMa
                     if (Utils.isNotEmpty(logFileRegexString)) {
                         fileNameRegex.set(Pattern.compile(logFileRegexString));
                     } else {
-                        fileNameRegex.set(Pattern.compile(String.format(DEFAULT_FILE_REGEX,
-                                Pattern.quote(LogManager.getRootLogConfiguration().getFileName()))));
+                        fileNameRegex.set(Pattern.compile(String.format(DEFAULT_FILE_REGEX, Pattern.quote(
+                                Coerce.toString(componentConfigMap.get(COMPONENT_NAME_CONFIG_TOPIC_NAME))))));
                     }
```
rbattle, line 244: do you even need this else statement? Further down on line 269 it checks if the regex is set, and does a similar Pattern.compile if not.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -128,7 +131,13 @@
 
         topics.lookup(CONFIGURATION_CONFIG_KEY, LOGS_UPLOADER_PERIODIC_UPDATE_INTERVAL_SEC)
                 .dflt(DEFAULT_PERIODIC_UPDATE_INTERVAL_SEC)
-                .subscribe((why, newv) -> periodicUpdateIntervalSec = Coerce.toInt(newv));
+                .subscribe((why, newv) -> {
+                    if (why == WhatHappened.removed) {
+                        periodicUpdateIntervalSec = DEFAULT_PERIODIC_UPDATE_INTERVAL_SEC;
+                    } else {
+                        periodicUpdateIntervalSec = Coerce.toInt(newv);
```
rbattle, line 138: check if int is > than some minimum - do we want to allow 1s, negative numbers, 0, etc...

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -283,4 +291,60 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_interval_elapses_THEN
...
+
+        tempDirectoryPath = Files.createTempDirectory(tempRootDir, "IntegrationTestsTemporaryLogFiles");
+        setupKernel(tempDirectoryPath, "configsDifferentFromDefaults.yaml");
+        TimeUnit.SECONDS.sleep(30);
```
rbattle, line 302: why sleep 30? Is there a way to execute this test without it taking so long

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
J-C-L: Good call. Changed.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
J-C-L: Thought I needed wait time after launching the kernel since that's what other tests do. Seem to work consistently without it though, so will remove the wait time.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
J-C-L: As discussed offline, I'll check > 0, and if it fails, I'll log a warning and use the default.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
J-C-L: changed and moved.

junfuchen99 (approved):

PR #82, opened 2022-02-15
fix: bug when config values are replaced after a reset

J-C-L:
**Issue #, if available:**

**Description of changes:**
Fixing bug, so that when periodicUploadIntervalSec or logsUploaderConfiguration config keys are removed via reset, and subsequently replaced, then the new values will be used.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -101,8 +101,8 @@ public class LogManagerService extends PluginService {
             Collections.synchronizedMap(new LinkedHashMap<>());
     final Map<String, CurrentProcessingFileInformation> componentCurrentProcessingLogFile =
             new ConcurrentHashMap<>();
-    // public only for integ tests
-    public Map<String, ComponentLogConfiguration> componentLogConfigurations = new ConcurrentHashMap<>();
+    @Getter
+    Map<String, ComponentLogConfiguration> componentLogConfigurations = new ConcurrentHashMap<>();
     @Getter
     private final CloudWatchLogsUploader uploader;
     private final CloudWatchAttemptLogsProcessor logsProcessor;
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -8,6 +8,7 @@ package com.aws.greengrass.logmanager;
 import ch.qos.logback.core.util.FileSize;
 import com.aws.greengrass.config.Topic;
 import com.aws.greengrass.config.Topics;
+import com.aws.greengrass.config.WhatHappened;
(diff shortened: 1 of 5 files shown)


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -129,38 +128,44 @@
...
-                        } else {
-                            logger.atWarn().log("Invalid config value, {}, for periodicUploadIntervalSec. Must be an "
-                                    + "integer greater than 0. Using default value of 300 (5 minutes)",
-                                    Coerce.toInt(newv));
-                            periodicUpdateIntervalSec = DEFAULT_PERIODIC_UPDATE_INTERVAL_SEC;
-                        }
-                    }
-                });
+        topics.lookupTopics(CONFIGURATION_CONFIG_KEY).subscribe((why, newv) -> {
+            logger.atWarn().log("Log manager config change. Why: {}, Node: {}", why, newv);
```
rbattle, line 132: you can remove this

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -129,38 +128,44 @@
...
+            handlePeriodicUploadIntervalSecConfig(topics);
+            handleLogsUploaderConfig(topics);
```
rbattle, line 134: subscriptions can fire more often then you think -`why` can just be a time stamp updated and the value doesn't change. This can happen around deployments.

Will this behavior (no actual diff in value) cause any extra work to occur?

junfuchen99 (commented):
I think `multiLineStartPattern` is also user configurable in component log config. We should also test resetting it.

Not sure why it's not documented in public doc.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
J-C-L: Oh. I'd meant to change it to DEBUG at have it there for debugging. Is that fine or would you prefer it be removed?

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
J-C-L: Updated to exclude WhatHappened.timestampUpdated. The other options for WhatHappened seem to involve potential value changes.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
rbattle: if you keep it, use the `.kv("why", why).kv("node", newv)`

J-C-L: Done.

junfuchen99 (approved):

rbattle (approved):

PR #83, opened 2022-02-17
fix: system config reset bug

J-C-L:
**Issue #, if available:**

**Description of changes:**
Fixing bugs so that when uploadToCloudWatch or diskSpaceLimitUnit system configs are reset and then replaces, the correct values are used.

Added testing for resetting and replacing system configs

**Checklist:**


src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerConfigTest.java
```diff
@@ -0,0 +1,327 @@
...
+                    eventuallyEval(is(LogManagerService.DEFAULT_PERIODIC_UPDATE_INTERVAL_SEC), Duration.ofSeconds(30)));
+
+            logManagerService.getConfig().lookup(CONFIGURATION_CONFIG_KEY, LOGS_UPLOADER_PERIODIC_UPDATE_INTERVAL_SEC).withValue(600);
+            assertThat(()-> logManagerService.getPeriodicUpdateIntervalSec(), eventuallyEval(is(600),
+                    Duration.ofSeconds(30)));
+    }
+
+    @Test
+    void GIVEN_uploadToCloudWatch_config_WHEN_value_is_reset_and_replaced_THEN_correct_values_are_used() {
+        // uploadToCloudWatch=true in recipe, so system configs expected in componentLogConfigurations map
```
tilo-chen, line 119: I remember it's not in recipe. But the yaml config we used here has system configs. So I think this test case needs to assert it's true at first, then do removal and assert that it resets to null/false.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -355,7 +355,10 @@ private void setMinimumLogLevel(String minimumLogLevel, ComponentLogConfiguratio
 
     private void setDiskSpaceLimit(String diskSpaceLimit, String diskSpaceLimitUnit,
                                    ComponentLogConfiguration componentLogConfiguration) {
-        if (!StringUtils.isEmpty(diskSpaceLimit) && !StringUtils.isEmpty(diskSpaceLimitUnit)) {
+        if (!StringUtils.isEmpty(diskSpaceLimit)) {
+             if (StringUtils.isEmpty(diskSpaceLimitUnit)) {
+                    diskSpaceLimitUnit = "KB";
+                }
```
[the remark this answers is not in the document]
J-C-L: Fixed.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerConfigTest.java
J-C-L: Agreed that since it's true in the yaml file, the test needs to first verify that it's true, then remove it and verify that it's false, and then replace with true and verify that it's true again. For uploadToCloudWatch, the value itself isn't saved into componentLogConfigurations,  but rather if it is false, the whole system component configs aren't added to componentLogConfigurations. So to verify whether uploadToCloudWatch is true/false, the test is checking whether or not componentLogConfigurations contains the system component configs.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -355,7 +355,10 @@ private void setMinimumLogLevel(String minimumLogLevel, ComponentLogConfiguratio
 
     private void setDiskSpaceLimit(String diskSpaceLimit, String diskSpaceLimitUnit,
                                    ComponentLogConfiguration componentLogConfiguration) {
-        if (!StringUtils.isEmpty(diskSpaceLimit) && !StringUtils.isEmpty(diskSpaceLimitUnit)) {
+        if (!StringUtils.isEmpty(diskSpaceLimit)) {
```
junfuchen99, line 358: Is `diskSpaceLimitUnit` still nonempty when there is no node change for `DISK_SPACE_LIMIT_UNIT_CONFIG_TOPIC_NAME`? Suppose the initial disk space limit is 20MB, if customer updates it to 30MB, so that the `diskSpaceLimitUnit` is not updated, will it be changed to 30KB since the default is KB?

J-C-L: The 2 configs are each set independently from their value in the input map. So if the user changes the diskSpaceLimit but does not change the diskSpaceLimitUnit, then the units will remain the same. In the example you proposed, the disk space limit will change from 20MB to 30MB.  The units will only change to KB if the diskSpaceLimitUnit value is explicitly removed

PR #85, opened 2022-02-28
fix: restart space management thread if it dies for some reason

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Updates log space-based deletion so that we only walk the directory once and then immediately attempt to delete files.

Catches all errors in the space management thread and reschedule it to run again if it fails for whatever reason. It was failing for some customers due to an unchecked io exception when getting the length of a file which no longer exists.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -59,7 +59,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-import java.util.stream.LongStream;
+import java.util.stream.Stream;
 import javax.inject.Inject;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -342,6 +344,9 @@ private void scheduleSpaceManagementThread() {
                     logger.atError().cause(e).log("Unable to start space management thread.");
                 } catch (InterruptedException e) {
                     Thread.currentThread().interrupt();
+                } catch (Throwable e) {
+                    logger.atError().log("Failure in log manager space management", e);
+                    scheduleSpaceManagementThread(); // restart space management
```
tilo-chen, line 349: nit: also say we're restarting the thread in the log?

tilo-chen (approved):

shaguptashaikh (approved):

shaguptashaikh (commented):
Would be nice to have a test for this

PR #88, opened 2022-03-08
fix: fix NPE in checking diskSpaceLimit

shaguptashaikh:
**Issue #, if available:**

**Description of changes:**
Adding null check to diskSpaceLimit config usage 

Long running tests failed

Existing tests

**Checklist:**


MikeDombo (dismissed):

MikeDombo:
https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/6e6aacb6c53ed8122c2761623eb3e7d4fd6ff273/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java#L734
^ Should also check for null here. May consider changing `Long` to `long` to that it just can't be null.

shaguptashaikh:
> https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/6e6aacb6c53ed8122c2761623eb3e7d4fd6ff273/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java#L734
> 
> 
> ^ Should also check for null here. May consider changing `Long` to `long` to that it just can't be null.


Added the check. Changing to long seems to have a larger blast radius. Since we don't have a default for diskSpaceLimit when it's missing in component config, it will effectively not attempt to delete any files when the config is missing. Not sure that's how we always wanted it though

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -731,7 +732,8 @@ private void deleteFilesIfNecessary(ComponentLogConfiguration componentLogConfig
                 })) {
             List<Path> paths = fileStream.collect(Collectors.toList());
             long totalBytes = paths.stream().mapToLong(p -> p.toFile().length()).sum();
-            long minimumBytesToBeDeleted = totalBytes - componentLogConfiguration.getDiskSpaceLimit();
+            long minimumBytesToBeDeleted = componentLogConfiguration.getDiskSpaceLimit() == null ? 0
```
MikeDombo, line 735: if it is 0, I'd just return immediately, there's no work to do

MikeDombo: You can do that check at the top of this method, or at the call-site

MikeDombo:
> > https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/6e6aacb6c53ed8122c2761623eb3e7d4fd6ff273/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java#L734
> > 
> > ^ Should also check for null here. May consider changing `Long` to `long` to that it just can't be null.
> 
> Added the check. Changing to long seems to have a larger blast radius. Since we don't have a default for diskSpaceLimit when it's missing in component config, it will effectively not attempt to delete any files when the config is missing. Not sure that's how we always wanted it though

That's the default, nothing wrong with that right? Nucleus logger still manages the memory limits

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
rightpeter: It'll return after this. I guess it's fine

    if (minimumBytesToBeDeleted <= 0 || paths.size() < 2) {
        return;
    }

MikeDombo: You can avoid the path listing which is expensive. Do it earlier on.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -642,7 +642,8 @@ private void startWatchServiceOnLogFilePaths() throws IOException, InterruptedEx
             try {
                 componentLogConfigurations.forEach((componentName, componentLogConfiguration) -> {
                     // Only register the path of a component if the disk space limit is set.
-                    if (componentLogConfiguration != null && componentLogConfiguration.getDiskSpaceLimit() > 0) {
+                    if (componentLogConfiguration != null && componentLogConfiguration.getDiskSpaceLimit() != null
```
rightpeter, line 645: Do we let `.getDiskSpaceLimit()` return `null` on purpose? Is there a reason not asking it to return a default value if not it's not configured? (Don't know if it's a good pattern though).

MikeDombo (approved):

rightpeter (approved):

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
shaguptashaikh: If we had a default that's what we'd do but we don't have a default

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
rightpeter: Right. Good point

PR #92, opened 2022-05-18
fix: address null filename stored in runtime configuration

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Does not store `currentProcessingFileInformation` when the filename is null or empty which should not ever happen. Synchronizes `currentComponentFileProcessingInformation` writing by running all of the writes in a single batch which should prevent any possibility for reads and writes overlapping.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -8,6 +8,7 @@ package com.aws.greengrass.logmanager;
 import ch.qos.logback.core.util.FileSize;
 import com.aws.greengrass.config.Topic;
 import com.aws.greengrass.config.Topics;
+import com.aws.greengrass.config.UpdateBehaviorTree;
 import com.aws.greengrass.config.WhatHappened;
 import com.aws.greengrass.dependency.ImplementsService;
 import com.aws.greengrass.lifecyclemanager.PluginService;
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -468,11 +472,17 @@ private void handleCloudWatchAttemptStatus(CloudWatchAttempt cloudWatchAttempt)
...
+        context.runOnPublishQueueAndWait(() -> {
+            componentCurrentProcessingLogFile.forEach((componentName, currentProcessingFileInformation) -> {
+                Topics componentTopics =
+                        getRuntimeConfig().lookupTopics(PERSISTED_COMPONENT_CURRENT_PROCESSING_FILE_INFORMATION,
+                                componentName);
+                componentTopics.updateFromMap(currentProcessingFileInformation.convertToMapOfObjects(),
+                        new UpdateBehaviorTree(UpdateBehaviorTree.UpdateBehavior.REPLACE, System.currentTimeMillis()));
+            });
         });
+        context.waitForPublishQueueToClear();
```
rbattle, line 484: nice - I was wondering if the publish was somehow responsible for the issue

MikeDombo: Since I could not reproduce the issue, I'm not entirely sure that it is, but I just want to remove it as a possible source.

rbattle: yeah - makes sense to batch up the updates though

rbattle (approved):

junfuchen99 (approved):

PR #94, opened 2022-09-09
feat: add hashValue in LogFileInformation and calculate it for all files

dgggIoT:
**Issue #, if available:**

**Description of changes:**
1. apply the digest for all files
2. add the hashValue into the LogFileInformation structure

Unit testing and integration testing will be written in next commit

This is a partial commit, not ready to be merged.

**Checklist:**


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -628,6 +648,55 @@ private void processLogsAndUpload() throws InterruptedException {
...
+     * @param activeFile
+     * @return the nullable calculated hash for active file
+     */
+    private Optional<String> calculateHashForActiveFile(File activeFile) {
+        Optional activeFileHash = Optional.ofNullable(null);
+        // String activeFileHash = null;
+        try (SeekableByteChannel chan = Files.newByteChannel(activeFile.toPath(), StandardOpenOption.READ);
+             PositionTrackingBufferedReader r =
+                new PositionTrackingBufferedReader(new InputStreamReader(Channels.newInputStream(chan),
+                        StandardCharsets.UTF_8))) {
```
MikeDombo, line 663: you don't need any of this complication for reading the file. Just use `Files.newBufferedReader`

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -628,6 +648,55 @@ private void processLogsAndUpload() throws InterruptedException {
...
+            r.setInitialPosition(0);
+            chan.position(0);
+            StringBuilder data = new StringBuilder(r.readLine());
+
+            // if the lines for digest is less than 1 line, skip this file and generate error.
+            if (DEFAULT_LINES_FOR_DIGEST_NUM < 1) {
+                logger.atError().cause(new InvalidDigestTargetLinesNumberException("The active log file is " +
```
MikeDombo, line 670: if you want to log something, you must call `.log()`.


This is _not_ an error that should be logged. No one cares that there weren't enough lines, there's absolutely nothing that anyone can do about this error.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -628,6 +648,55 @@ private void processLogsAndUpload() throws InterruptedException {
...
+        try (SeekableByteChannel chan = Files.newByteChannel(activeFile.toPath(), StandardOpenOption.READ);
+             PositionTrackingBufferedReader r =
+                new PositionTrackingBufferedReader(new InputStreamReader(Channels.newInputStream(chan),
+                        StandardCharsets.UTF_8))) {
+            r.setInitialPosition(0);
+            chan.position(0);
+            StringBuilder data = new StringBuilder(r.readLine());
+
+            // if the lines for digest is less than 1 line, skip this file and generate error.
+            if (DEFAULT_LINES_FOR_DIGEST_NUM < 1) {
```
MikeDombo, line 669: You're checking `1 < 1`, we always know that is false, this can not ever be true.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -579,12 +595,15 @@ private void processLogsAndUpload() throws InterruptedException {
...
                     allFiles.sort(Comparator.comparingLong(File::lastModified));
-                    // If there are no rotated log files for the component, then return.
-                    if (allFiles.size() - 1 <= 0) {
-                        continue;
-                    }
-                    // Don't consider the active log file.
+
+                     // sublist allFiles to rotated file and active file, calculate the hash for active file with
+                    // some lines, default is 1 line
                     allFiles = allFiles.subList(0, allFiles.size() - 1);
```
MikeDombo, line 601: `allFiles` is already excluding the active log file. Your added code is further slicing a list which has already had the active file removed

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -579,12 +595,15 @@ private void processLogsAndUpload() throws InterruptedException {
...
+                    rotatedFiles = allFiles.subList(0, allFiles.size() - 1);
+                    File activeFile = allFiles.get(allFiles.size() - 1);
+
+                    //calculate the hash value for activeFile
```
MikeDombo, line 605: we're making the assumption that the active file has the latest modified time. Can we remove that assumption and still be correct?

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
dgggIoT: This check is written for a future case: if the customer is allowed to customize the `DEFAULT_LINES_FOR_DIGEST_NUM`, then any non-positive integers will be rejected.

MikeDombo: If you do add that, this isn't where that logic would go. This is premature, so I think you should remove it.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
dgggIoT: Good question. But will this assumption is naturally true? Because the active file must be created after the rotated file.

MikeDombo: Can we remove the assumption? It would be much better if we don't need to make any assumption at all.

Why do we need to treat the files differently?

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -628,6 +648,55 @@ private void processLogsAndUpload() throws InterruptedException {
...
+                        StandardCharsets.UTF_8))) {
+            r.setInitialPosition(0);
+            chan.position(0);
+            StringBuilder data = new StringBuilder(r.readLine());
+
+            // if the lines for digest is less than 1 line, skip this file and generate error.
+            if (DEFAULT_LINES_FOR_DIGEST_NUM < 1) {
+                logger.atError().cause(new InvalidDigestTargetLinesNumberException("The active log file is " +
+                        "not read because the Digest requires at least one line message."));
+            } else {
```
Nelsonochoam, line 672: IMO: it is generally better not to nest conditional statement it makes the code harder to follow. A few ways to remedy this is to 

* Extract code into methods
* Use guard pattern [See](https://refactoring.guru/replace-nested-conditional-with-guard-clauses)

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -628,6 +648,55 @@ private void processLogsAndUpload() throws InterruptedException {
         }
     }
 
+    /**
+     * calculate the hash for the active file, which takes and digests first some lines (default is 1) in the file.
+     * If the content of file is insufficient, or incorrect setting for DEFAULT_LINES_FOR_DIGEST_NUM, the hash is null.
+     * @param activeFile
+     * @return the nullable calculated hash for active file
+     */
+    private Optional<String> calculateHashForActiveFile(File activeFile) {
```
Nelsonochoam, line 657: This seems like important functionality - lets write some tests for it

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -579,12 +595,15 @@ private void processLogsAndUpload() throws InterruptedException {
...
-                    // Don't consider the active log file.
+
+                     // sublist allFiles to rotated file and active file, calculate the hash for active file with
+                    // some lines, default is 1 line
                     allFiles = allFiles.subList(0, allFiles.size() - 1);
+                    rotatedFiles = allFiles.subList(0, allFiles.size() - 1);
+                    File activeFile = allFiles.get(allFiles.size() - 1);
+
+                    //calculate the hash value for activeFile
+                    Optional activeFileHash = Optional.ofNullable(calculateHashForActiveFile(activeFile));
```
Nelsonochoam, line 606: This line is not yet being used

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
Nelsonochoam: +1

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
dgggIoT: Here will be used in the next commit.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -649,52 +639,38 @@ private void processLogsAndUpload() throws InterruptedException {
...
-    private Optional<String> calculateHashForActiveFile(File activeFile) {
-        Optional activeFileHash = Optional.ofNullable(null);
-        // String activeFileHash = null;
-        try (SeekableByteChannel chan = Files.newByteChannel(activeFile.toPath(), StandardOpenOption.READ);
-             PositionTrackingBufferedReader r =
-                new PositionTrackingBufferedReader(new InputStreamReader(Channels.newInputStream(chan),
-                        StandardCharsets.UTF_8))) {
-            r.setInitialPosition(0);
-            chan.position(0);
+    private Optional<String> calculateHashForLogFile(File logFile) {
```
Nelsonochoam, line 647: Tests?

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -649,52 +639,38 @@ private void processLogsAndUpload() throws InterruptedException {
...
+        Optional fileHash = Optional.ofNullable(null);
```
Nelsonochoam, line 648: Why not use `Optional.empty()`?

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -649,52 +639,38 @@ private void processLogsAndUpload() throws InterruptedException {
...
+        try (BufferedReader r = Files.newBufferedReader(logFile.toPath(), StandardCharsets.UTF_8)) {
             StringBuilder data = new StringBuilder(r.readLine());
```
Nelsonochoam, line 650: Why not start the string builder empty. The while loop should handle the `r.readLine()`

```
StringBuilder data = new StringBuilder();
```

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -615,8 +602,10 @@ private void processLogsAndUpload() throws InterruptedException {
 
                     allFiles.forEach(file -> {
                         long startPosition = 0;
+                        Optional fileHash = Optional.ofNullable(calculateHashForLogFile(file));
```
Nelsonochoam, line 605: Shouldn't need to wrap it around a `Optional.ofNullable`. `calculateHashForLogFile` already returns an optional

Issue #51, opened 2021-03-15
(LogManager): Consider uploading active log file


fiLLLip:
**Feature Description**
This line skips the active log file for some reason. Why? We wasted an entire day on debugging why the current logs did not upload.

https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/e52b9d120521ebb06205a273b8349015669252e0/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java#L513

**Use Case**
Getting "real time" logs up in CloudWatch.

**Proposed Solution**
Remove this line to enable uploading of partial active logs. It looks like the partial upload is in place already.

**Other**

* [ ] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

jordanjennings:
In the short term the documentation should be updated to indicate that logs are not uploaded until the file is rotated, because I also wasted some time trying to figure out why logs were not uploading.

But that's just a band-aid for the real problem - I think all users expect logs to be shipped in (near) real-time.

mora0136:
I would like to see this become a feature of this component, I can help in getting that available. Are there any blockers for this? It has been a year with no solution suggested so I am cautious to start work in case there is a known reason why this can't be done.

vcarnogu:
Any plans to implement it? Not having the current file when troubleshooting remotely is a pain.

Nelsonochoam:
Hi all -- we just released support for uploading the active log file with Log Manager 2.3.0 https://docs.aws.amazon.com/greengrass/v2/developerguide/greengrass-release-2022-11-18.html

PR #98, opened 2022-09-27
feat: add log file class with hash generation

dgggIoT:
**Issue #, if available:**
#51 
**Description of changes:**
Add a class LogFile to include business logic, i.e. hash generation
This is one step of #51 , and it refines the code structure
Unittest
Can be merged but not ready for release.
**Checklist:**
 - [ x] Updated or added new unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -99,8 +99,8 @@ public class LogManagerService extends PluginService {
     public static final String UPLOAD_TO_CW_CONFIG_TOPIC_NAME = "uploadToCloudWatch";
     public static final String MULTILINE_PATTERN_CONFIG_TOPIC_NAME = "multiLineStartPattern";
     public static final int DEFAULT_PERIODIC_UPDATE_INTERVAL_SEC = 300;
-    public static final int DEFAULT_LINES_FOR_DIGEST_NUM = 1;
     private final Object spaceManagementLock = new Object();
+    private static final String HASH_VALUE_OF_EMPTY_STRING = "";
 
     // public only for integ tests
     public final Map<String, Instant> lastComponentUploadedLogFileInstantMap =
@@ -588,6 +588,12 @@ public class LogManagerService extends PluginService {
                     // Sort the files by the last modified time. Then try to proceed oldest file first to avoid data
                     // loss caused by auto deletion
                     allFiles.sort(Comparator.comparingLong(LogFile::lastModified));
+                    // If there are no rotated log files for the component, then return.
+                    if (allFiles.size() - 1 <= 0) {
+                        continue;
+                    }
+                    // Don't consider the active log file.
+                    allFiles = allFiles.subList(0, allFiles.size() - 1);
 
                     componentLogFileInformation.set(Optional.of(
                             ComponentLogFileInformation.builder()
@@ -599,23 +605,26 @@ public class LogManagerService extends PluginService {
 
                     allFiles.forEach(file -> {
                         long startPosition = 0;
-                        // If the file was partially read in the previous run, then get the starting position for
-                        // new log lines.
-                        //TODO: replace the filename with hashvalue
-                        if (componentCurrentProcessingLogFile.containsKey(componentName)) {
-                            CurrentProcessingFileInformation processingFileInformation =
-                                    componentCurrentProcessingLogFile.get(componentName);
-                            if (processingFileInformation.fileName.equals(file.getAbsolutePath())
-                                    && processingFileInformation.lastModifiedTime == file.lastModified()) {
-                                startPosition = processingFileInformation.startPosition;
+                        String fileHash = file.hashString();
+                        // It must be a valid file for uploading
+                        if (!fileHash.equals(HASH_VALUE_OF_EMPTY_STRING)) {
+                            // If the file was partially read in the previous run, then get the starting position for
+                            // new log lines.
+                            if (componentCurrentProcessingLogFile.containsKey(componentName)) {
+                                CurrentProcessingFileInformation processingFileInformation =
+                                        componentCurrentProcessingLogFile.get(componentName);
+                                if (processingFileInformation.fileName.equals(file.getAbsolutePath())
+                                        && processingFileInformation.lastModifiedTime == file.lastModified()) {
+                                    startPosition = processingFileInformation.startPosition;
+                                }
                             }
+                            LogFileInformation logFileInformation = LogFileInformation.builder()
+                                            .logFile(file)
+                                            .startPosition(startPosition)
+                                            .fileHash(fileHash)
+                                            .build();
+                            componentLogFileInformation.get().get().getLogFileInformationList().add(logFileInformation);
                         }
-                        LogFileInformation logFileInformation = LogFileInformation.builder()
-                                .logFile(file)
-                                .startPosition(startPosition)
-                                .fileHash(file.hashString())
-                                .build();
-                        componentLogFileInformation.get().get().getLogFileInformationList().add(logFileInformation);
                     });
                     break;
                 } catch (SecurityException e) {
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -163,8 +163,8 @@ public class LogManagerService extends PluginService {
     }
 
     /**
-     * Find the current 'logsUploaderConfiguration' configuration from runtime
-     * @param topics
+     * Find the current 'logsUploaderConfiguration' configuration from runtime.
+     * @param topics The topics to search.
      */
     private void handleLogsUploaderConfig(Topics topics) {
         Topics logsUploaderTopics = topics.lookupTopics(CONFIGURATION_CONFIG_KEY, LOGS_UPLOADER_CONFIGURATION_TOPIC);
@@ -577,7 +577,7 @@ public class LogManagerService extends PluginService {
                     LogFile[] files = LogFile.of(folder.listFiles());
                     if (files.length != 0) {
                         for (LogFile file : files) {
-                            if (file.isFile()
+                            if (file.isLogFile()
                                     && lastUploadedLogFileTimeMs.isBefore(Instant.ofEpochMilli(file.lastModified()))
                                     && componentLogConfiguration.getFileNameRegex().matcher(file.getName()).find()
                                     && file.length() > 0) {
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -0,0 +1,89 @@
...
+            // read target number of lines
+            while (linesReadForDigest < linesNeededNum) {
+                oneLine = r.readLine();
+                if (oneLine == null) break;
+                linesReadForDigest ++;
+                linesReaded.add(oneLine);
+            }
+            // read one more line to prevent incomplete last line
+            // if lines in file less than minLines + 1, then return empty
+            if (r.readLine() == null || linesReadForDigest < linesNeededNum) {
```
Nelsonochoam, line 62: No need for all this if logic. Just return whatever is on `linesRead`

src/test/java/com/aws/greengrass/logmanager/model/LogFileTest.java
```diff
@@ -0,0 +1,73 @@
...
+        File file = new File(directoryPath.resolve("greengrass_test.log").toUri());
+        writeFiles(file, DEFAULT_LINES_FOR_DIGEST_NUM + 1);
+        StringBuilder msg = new StringBuilder();
+        for (int i = 0; i < DEFAULT_LINES_FOR_DIGEST_NUM; i++) msg.append("line").append(i + 1);
+
+        LogFile logFile = LogFile.of(file);
+        Optional<String> fileHash = logFile.hashString();
+        assertEquals(fileHash.get(), calculate(msg.toString()));
+        file.delete();
+
```
Nelsonochoam, line 71:
```suggestion
```

src/test/java/com/aws/greengrass/logmanager/model/LogFileTest.java
```diff
@@ -0,0 +1,73 @@
...
+                fileOutputStream.write("\n".getBytes(StandardCharsets.UTF_8));
+            }
+        }
+    }
+
+
+    @Test
+    void GIVEN_log_file_with_equal_to_target_lines_WHEN_calculate_file_hash_THEN_we_get_null() throws IOException {
+        File file = new File(directoryPath.resolve("greengrass_test.log").toUri());
+        writeFiles(file, DEFAULT_LINES_FOR_DIGEST_NUM);
```
Nelsonochoam, line 52: Why if we only need one line and the file has 1 line we can't process the hash?

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+        return String.join("", linesReaded);
+    }
+
+    /**
+     * Get the hash of the logfile with target lines.
+     * @return the calculated hash value of the logfile
+     */
+    public Optional<String> hashString() {
+        if (!this.exists()) return Optional.ofNullable(null);
```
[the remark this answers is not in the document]
dgggIoT: Not sure if hash a empty file can always generate an identical hash value. 
But actually, we have other logic to filter out the case that file does not exist.

src/test/java/com/aws/greengrass/logmanager/model/LogFileTest.java
dgggIoT: This is to prevent this line is partially written.

Nelsonochoam (dismissed):

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -593,23 +602,29 @@ private void processLogsAndUpload() throws InterruptedException {
...
-                        // new log lines.
-                        if (componentCurrentProcessingLogFile.containsKey(componentName)) {
-                            CurrentProcessingFileInformation processingFileInformation =
-                                    componentCurrentProcessingLogFile.get(componentName);
-                            if (processingFileInformation.fileName.equals(file.getAbsolutePath())
-                                    && processingFileInformation.lastModifiedTime == file.lastModified()) {
-                                startPosition = processingFileInformation.startPosition;
+                        String fileHash = file.hashString();
+                        // It must be a valid file for uploading
+                        if (!fileHash.equals(HASH_VALUE_OF_EMPTY_STRING)) {
```
MikeDombo, line 610: flip this equals around. `!HASH_VALUE_OF_EMPTY_STRING.equals(fileHash)`

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -0,0 +1,100 @@
...
+    public static LogFile[] of(File... files) {
+        if (files == null) {
+            return new LogFile[] {};
+        }
+        return Arrays.stream(files).map(LogFile::of).toArray(LogFile[]::new);
+    }
+
+    public boolean isLogFile() {
+        return super.isFile();
+    }
```
MikeDombo, line 52: this name is confusing and not accurate. this method should just be called `isFile` if that's what it is doing. `isLogFile` sounds like it should only return true when it is a log file, this returns true if it is any file.

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * Read target lines from the file.
+     * The file must contain (minLine + 1) lines. One extra line is needed to prevent incomplete line when hashing.
+     * @return an ArrayList of lines or empty ArrayList
+     */
+    private List<String> getLines() {
+        List<String> linesRead = new ArrayList<>();
+        String oneLine;
+        try (BufferedReader r = Files.newBufferedReader(this.toPath(), StandardCharsets.UTF_8)) {
+            // read target number of lines
```
MikeDombo, line 63: Do NOT read the entire file into memory, this is going to cause us to run out of memory.

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -0,0 +1,100 @@
...
+            while (linesRead.size() < linesNeeded) {
+                oneLine = r.readLine();
+                if (oneLine == null) {
+                    break;
+                }
+                linesRead.add(oneLine);
+            }
+        } catch (IOException e) {
```
MikeDombo, line 71: if the file does not exist anymore, do not log an error at the error level. There's nothing we can do about it.

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+        return linesRead;
+    }
+
+    /**
+     * Get the hash of the logfile with target lines.
+     * @return the calculated hash value of the logfile
+     */
+    public String hashString() {
+        String fileHash = "";
```
MikeDombo, line 83: your constant `HASH_VALUE_OF_EMPTY_STRING` should be defined and used here.

src/test/java/com/aws/greengrass/logmanager/model/LogFileTest.java
```diff
@@ -0,0 +1,76 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+
+@ExtendWith({MockitoExtension.class, GGExtension.class})
+public class LogFileTest {
+
+    @TempDir
+    static Path directoryPath;
+    private final static int DEFAULT_LINES_FOR_DIGEST_NUM = 1;
+    private static final String HASH_VALUE_OF_EMPTY_STRING = "";
```
MikeDombo, line 27: you already have a constant, don't re-define it here, just use it.

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
dgggIoT: Changed it to Debug.

MikeDombo: Don't just change it to debug. Check if the error is `FileNotFoundException`. If it isn't that error, it is reasonable to log at error.

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public static LogFile[] of(File... files) {
+        if (files == null) {
+            return new LogFile[] {};
+        }
+        return Arrays.stream(files).map(LogFile::of).toArray(LogFile[]::new);
+    }
+
+    // name isLogFile instead of isFile to pass the checkstyle: PMD Overriding method merely calls super
+    public boolean isLogFile() {
```
dgggIoT, line 53: Need some suggestions here. If simply override ```isFile()``` will fail the PMD checkstyle as "Overriding method merely calls super".

MikeDombo: Then why does this method exist at all?

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -98,6 +100,7 @@ public class LogManagerService extends PluginService {
     public static final String MULTILINE_PATTERN_CONFIG_TOPIC_NAME = "multiLineStartPattern";
     public static final int DEFAULT_PERIODIC_UPDATE_INTERVAL_SEC = 300;
     private final Object spaceManagementLock = new Object();
+    public static final String HASH_VALUE_OF_EMPTY_STRING = "";
```
MikeDombo, line 103: move this constant, it does not belong here. This belongs in `LogFile`

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Read target lines from the file.
+     * The file must contain (minLine + 1) lines. One extra line is needed to prevent incomplete line when hashing.
+     * @return an ArrayList of lines or empty ArrayList
+     */
+    private List<String> getLines() {
+        List<String> linesRead = new ArrayList<>();
+        String oneLine;
```
MikeDombo, line 59: this variable is declared too early, you should declare it inside the while loop because that's the only place it is used.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -593,23 +602,29 @@ private void processLogsAndUpload() throws InterruptedException {
...
-                        // If the file was partially read in the previous run, then get the starting position for
-                        // new log lines.
-                        if (componentCurrentProcessingLogFile.containsKey(componentName)) {
-                            CurrentProcessingFileInformation processingFileInformation =
-                                    componentCurrentProcessingLogFile.get(componentName);
-                            if (processingFileInformation.fileName.equals(file.getAbsolutePath())
-                                    && processingFileInformation.lastModifiedTime == file.lastModified()) {
-                                startPosition = processingFileInformation.startPosition;
+                        String fileHash = file.hashString();
+                        // It must be a valid file for uploading
```
MikeDombo, line 609: this comment isn't clear. What makes a file valid or invalid for uploading?

MikeDombo (approved):

Nelsonochoam (approved):

PR #99, opened 2022-09-29
feat: read bytes rather than read lines when digest to avoid memory issue

dgggIoT:
**Issue #, if available:**
#51 
**Description of changes:**
Rather than using readLines to read the entire line, now we read amount of bytes (default 1KB) for digest. 
If the file does not have enough bytes but contains an entire line, we still digest it.
To avoid memory issue if one line is super large.
Unit test

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
+++ b/src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
@@ -4,7 +4,6 @@ import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.logmanager.LogManagerService;
 
-
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
@@ -13,30 +12,43 @@ import java.nio.file.Files;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
 
 import static com.aws.greengrass.util.Digest.calculate;
 
 public class LogFile extends File {
+    // custom serialVersionUID for class extends Serializable class
+    private static final long serialVersionUID = 123;
     private static final Logger logger = LogManager.getLogger(LogManagerService.class);
-    private static final String EMPTY_STRING = "";
-    private static final int linesNeededNum = 1;
+    private static final int linesNeeded = 1;
 
     public LogFile(String pathname) {
         super(pathname);
     }
 
+    /**
+     * Convert the file to LogFile.
+     * @param file The file to be converted.
+     * @return
+     */
     public static LogFile of(File file) {
         return new LogFile(file.getAbsolutePath());
     }
 
-    public static LogFile[] of(File[] files) {
-        if (files == null) return new LogFile[] {};
+    /**
+     * Convert list of files to list of LogFiles.
+     * @param files The list of files to be converted.
+     * @return
+     */
+    public static LogFile[] of(File... files) {
+        if (files == null) {
+            return new LogFile[] {};
+        }
         return Arrays.stream(files).map(LogFile::of).toArray(LogFile[]::new);
     }
 
-    @Override
-    public boolean isFile() {
+    public boolean isLogFile() {
         return super.isFile();
     }
 
@@ -45,29 +57,23 @@ public class LogFile extends File {
      * The file must contain (minLine + 1) lines. One extra line is needed to prevent incomplete line when hashing.
      * @return an ArrayList of lines or empty ArrayList
      */
-    private String getLines() {
-        ArrayList<String> linesReaded = new ArrayList<>();
+    private List<String> getLines() {
+        List<String> linesRead = new ArrayList<>();
         String oneLine;
         try (BufferedReader r = Files.newBufferedReader(this.toPath(), StandardCharsets.UTF_8)) {
-            int linesReadForDigest = 0;
             // read target number of lines
-            while (linesReadForDigest < linesNeededNum) {
+            while (linesRead.size() < linesNeeded) {
                 oneLine = r.readLine();
-                if (oneLine == null) break;
-                linesReadForDigest ++;
-                linesReaded.add(oneLine);
-            }
-            // read one more line to prevent incomplete last line
-            // if lines in file less than minLines + 1, then return empty
-            if (r.readLine() == null || linesReadForDigest < linesNeededNum) {
-                logger.atTrace().log("Not enough lines to digest file {} ", this.getAbsolutePath());
-                return EMPTY_STRING;
+                if (oneLine == null) {
+                    break;
+                }
+                linesRead.add(oneLine);
             }
         } catch (IOException e) {
             // File may not exist
             logger.atError().cause(e).log("Unable to read file {}", this.getAbsolutePath());
         }
-        return String.join("", linesReaded);
+        return linesRead;
     }
 
     /**
@@ -75,15 +81,20 @@ public class LogFile extends File {
      * @return the calculated hash value of the logfile
      */
     public Optional<String> hashString() {
-        if (!this.exists()) return Optional.ofNullable(null);
-        String lines = getLines();
-        if (!lines.isEmpty()) {
-            try {
-                return Optional.ofNullable(calculate(lines));
-            } catch (NoSuchAlgorithmException e) {
-                logger.atError().cause(e).log("The Digest Algorithm is invalid.");
-            }
+        if (!this.exists()) {
+            return Optional.empty();
         }
-        return Optional.ofNullable(null);
+        List<String> lines = getLines();
+        if (lines.size() < linesNeeded) {
+            return Optional.empty();
+        }
+
+        try {
+            return Optional.of(calculate(String.join("", lines)));
+        } catch (NoSuchAlgorithmException e) {
+            logger.atError().cause(e).log("The Digest Algorithm is invalid.");
+        }
+
+        return Optional.empty();
     }
 }
--- a/src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
+++ b/src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
@@ -13,7 +13,6 @@ import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Optional;
 
 import static com.aws.greengrass.util.Digest.calculate;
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -50,29 +51,21 @@ public static LogFile[] of(File... files) {
...
-                String oneLine = r.readLine();
-                if (oneLine == null) {
-                    break;
-                }
-                linesRead.add(oneLine);
-            }
+    private byte[] readBytes() {
+        byte[] bytesReadArray = new byte[bytesNeeded];
+        try (InputStream r = Files.newInputStream(this.toPath())) {
+            didRead = r.read(bytesReadArray);
```
Nelsonochoam, line 60:
```suggestion
            bytesRead = r.read(bytesReadArray);
```

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -85,16 +78,19 @@ public String hashString() {
             if (!this.exists()) {
                 return fileHash;
             }
-            List<String> lines = getLines();
-            if (lines.size() < linesNeeded) {
-                return fileHash;
+            String bytesReadString = new String(readBytes(), StandardCharsets.UTF_8);
+            // if the we read less then 1KB but it contains an entire line, it is still valid.
+            if (didRead < bytesNeeded && bytesReadString.indexOf('\n') > -1) {
```
Nelsonochoam, line 83: Use variable names that represent your intent

```suggestion
    boolean hasLessThanBytesNeeded = didRead < bytesNeeded
    boolean containsEntireLine = bytesReadString.indexOf('\n') > -1
    if (hasLessThanBytesNeeded && containsEntireLine) {
```

Nelsonochoam: What if the log file just has an empty line?, do we really care about it being a full line? - I think what we should care about is we have read "enough bytes" so we should read hopefully all the ones we require but if there are less than that guarantee that they are more than a minimum

src/test/java/com/aws/greengrass/logmanager/LogManagerServiceTest.java
```diff
@@ -135,13 +137,13 @@ static void setupBefore() throws IOException, InterruptedException {
             assertTrue(file.setWritable(true));
 
             try (OutputStream fileOutputStream = Files.newOutputStream(file.toPath())) {
-                fileOutputStream.write("TEST".getBytes(StandardCharsets.UTF_8));
+                fileOutputStream.write(getRandomTestStrings(bytesNeeded).getBytes(StandardCharsets.UTF_8));
```
Nelsonochoam, line 140: I like this

src/test/java/com/aws/greengrass/logmanager/LogManagerServiceTest.java
```diff
@@ -83,6 +83,8 @@
 import static com.aws.greengrass.logmanager.LogManagerService.SYSTEM_LOGS_COMPONENT_NAME;
 import static com.aws.greengrass.logmanager.LogManagerService.SYSTEM_LOGS_CONFIG_TOPIC_NAME;
 import static com.aws.greengrass.logmanager.LogManagerService.UPLOAD_TO_CW_CONFIG_TOPIC_NAME;
+import static com.aws.greengrass.logmanager.model.LogFile.bytesNeeded;
+import static com.aws.greengrass.logmanager.model.LogFileTest.getRandomTestStrings;
```
Nelsonochoam, line 87: This seems weird importing from another test. Maybe move that method to a LogFileTestUtils class inside the `test` folder

src/test/java/com/aws/greengrass/logmanager/model/LogFileTest.java
```diff
@@ -25,53 +28,86 @@ public class LogFileTest {
 
     @TempDir
     static Path directoryPath;
-    private final static int DEFAULT_LINES_FOR_DIGEST_NUM = 1;
+    private final static int DEFAULT_BYTES_FOR_DIGEST_NUM = 1024;
 
-    private void writeFiles(File file, int linesNeeded) throws IOException {
+    public static String getRandomTestStrings(int bytesNeeded) {
+        StringBuilder testStrings = new StringBuilder();
```
Nelsonochoam, line 34: The only bad thing about using this is it doesn't add ned lines hence doesn't simulate a real-ish scenario

src/test/java/com/aws/greengrass/logmanager/model/LogFileTest.java
```diff
@@ -25,53 +28,86 @@ public class LogFileTest {
...
             throws IOException, NoSuchAlgorithmException {
         File file = new File(directoryPath.resolve("greengrass_test.log").toUri());
-        writeFiles(file, DEFAULT_LINES_FOR_DIGEST_NUM + 1);
-        StringBuilder msg = new StringBuilder();
-        for (int i = 0; i < DEFAULT_LINES_FOR_DIGEST_NUM; i++) {
-            msg.append("line").append(i + 1);
-        }
+        byte[] bytesArray = getRandomTestStrings(DEFAULT_BYTES_FOR_DIGEST_NUM + 100).getBytes(StandardCharsets.UTF_8);
+        writeFiles(file, bytesArray);
+        LogFile logFile = LogFile.of(file);
```
Nelsonochoam, line 93: Why do this? You could create a LogFile directly since it is just a file. 

```
LogFile file = new LogFile(directoryPath.resolve("greengrass_test.log").toUri());
```

src/test/java/com/aws/greengrass/logmanager/model/LogFileTest.java
```diff
@@ -25,53 +28,86 @@ public class LogFileTest {
...
+        String fileHash = logFile.hashString();
+        String msg = new String(Arrays.copyOfRange(bytesArray, 0, DEFAULT_BYTES_FOR_DIGEST_NUM));
```
Nelsonochoam, line 95: Why do ew need to `copyOfRange`?  couldn't we do `new String(bytesArray, StandardCharsets.UTF_8)`

src/test/java/com/aws/greengrass/logmanager/model/LogFileTest.java
```diff
@@ -25,53 +28,86 @@ public class LogFileTest {
...
+        assertEquals(fileHash, calculate(msg));
+        file.delete();
+    }
+
+    @Test
+    void GIVEN_log_file_with_less_than_target_lines_but_has_new_line_WHEN_calculate_file_hash_THEN_we_get_hash()
+            throws IOException, NoSuchAlgorithmException {
+        File file = new File(directoryPath.resolve("greengrass_test.log").toUri());
+        StringBuilder testString = new StringBuilder(getRandomTestStrings(DEFAULT_BYTES_FOR_DIGEST_NUM - 100));
```
Nelsonochoam, line 104: Why use a `StringBuilder` here? We could just use the returned string

```java
        StringBuilder testString = getRandomTestStrings(DEFAULT_BYTES_FOR_DIGEST_NUM - 100);
        byte[] bytesArray = testString.getBytes(StandardCharsets.UTF_8);

```

src/test/java/com/aws/greengrass/logmanager/model/LogFileTest.java
```diff
@@ -25,53 +28,86 @@ public class LogFileTest {
...
+        testString.append(System.lineSeparator());
+        byte[] bytesArray = testString.toString().getBytes(StandardCharsets.UTF_8);
```
Nelsonochoam, line 106: I just find that reading this makes it hard to follow what you are doing we can make it simple and not be fancy like

```java
StringBuilder builder = new StringBuilder();
builder.append(getRandomTestStrings(DEFAULT_BYTES_FOR_DIGEST_NUM - 100));
builder.append(System.lineSeparator());

writeFile(file, builder.build()) // We can still pass a string here and let the write file convert it to an array
```

src/test/java/com/aws/greengrass/logmanager/model/LogFileTest.java
Nelsonochoam: Edit I saw how you are building the strings below

src/test/java/com/aws/greengrass/logmanager/model/LogFileTest.java
```diff
@@ -25,53 +28,86 @@ public class LogFileTest {
 
     @TempDir
     static Path directoryPath;
-    private final static int DEFAULT_LINES_FOR_DIGEST_NUM = 1;
+    private final static int DEFAULT_BYTES_FOR_DIGEST_NUM = 1024;
 
-    private void writeFiles(File file, int linesNeeded) throws IOException {
+    public static String getRandomTestStrings(int bytesNeeded) {
```
Nelsonochoam, line 33:
```suggestion
    public static String givenAStringOfSize(int bytesNeeded) {
```

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
MikeDombo, line 60: You should handle `didRead` here, rather than having it as a field of this class. I think that makes the API too complicated.

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -85,16 +78,19 @@ public String hashString() {
...
+                fileHash = calculate(bytesReadString.substring(0, bytesReadString.indexOf('\n') + 1));
+            }
+            if (didRead == bytesNeeded) {
```
MikeDombo, line 86: use `>=` just because

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
dgggIoT: I think there are two reasons:
1. The case of empty line is rare, because it is not a standard way to write logs.  
2. There may be the case that the log file only has content < 1KB, but we should not miss uploading these contents.

Nelsonochoam: 1) We should handle it and not make that assume it wont happen.

2) Why wouldn't we want to create a hash if we have less bytes than what we want but not a full line? - I think it is better to require at least a minimum instead of expecting a new line

dgggIoT: It is hard to define the minimum number. For example, the log file may only has one line, and it only contains 30 bytes. Then the file is rotated. In this example, we still need to upload this line since it is actually an valid log line. But it is hard for us to say 30 is the right number as the minimum number.

src/test/java/com/aws/greengrass/logmanager/LogManagerServiceTest.java
dgggIoT: Probably keep it for now, if we need more such utils method, then we can create such class?

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
Nelsonochoam: I think we are mixing 2 separate things here:

1) The how many bytes we need to compute a hash has nothing to do with the fact that the file might have a single line or multiple lines. To address your example we might have a file with a single line and not line break that is below the bytes you require that gets rotated and we still would have to upload it.

2) The other problem with this is that calculating the hash with this implementation is not deterministic - Depending on when you call `hashString` it might return different values
     2.1) If you call it and you have less than what you require it will return a value
     2.2) If the file then get more lines written to it and you can get all the bytes you need you will get a different hash.


We either have to memoize the hash value after you first compute it or have a firm required number of bytes you need.

dgggIoT: Agree. I will change it to the following:
1. If there is a new line before 1KB, then we take a new line
2. Otherwise, we read 1KB to prevent crazy long one line message.

MikeDombo (approved):

PR #101, opened 2022-10-10
feat: include active file in every cycle

dgggIoT:
**Issue #, if available:**
#51 
**Description of changes:**
Add the active file and fileHash into the current process

Unit test and Integrat tests are tested, but more needs to be added
This commit is not ready to merge. More integration and unit tests may be needed.
**Checklist:**
 - [ x] Updated or added new unit tests

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
@@ -51,6 +51,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.TimeZone;
+import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
@@ -96,15 +97,13 @@ class LogManagerTest extends BaseITCase {
 
     private String calculateLogStreamName(String thingName, String group) {
         synchronized (DATE_FORMATTER) {
-            return DEFAULT_LOG_STREAM_NAME
-                    .replace("{thingName}", thingName)
-                    .replace("{ggFleetId}", group)
+            return DEFAULT_LOG_STREAM_NAME.replace("{thingName}", thingName).replace("{ggFleetId}", group)
                     .replace("{date}", DATE_FORMATTER.format(new Date()));
         }
     }
 
-    void setupKernel(Path storeDirectory, String configFileName) throws InterruptedException,
-            URISyntaxException, IOException, DeviceConfigurationException {
+    void setupKernel(Path storeDirectory, String configFileName)
+            throws InterruptedException, URISyntaxException, IOException, DeviceConfigurationException {
 
         System.setProperty("root", tempRootDir.toAbsolutePath().toString());
         CountDownLatch logManagerRunning = new CountDownLatch(1);
@@ -114,22 +113,21 @@ class LogManagerTest extends BaseITCase {
 
         Path testRecipePath = Paths.get(LogManagerTest.class.getResource(configFileName).toURI());
         String content = new String(Files.readAllBytes(testRecipePath), StandardCharsets.UTF_8);
-        content = content.replaceAll("\\{\\{logFileDirectoryPath}}",
-                storeDirectory.toAbsolutePath().toString());
+        content = content.replaceAll("\\{\\{logFileDirectoryPath}}", storeDirectory.toAbsolutePath().toString());
 
         Map<String, Object> objectMap = YAML_OBJECT_MAPPER.readValue(content, Map.class);
         kernel.parseArgs();
         kernel.getConfig().mergeMap(System.currentTimeMillis(), ConfigPlatformResolver.resolvePlatformMap(objectMap));
 
         kernel.getContext().addGlobalStateChangeListener((service, oldState, newState) -> {
-            if (service.getName().equals(LogManagerService.LOGS_UPLOADER_SERVICE_TOPICS)
-                    && newState.equals(State.RUNNING)) {
+            if (service.getName().equals(LogManagerService.LOGS_UPLOADER_SERVICE_TOPICS) && newState.equals(State.RUNNING)) {
                 logManagerRunning.countDown();
                 logManagerService = (LogManagerService) service;
             }
         });
-        deviceConfiguration = new DeviceConfiguration(kernel, "ThingName", "xxxxxx-ats.iot.us-east-1.amazonaws.com", "xxxxxx.credentials.iot.us-east-1.amazonaws.com", "privKeyFilePath",
-                "certFilePath", "caFilePath", "us-east-1", "roleAliasName");
+        deviceConfiguration = new DeviceConfiguration(kernel, "ThingName", "xxxxxx-ats.iot.us-east-1.amazonaws.com",
+                "xxxxxx.credentials.iot.us-east-1.amazonaws.com", "privKeyFilePath", "certFilePath", "caFilePath",
+                "us-east-1", "roleAliasName");
 
         kernel.getContext().put(DeviceConfiguration.class, deviceConfiguration);
         // set required instances from context
@@ -154,17 +152,17 @@ class LogManagerTest extends BaseITCase {
     }
 
     @Test
-    void GIVEN_user_component_config_with_small_periodic_interval_WHEN_interval_elapses_THEN_logs_are_uploaded_to_cloud()
+    void GIVEN_user_component_config_with_small_periodic_interval_WHEN_interval_elapses_THEN_logs_are_uploaded_to_cloud_and_active_file_exists()
             throws Exception {
-        when(cloudWatchLogsClient.putLogEvents(any(PutLogEventsRequest.class)))
-                .thenReturn(PutLogEventsResponse.builder().nextSequenceToken("nextToken").build());
+        when(cloudWatchLogsClient.putLogEvents(any(PutLogEventsRequest.class))).thenReturn(
+                PutLogEventsResponse.builder().nextSequenceToken("nextToken").build());
 
         tempDirectoryPath = Files.createTempDirectory(tempRootDir, "IntegrationTestsTemporaryLogFiles");
 
         for (int i = 0; i < 5; i++) {
-            createTempFileAndWriteData(tempDirectoryPath, "integTestRandomLogFiles.log_",  "");
+            createTempFileAndWriteData(tempDirectoryPath, "integTestRandomLogFiles.log_", "");
         }
-        createTempFileAndWriteData(tempDirectoryPath, "integTestRandomLogFiles.log",  "");
+        createTempFileAndWriteData(tempDirectoryPath, "integTestRandomLogFiles.log", "");
 
         setupKernel(tempDirectoryPath, "smallPeriodicIntervalUserComponentConfig.yaml");
         TimeUnit.SECONDS.sleep(30);
@@ -174,12 +172,10 @@ class LogManagerTest extends BaseITCase {
         assertEquals(1, putLogEventsRequests.size());
         for (PutLogEventsRequest request : putLogEventsRequests) {
             assertEquals(calculateLogStreamName(THING_NAME, LOCAL_DEPLOYMENT_GROUP_NAME), request.logStreamName());
-            assertEquals("/aws/greengrass/UserComponent/" + AWS_REGION + "/UserComponentA",
-                    request.logGroupName());
+            assertEquals("/aws/greengrass/UserComponent/" + AWS_REGION + "/UserComponentA", request.logGroupName());
             assertNotNull(request.logEvents());
-            assertEquals(50, request.logEvents().size());
-            assertEquals(51200, request.logEvents().stream().mapToLong(value -> value.message().length())
-                    .sum());
+            assertEquals(60, request.logEvents().size());
+            assertEquals(61440, request.logEvents().stream().mapToLong(value -> value.message().length()).sum());
         }
         File folder = tempDirectoryPath.toFile();
         Pattern logFileNamePattern = Pattern.compile("^integTestRandomLogFiles.log\\w*");
@@ -187,9 +183,7 @@ class LogManagerTest extends BaseITCase {
         File[] files = folder.listFiles();
         if (files != null) {
             for (File file : files) {
-                if (file.isFile()
-                        && logFileNamePattern.matcher(file.getName()).find()
-                        && file.length() > 0) {
+                if (file.isFile() && logFileNamePattern.matcher(file.getName()).find() && file.length() > 0) {
                     allFiles.add(file);
                 }
             }
@@ -198,10 +192,10 @@ class LogManagerTest extends BaseITCase {
     }
 
     @Test
-    void GIVEN_user_component_config_with_small_periodic_interval_and_only_required_config_WHEN_interval_elapses_THEN_logs_are_uploaded_to_cloud()
+    void GIVEN_user_component_config_with_small_periodic_interval_and_only_required_config_WHEN_interval_elapses_THEN_logs_are_uploaded_to_cloud_and_active_file_exists()
             throws Exception {
-        when(cloudWatchLogsClient.putLogEvents(any(PutLogEventsRequest.class)))
-                .thenReturn(PutLogEventsResponse.builder().nextSequenceToken("nextToken").build());
+        when(cloudWatchLogsClient.putLogEvents(any(PutLogEventsRequest.class))).thenReturn(
+                PutLogEventsResponse.builder().nextSequenceToken("nextToken").build());
 
         tempDirectoryPath = Files.createDirectory(tempRootDir.resolve("logs"));
         LogConfig.getRootLogConfig().setLevel(Level.TRACE);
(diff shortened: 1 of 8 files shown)


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -585,54 +594,66 @@ private void processLogsAndUpload() throws InterruptedException {
...
                     break;
                 } catch (SecurityException e) {
                     logger.atError().cause(e).log("Unable to get log files for {} from {}",
                             componentName, componentLogConfiguration.getDirectoryPath());
                 }
             }
-            if (componentLogFileInformation.get().isPresent()) {
+            if (!hasOnlyActiveFileButUploaded && componentLogFileInformation.get().isPresent()) {
+                // This is only for passing the integrate test.
+                TimeUnit.SECONDS.sleep(1);
```
dgggIoT, line 656: This is only for passing the integration test. When not sleeping, the integration tests will receive infinite following errors. Please let me know how I can solve it. 
```
com.aws.greengrass.lifecyclemanager.KernelExceptionHandler: uncaught-exception. {thread=Thread[pool-2-thread-8,5,main]}
software.amazon.awssdk.crt.CrtRuntimeException: aws_tls_ctx_options_init_client_mtls_from_path failed (aws_last_error: AWS_ERROR_FILE_INVALID_PATH(44), Invalid file path.)
```

src/test/java/com/aws/greengrass/logmanager/LogManagerServiceTest.java
```diff
@@ -842,6 +856,14 @@ void GIVEN_user_component_logs_delete_file_after_upload_set_WHEN_upload_logs_THE
...
 
         logsUploaderService = new LogManagerService(config, mockUploader, mockMerger, executor);
+        logsUploaderService.fileHashNamePairs.putIfAbsent(fileHash1, file1.getAbsolutePath());
+        if (!logsUploaderService.fileHashNamePairs.get(fileHash1).equals(file1.getAbsolutePath())) {
+            logsUploaderService.fileHashNamePairs.replace(fileHash1, file1.getAbsolutePath());
+        }
+        logsUploaderService.fileHashNamePairs.putIfAbsent(fileHash2, file2.getAbsolutePath());
+        if (!logsUploaderService.fileHashNamePairs.get(fileHash2).equals(file2.getAbsolutePath())) {
+            logsUploaderService.fileHashNamePairs.replace(fileHash2, file2.getAbsolutePath());
+        }
```
MikeDombo, line 866: what is the purpose of this code? Please explain in a comment.

We shouldn't really be storing data into the service, right? The log manager should figure out the hashes and pairs correctly itself.

src/test/java/com/aws/greengrass/logmanager/LogManagerServiceTest.java
```diff
@@ -588,15 +589,17 @@ void GIVEN_cloud_watch_attempt_handler_WHEN_attempt_completes_THEN_successfully_
         CloudWatchAttempt attempt = new CloudWatchAttempt();
         Map<String, CloudWatchAttemptLogInformation> logStreamsToLogInformationMap = new HashMap<>();
         File file1 = new File(getClass().getResource("testlogs2.log").toURI());
+        String filehash1 = "123456";
         File file2 = new File(getClass().getResource("testlogs1.log").toURI());
+        String filehash2 = "987654";
```
MikeDombo, line 594: this definitely isn't the hash value for this log file, so why is this being used?

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -174,22 +172,18 @@ void GIVEN_user_component_config_with_small_periodic_interval_WHEN_interval_elap
...
             assertEquals(calculateLogStreamName(THING_NAME, LOCAL_DEPLOYMENT_GROUP_NAME), request.logStreamName());
-            assertEquals("/aws/greengrass/UserComponent/" + AWS_REGION + "/UserComponentA",
-                    request.logGroupName());
+            assertEquals("/aws/greengrass/UserComponent/" + AWS_REGION + "/UserComponentA", request.logGroupName());
             assertNotNull(request.logEvents());
-            assertEquals(50, request.logEvents().size());
-            assertEquals(51200, request.logEvents().stream().mapToLong(value -> value.message().length())
-                    .sum());
+            assertEquals(60, request.logEvents().size());
+            assertEquals(61440, request.logEvents().stream().mapToLong(value -> value.message().length()).sum());
```
MikeDombo, line 178: why is the size and number of events changed?

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -350,4 +335,51 @@ void GIVEN_log_manager_in_errored_state_WHEN_restarted_THEN_logs_upload_is_reatt
         }
         assertEquals(1, allFiles.size());
     }
-}
+
+    @Test
```
MikeDombo, line 339: https://www.baeldung.com/junit-5-repeated-test

Let's repeat this several times to make sure that we are doing the right thing, as much as we can.

src/integrationtests/resources/com/aws/greengrass/integrationtests/logmanager/randomlyNamedFilesWithoutComponentsConfig.yaml
```diff
@@ -0,0 +1,20 @@
...
+      periodicUploadIntervalSec: 10
+      logsUploaderConfiguration:
+        componentLogsConfiguration:
+          - componentName: 'UserComponentB'
+            logFileRegex: '\w*UserComponentB\w*.\w*'
+            logFileDirectoryPath: '{{logFileDirectoryPath}}'
+            minimumLogLevel: 'INFO'
+            diskSpaceLimit: '25'
+            diskSpaceLimitUnit: 'MB'
+            deleteLogFileAfterCloudUpload: 'true'
```
MikeDombo, line 14: don't delete after upload. I want to ensure that the log files are not uploading duplicates, even if their name changes (after being uploaded)

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -350,4 +335,51 @@ void GIVEN_log_manager_in_errored_state_WHEN_restarted_THEN_logs_upload_is_reatt
...
+                PutLogEventsResponse.builder().nextSequenceToken("nextToken").build());
+
+        tempDirectoryPath = Files.createDirectory(tempRootDir.resolve("logs"));
+        LogConfig.getRootLogConfig().setLevel(Level.TRACE);
+        LogConfig.getRootLogConfig().setStore(LogStore.FILE);
+        LogManager.getLogConfigurations().putIfAbsent("UserComponentB",
+                newLogConfigFromRootConfig(LogConfigUpdate.builder().fileName("UserComponentB.log").build()));
+
+        for (int i = 0; i < 6; i++) {
+            String randomName1 = UUID.randomUUID().toString();
```
MikeDombo, line 352: I don't just want the files to be randomly named and then uploaded. I want that, and then I want the files to be randomly re-named and verify that we _do not_ upload them again.

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -128,9 +128,12 @@ public CloudWatchAttempt processLogFiles(ComponentLogFileInformation componentLo
         while (!componentLogFileInformation.getLogFileInformationList().isEmpty() && !reachedMaxSize.get()) {
             LogFile logFile = componentLogFileInformation.getLogFileInformationList().get(0).getLogFile();
```
MikeDombo, line 129: we should stop calling `get(0)`, just get the value once, and store it in a temporary variable.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -499,26 +476,63 @@ private void handleCloudWatchAttemptStatus(CloudWatchAttempt cloudWatchAttempt)
...
+            String key = entry.getKey();
+            Set<String> fileNames = entry.getValue();
+            fileNames.stream().map(LogFile::new).forEach(file -> {
+                if (!lastComponentUploadedLogFileInstantMap.containsKey(key)
+                        || lastComponentUploadedLogFileInstantMap.get(key)
+                        .isBefore(Instant.ofEpochMilli(file.lastModified()))) {
+                    lastComponentUploadedLogFileInstantMap.put(key, Instant.ofEpochMilli(file.lastModified()));
+                }
+                // delete the entry if one file is read and uploaded completely
+                fileHashNamePairs.remove(file.hashString());
```
MikeDombo, line 490: if we remove it here, but do _not_ delete the file from the disk, won't we upload it again? We wouldn't want to do that, right?

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
MikeDombo: Sleeping shouldn't matter. This error is pretty clear, it should be handled properly in the upload method.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -585,25 +601,25 @@ private void processLogsAndUpload() throws InterruptedException {
                             }
                         }
                     }
+                    // If there are no log files for the component, then return.
+                    if (allFiles.isEmpty()) {
+                        continue;
+                    }
                     // Sort the files by the last modified time. Then try to proceed oldest file first to avoid data
                     // loss caused by auto deletion
                     allFiles.sort(Comparator.comparingLong(LogFile::lastModified));
```
MikeDombo, line 610: why do we need any of this logic anymore? We want to treat all files exactly the same now, don't we?

src/test/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessorTest.java
```diff
@@ -536,35 +536,34 @@ void GIVEN_unstructured_log_WHEN_breaches_event_size_and_batch_size_limits_THEN_
...
+                    logEventsForStream1.getAttemptLogFileInformationMap().get(fileHash).getBytesRead());
             assertEquals("TestComponent", logEventsForStream1.getComponentName());
 
         } finally {
-            assertTrue(file.delete());
+            assertTrue(logFile.delete());
         }
     }
     @Test
-    void GIVEN_null_log_message_WHEN_upload_attempted_THEN_null_message_considered_as_unstructured_log(ExtensionContext ec) throws IOException {
```
MikeDombo, line 551: this test is now testing the exact opposite of what it was testing before. Why are we changing the behavior? Is the behavior change correct?

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
dgggIoT: Because the active log file is also included

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -585,25 +601,25 @@ private void processLogsAndUpload() throws InterruptedException {
...
-                    // If there are no rotated log files for the component, then return.
-                    if (allFiles.size() - 1 <= 0) {
-                        continue;
-                    }
-                    // Don't consider the active log file.
-                    allFiles = allFiles.subList(0, allFiles.size() - 1);
+                    String activeFileName = allFiles.get(allFiles.size() - 1).getName();
```
Nelsonochoam, line 611: I'd rather call this again that storing the activeFileName. I generally want to try to avoid relying as much as possible on file names given we can't be sure how the customer is changing or doing with the names. Instead of storing this on the `ComponentLogFileInformation` has some construct like `LogDirectory` that you can ask it for the active file rather than relying on names to determine if something is active.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -613,26 +629,49 @@ private void processLogsAndUpload() throws InterruptedException {
                             if (componentCurrentProcessingLogFile.containsKey(componentName)) {
                                 CurrentProcessingFileInformation processingFileInformation =
                                         componentCurrentProcessingLogFile.get(componentName);
-                                if (processingFileInformation.fileName.equals(file.getAbsolutePath())
-                                        && processingFileInformation.lastModifiedTime == file.lastModified()) {
+                                //if (processingFileInformation.fileName.equals(file.getAbsolutePath())
```
Nelsonochoam, line 632: Why leave this commented here?

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -585,25 +601,25 @@ private void processLogsAndUpload() throws InterruptedException {
                             }
                         }
                     }
+                    // If there are no log files for the component, then return.
+                    if (allFiles.isEmpty()) {
```
Nelsonochoam, line 605: This should not be needed. The only reason you have to do this is because now we are relying on it always having at least 1 because you are using it to get the activeFileName

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -613,26 +629,49 @@ private void processLogsAndUpload() throws InterruptedException {
...
-                            LogFileInformation logFileInformation = LogFileInformation.builder()
-                                            .logFile(file)
-                                            .startPosition(startPosition)
-                                            .fileHash(fileHash)
-                                            .build();
+
+                            // If the file is not completed read or the file is recorded as
+                            // update the global map of hash and name pair
+                            fileHashNamePairs.putIfAbsent(fileHash, file.getAbsolutePath());
+                            if (!fileHashNamePairs.get(fileHash).equals(file.getAbsolutePath())) {
```
Nelsonochoam, line 642: Seems unnecessary - We are saying if the absolute path has changed for this file has that I have stored. Update the value. Why not just always update it?

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -613,26 +629,49 @@ private void processLogsAndUpload() throws InterruptedException {
...
+                                fileHashNamePairs.replace(fileHash, file.getAbsolutePath());
+                            }
+
+                            // if the file is uploaded and it is not active file, means it rotates, thus delete it
+                            // if the file is uploaded and it is active file, means all files are uploaded
+                            if (startPosition == file.length()) {
```
Nelsonochoam, line 648: Use names that represent intent rather than comments. 

```java
boolean wasFileUploaded = startPosition == file.length();
boolean isNotActiveFile = !file.equals(activeFile);

if (wasFileUploaded && isNotActiveFile) {
}
```

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
MikeDombo: If this is only used for deletion, then it should be filtered in the deletion logic. This logic doesn't sound like it is required right here.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -499,26 +476,63 @@ private void handleCloudWatchAttemptStatus(CloudWatchAttempt cloudWatchAttempt)
...
     private void processCloudWatchAttemptLogInformation(Map<String, Set<String>> completedLogFilePerComponent,
                                                         Map<String, CurrentProcessingFileInformation>
                                                                 currentProcessingLogFilePerComponent,
                                                         CloudWatchAttemptLogInformation attemptLogInformation,
-                                                        String fileName,
+                                                        String fileHash,
                                                         CloudWatchAttemptLogFileInformation
                                                                 cloudWatchAttemptLogFileInformation) {
-        File file = new File(fileName);
+        String fileName = fileHashNamePairs.get(fileHash);
```
Nelsonochoam, line 521: I thought one benefit of using the hash was so that we did not rely anymore on fileNames... It might happen as we saw that the fileNames might not change but the hash might (depending on the rotation strategy). So we might get a scenario where we have multiple hashes pointing to the same file

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -117,6 +119,8 @@ public class LogManagerService extends PluginService {
     @Getter
     private int periodicUpdateIntervalSec;
     private Future<?> spaceManagementThread;
+    // public only for integ tests
+    public ConcurrentHashMap<String, String> fileHashNamePairs = new ConcurrentHashMap<>();
```
Nelsonochoam, line 123: I don't think we should rely on file names

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -499,26 +476,63 @@ private void handleCloudWatchAttemptStatus(CloudWatchAttempt cloudWatchAttempt)
         isCurrentlyUploading.set(false);
     }
 
+    private void deleteFilesInComponents(Map<String, Set<String>> fileDeleteSet) {
+        for (Map.Entry<String, Set<String>> entry : fileDeleteSet.entrySet()) {
```
Nelsonochoam, line 480: Why not pass a Set directly?

PR #102, opened 2022-10-12
feat: add LogFileGroup to find the active file

dgggIoT:
**Issue #, if available:**
#51
**Description of changes:**
Introduce the LogFileGroup to grab the list of logFiles and can use lastModifiedTime to get the active file
THis is an approach to totally get rid of filenames.
Not yet
This PR is using enabledFlag to make current logic not changed and pass the Tests. 
**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
+++ b/src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
@@ -14,6 +14,8 @@ import com.aws.greengrass.logmanager.model.CloudWatchAttemptLogFileInformation;
 import com.aws.greengrass.logmanager.model.CloudWatchAttemptLogInformation;
 import com.aws.greengrass.logmanager.model.ComponentLogFileInformation;
 import com.aws.greengrass.logmanager.model.LogFile;
+import com.aws.greengrass.logmanager.model.LogFileGroup;
+import com.aws.greengrass.logmanager.model.LogFileInformation;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.Pair;
 import com.aws.greengrass.util.Utils;
@@ -122,15 +124,20 @@ public class CloudWatchAttemptLogsProcessor {
                 .replace("{componentName}", componentLogFileInformation.getName());
         attempt.setLogGroupName(logGroupName);
         String logStreamName = getLogStreamName(thingName);
-
+        LogFileGroup logFileGroup = componentLogFileInformation.getLogFileGroup();
         // Run the loop until all the log files from the component have been read or the max message
         // size has been reached.
         while (!componentLogFileInformation.getLogFileInformationList().isEmpty() && !reachedMaxSize.get()) {
-            LogFile logFile = componentLogFileInformation.getLogFileInformationList().get(0).getLogFile();
-            long startPosition = componentLogFileInformation.getLogFileInformationList().get(0).getStartPosition();
-            //TODO Optional<String> fileHash = componentLogFileInformation.getLogFileInformationList().get(0)
-            // .getFileHash();
+            LogFileInformation tempLogFileInfomation = componentLogFileInformation.getLogFileInformationList().get(0);
+            LogFile logFile = tempLogFileInfomation.getLogFile();
+            long startPosition = tempLogFileInfomation.getStartPosition();
+            String fileHash = tempLogFileInfomation.getFileHash();
             String fileName = logFile.getAbsolutePath();
+            if (Utils.isEmpty(fileHash)) {
+                componentLogFileInformation.getLogFileInformationList().remove(0);
+                continue;
+            }
+
             long lastModified = logFile.lastModified();
 
             // If we have read the file already, we are at the correct offset in the file to start reading from
@@ -156,7 +163,7 @@ public class CloudWatchAttemptLogsProcessor {
                             reachedMaxSize.set(processLogLine(totalBytesRead,
                                     componentLogFileInformation.getDesiredLogLevel(), logStreamName,
                                     logStreamsMap, data, fileName, startPosition, componentLogFileInformation.getName(),
-                                    tempStartPosition, lastModified));
+                                    tempStartPosition, lastModified, logFileGroup));
                             componentLogFileInformation.getLogFileInformationList().remove(0);
                             break;
                         }
@@ -169,7 +176,7 @@ public class CloudWatchAttemptLogsProcessor {
                             reachedMaxSize.set(processLogLine(totalBytesRead,
                                     componentLogFileInformation.getDesiredLogLevel(), logStreamName,
                                     logStreamsMap, data, fileName, startPosition, componentLogFileInformation.getName(),
-                                    tempStartPosition, lastModified));
+                                    tempStartPosition, lastModified, logFileGroup));
                             data = new StringBuilder();
                         }
 
@@ -218,7 +225,8 @@ public class CloudWatchAttemptLogsProcessor {
                                    long startPosition,
                                    String componentName,
                                    long currentPosition,
-                                   long lastModified) {
+                                   long lastModified,
+                                   LogFileGroup logFileGroup) {
         String dataStr = data.toString();
         int dataSize = dataStr.getBytes(StandardCharsets.UTF_8).length;
         CloudWatchAttemptLogInformation attemptLogInformation;
@@ -233,6 +241,7 @@ public class CloudWatchAttemptLogsProcessor {
             attemptLogInformation = logStreamsMap.computeIfAbsent(logStreamName,
                     key -> CloudWatchAttemptLogInformation.builder()
                             .componentName(componentName)
+                            .logFileGroup(logFileGroup)
                             .build());
             addEventResult = checkAndAddNewLogEvent(totalBytesRead, attemptLogInformation,
                     dataStr, dataSize, desiredLogLevel, logMessage.get());
@@ -256,6 +265,7 @@ public class CloudWatchAttemptLogsProcessor {
             attemptLogInformation = logStreamsMap.computeIfAbsent(logStreamName,
                     key -> CloudWatchAttemptLogInformation.builder()
                             .componentName(componentName)
+                            .logFileGroup(logFileGroup)
                             .build());
             addEventResult =
                     addNewLogEvent(totalBytesRead, attemptLogInformation, dataStr, dataSize, logTimestamp);
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -21,6 +21,7 @@ import com.aws.greengrass.logmanager.model.ComponentLogConfiguration;
 import com.aws.greengrass.logmanager.model.ComponentLogFileInformation;
 import com.aws.greengrass.logmanager.model.ComponentType;
 import com.aws.greengrass.logmanager.model.LogFile;
+import com.aws.greengrass.logmanager.model.LogFileGroup;
 import com.aws.greengrass.logmanager.model.LogFileInformation;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.Utils;
@@ -506,11 +507,14 @@ public class LogManagerService extends PluginService {
                                                         String fileName,
                                                         CloudWatchAttemptLogFileInformation
                                                                 cloudWatchAttemptLogFileInformation) {
-        File file = new File(fileName);
+        LogFile file = new LogFile(fileName);
         String componentName = attemptLogInformation.getComponentName();
+        LogFileGroup logFileGroup = attemptLogInformation.getLogFileGroup();
+        // TODO: the following logic is only for passing this small PR while not changing the current context
+        boolean isActiveFile = logFileGroup.isActiveFile("");
         // If we have completely read the file, then we need add it to the completed files list and remove it
         // it (if necessary) for the current processing list.
-        if (file.length() == cloudWatchAttemptLogFileInformation.getBytesRead()
+        if (!isActiveFile && file.length() == cloudWatchAttemptLogFileInformation.getBytesRead()
                 + cloudWatchAttemptLogFileInformation.getStartPosition()) {
             Set<String> completedFileNames = completedLogFilePerComponent.getOrDefault(componentName, new HashSet<>());
             completedFileNames.add(fileName);
@@ -570,40 +574,25 @@ public class LogManagerService extends PluginService {
                 Instant lastUploadedLogFileTimeMs =
                         lastComponentUploadedLogFileInstantMap.getOrDefault(componentName,
                                 Instant.EPOCH);
-                File folder = new File(componentLogConfiguration.getDirectoryPath().toUri());
-                List<LogFile> allFiles = new ArrayList<>();
 
                 try {
-                    LogFile[] files = LogFile.of(folder.listFiles());
-                    if (files.length != 0) {
-                        for (LogFile file : files) {
-                            if (file.isFile()
-                                    && lastUploadedLogFileTimeMs.isBefore(Instant.ofEpochMilli(file.lastModified()))
-                                    && componentLogConfiguration.getFileNameRegex().matcher(file.getName()).find()
-                                    && file.length() > 0) {
-                                allFiles.add(file);
-                            }
-                        }
-                    }
-                    // Sort the files by the last modified time. Then try to proceed oldest file first to avoid data
-                    // loss caused by auto deletion
-                    allFiles.sort(Comparator.comparingLong(LogFile::lastModified));
-                    // If there are no rotated log files for the component, then return.
-                    if (allFiles.size() - 1 <= 0) {
+                    LogFileGroup logFileGroup =
+                            LogFileGroup.create(componentLogConfiguration.getFileNameRegex(),
+                                    componentLogConfiguration.getDirectoryPath().toUri(), lastUploadedLogFileTimeMs);
+                    if (logFileGroup.getLogFiles().isEmpty()) {
                         continue;
                     }
-                    // Don't consider the active log file.
-                    allFiles = allFiles.subList(0, allFiles.size() - 1);
-
+                    
                     componentLogFileInformation.set(Optional.of(
                             ComponentLogFileInformation.builder()
                                     .name(componentName)
                                     .multiLineStartPattern(componentLogConfiguration.getMultiLineStartPattern())
                                     .desiredLogLevel(componentLogConfiguration.getMinimumLogLevel())
                                     .componentType(componentLogConfiguration.getComponentType())
+                                    .logFileGroup(logFileGroup)
                                     .build()));
 
-                    allFiles.forEach(file -> {
+                    logFileGroup.getLogFiles().forEach(file -> {
                         long startPosition = 0;
                         String fileHash = file.hashString();
                         // The file must contain enough lines for digest hash, otherwise fileHash is empty string
--- a/src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
+++ b/src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
@@ -1,0 +1,58 @@
+package com.aws.greengrass.logmanager.model;
+
+import com.aws.greengrass.util.Utils;
+import lombok.Getter;
+
+import java.io.File;
+import java.net.URI;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.regex.Pattern;
+
+public class LogFileGroup {
+    @Getter
+    List<LogFile> logFiles;
+
+    private LogFileGroup(List<LogFile> files) {
+        this.logFiles = files;
+    }
+
+    public static LogFileGroup create(Pattern filePattern, URI path, Instant lastUpdated) {
+        File folder = new File(path);
+
+        List<LogFile> allFiles = new ArrayList<>();
+
+        if (!folder.isDirectory()) {
+            throw new RuntimeException("Must be a folder.");
+        }
+
+        LogFile[] files = LogFile.of(folder.listFiles());
+        if (files.length != 0) {
+            for (LogFile file: files) {
+                if (file.isFile()
+                        && lastUpdated.isBefore(Instant.ofEpochMilli(file.lastModified()))
+                        && filePattern.matcher(file.getName()).find()
+                        && file.length() > 0) {
+                    allFiles.add(file);
+                }
+            }
+        }
+        allFiles.sort(Comparator.comparingLong(LogFile::lastModified));
+        //TODO: the following logic is only for ignoring the active file and keeping the PR small
+        if (allFiles.size() - 1 <= 0) {
+            return new LogFileGroup(new ArrayList<>());
+        }
+        allFiles = allFiles.subList(0, allFiles.size() - 1);
+        return new LogFileGroup(allFiles);
+    }
+
+    public boolean isActiveFile(String fileHash) {
+        if (Utils.isEmpty(fileHash)) {
+            return false;
+        }
+        LogFile activeFile = logFiles.get(logFiles.size() - 1);
+        return activeFile.hashString().equals(fileHash);
+    }
+}
--- a/src/main/java/com/aws/greengrass/logmanager/model/CloudWatchAttemptLogInformation.java
+++ b/src/main/java/com/aws/greengrass/logmanager/model/CloudWatchAttemptLogInformation.java
@@ -28,6 +28,7 @@ public class CloudWatchAttemptLogInformation {
     @Builder.Default
     private Map<String, CloudWatchAttemptLogFileInformation> attemptLogFileInformationMap = new HashMap<>();
     private String componentName;
+    private LogFileGroup logFileGroup;
 
     /**
      * Get the log events in chronological order.
--- a/src/main/java/com/aws/greengrass/logmanager/model/ComponentLogFileInformation.java
+++ b/src/main/java/com/aws/greengrass/logmanager/model/ComponentLogFileInformation.java
@@ -24,4 +24,5 @@ public class ComponentLogFileInformation {
     private Pattern multiLineStartPattern;
     private Level desiredLogLevel;
     private ComponentType componentType;
+    private LogFileGroup logFileGroup;
 }


src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -122,15 +124,20 @@ public CloudWatchAttempt processLogFiles(ComponentLogFileInformation componentLo
...
-            LogFile logFile = componentLogFileInformation.getLogFileInformationList().get(0).getLogFile();
-            long startPosition = componentLogFileInformation.getLogFileInformationList().get(0).getStartPosition();
-            //TODO Optional<String> fileHash = componentLogFileInformation.getLogFileInformationList().get(0)
-            // .getFileHash();
+            LogFileInformation tempLogFileInfomation = componentLogFileInformation.getLogFileInformationList().get(0);
+            LogFile logFile = tempLogFileInfomation.getLogFile();
+            long startPosition = tempLogFileInfomation.getStartPosition();
+            String fileHash = tempLogFileInfomation.getFileHash();
             String fileName = logFile.getAbsolutePath();
+            if ("".equals(fileHash)) {
```
Nelsonochoam, line 136: Avoid exposing this. Do instead

```java
if (logFile.hasContents()) { // that logic would be inside the log file, then you are giving more meaning to it and also we can change our hashing strategy later on.
}
```

Nelsonochoam: This scenario shouldn't even happen. Add a TODO here to ensure that we only ever add files that have content to the `componentLogFileInformation`  the we don't have to be mutating the list here

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.regex.Pattern;
+
+import static com.aws.greengrass.logmanager.LogManagerService.ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG;
+
+public final class LogFileGroup {
+    @Getter
+    List<LogFile> logFiles;
```
Nelsonochoam, line 18: Not good don't make it public enforce that callers have to tell this Group what to do on each of the stored files. This breaks object orientation https://martinfowler.com/bliki/TellDontAsk.html

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -0,0 +1,72 @@
...
+     * @param filePattern the fileNameRegex used for each component to recognize its log files.
+     * @param path the directory path of the log files of component.
+     * @param lastUpdated the saved updated time of the last uploaded log of a component.
+     * @return list of logFile.
+     */
+    public static LogFileGroup create(Pattern filePattern, URI path, Instant lastUpdated) {
+        File folder = new File(path);
+
+        if (!folder.isDirectory()) {
+            throw new RuntimeException("Must be a folder.");
```
Nelsonochoam, line 35: Maybe don't use a RuntimeException and have a custom checked exception like `InvalidLogGroupException`

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -0,0 +1,72 @@
...
+                if (file.isFile()
+                        && lastUpdated.isBefore(Instant.ofEpochMilli(file.lastModified()))
+                        && filePattern.matcher(file.getName()).find()
+                        && file.length() > 0) {
+                    allFiles.add(file);
+                }
+            }
+        }
+        allFiles.sort(Comparator.comparingLong(LogFile::lastModified));
+        if (!ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG) {
```
Nelsonochoam, line 51: ADD A BIG COMMENT HERE STATING you are just doing this so we can add the change incrementally without having to change all the tests yet and avoid a massive PR. Also mention this will go away but we want to limit the size of the change on the PR

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+        return new LogFileGroup(allFiles);
+    }
+
+    /**
+     * Given a fileHash of a file, using the sorted log files to detect if the file is the active file.
+     * @param fileHash the fileHash of the target file.
+     * @return boolean if this is the active file.
+     */
+    public boolean isActiveFile(String fileHash) {
```
Nelsonochoam, line 65: Just pass the entire LogFile and have an have a method like

```java
logFile.equals(activeFile)
```

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -570,40 +580,25 @@ private void processLogsAndUpload() throws InterruptedException {
...
                             ComponentLogFileInformation.builder()
                                     .name(componentName)
                                     .multiLineStartPattern(componentLogConfiguration.getMultiLineStartPattern())
                                     .desiredLogLevel(componentLogConfiguration.getMinimumLogLevel())
                                     .componentType(componentLogConfiguration.getComponentType())
+                                    .logFileGroup(logFileGroup)
                                     .build()));
 
-                    allFiles.forEach(file -> {
+                    logFileGroup.getLogFiles().forEach(file -> {
```
Nelsonochoam, line 601: Add a `forEach` method inside the logGroup

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -570,40 +580,25 @@ private void processLogsAndUpload() throws InterruptedException {
...
-                    }
-                    // Sort the files by the last modified time. Then try to proceed oldest file first to avoid data
-                    // loss caused by auto deletion
-                    allFiles.sort(Comparator.comparingLong(LogFile::lastModified));
-                    // If there are no rotated log files for the component, then return.
-                    if (allFiles.size() - 1 <= 0) {
+                    LogFileGroup logFileGroup =
+                            LogFileGroup.create(componentLogConfiguration.getFileNameRegex(),
+                                    componentLogConfiguration.getDirectoryPath().toUri(), lastUploadedLogFileTimeMs);
+                    if (logFileGroup.getLogFiles().isEmpty()) {
```
Nelsonochoam, line 588: Add an `isEmpty` method inside the log Group do not add a `getLogFiles` method

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -506,11 +510,17 @@ private void processCloudWatchAttemptLogInformation(Map<String, Set<String>> com
                                                         String fileName,
                                                         CloudWatchAttemptLogFileInformation
                                                                 cloudWatchAttemptLogFileInformation) {
-        File file = new File(fileName);
+        LogFile file = new LogFile(fileName);
         String componentName = attemptLogInformation.getComponentName();
+        boolean isActiveFile = false;
+        if (ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG) {
```
Nelsonochoam, line 516: ADD A BIG TODO EXPLAINING THAT THIS IS JUST A TEMPORARY THING TO AVOID CHANGING a bunch of tests and additional code

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -506,11 +510,17 @@ private void processCloudWatchAttemptLogInformation(Map<String, Set<String>> com
...
+            LogFileGroup logFileGroup = attemptLogInformation.getLogFileGroup();
+            // TODO: the following logic is only for passing this small PR while not changing the current context
+            isActiveFile = logFileGroup.isActiveFile("");
```
Nelsonochoam, line 519: Any reason we would want this in here. It would always return `false` right?

dgggIoT: We have not got the fileHash in this commit yet, so leave "" instead.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -350,4 +350,52 @@ void GIVEN_log_manager_in_errored_state_WHEN_restarted_THEN_logs_upload_is_reatt
...
+
+        List<PutLogEventsRequest> putLogEventsRequests = captor.getAllValues();
+        assertEquals(1, putLogEventsRequests.size());
+        for (PutLogEventsRequest request : putLogEventsRequests) {
+            assertEquals(calculateLogStreamName(THING_NAME, LOCAL_DEPLOYMENT_GROUP_NAME), request.logStreamName());
+            assertEquals("/aws/greengrass/UserComponent/" + AWS_REGION + "/UserComponentA",
+                    request.logGroupName());
+            assertNotNull(request.logEvents());
+            assertEquals(60, request.logEvents().size());
+            assertEquals(61440, request.logEvents().stream().mapToLong(value -> value.message().length())
```
Nelsonochoam, line 382: Avoid the magic numbers use a variable name that represent what this is

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -350,4 +350,52 @@ void GIVEN_log_manager_in_errored_state_WHEN_restarted_THEN_logs_upload_is_reatt
...
+        tempDirectoryPath = Files.createTempDirectory(tempRootDir, "IntegrationTestsTemporaryLogFiles");
+
+        for (int i = 0; i < 5; i++) {
+            createTempFileAndWriteData(tempDirectoryPath, "integTestRandomLogFiles.log_",  "");
+        }
+        createTempFileAndWriteData(tempDirectoryPath, "integTestRandomLogFiles.log",  "");
+
+        setupKernel(tempDirectoryPath, "smallPeriodicIntervalUserComponentConfig.yaml");
+
+        TimeUnit.SECONDS.sleep(30);
```
Nelsonochoam, line 371: Instead of sleeping it would be nice to have something that checks for a condition to become eventually True. I don't know if there is already on this repo but we have a bunch of test utils that already do that.

dgggIoT: Honestly, I can`t think of one condition can be checked because it is a periodic cycle.

Nelsonochoam: You can replace a dependency you inject and attach a spy to it. I'll help with that

src/test/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessorTest.java
```diff
@@ -557,7 +557,7 @@ void GIVEN_null_log_message_WHEN_upload_attempted_THEN_null_message_considered_a
         assertTrue(file.setReadable(true));
         assertTrue(file.setWritable(true));
         try (OutputStream fileOutputStream = Files.newOutputStream(file.toPath())) {
-            fileOutputStream.write("null".getBytes(StandardCharsets.UTF_8));
+            fileOutputStream.write("null\n".getBytes(StandardCharsets.UTF_8));
```
dgggIoT, line 560: The purpose of this unit test is handling the null message, however the new mechanism will only consider the null message with a line separator.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -350,4 +353,105 @@ void GIVEN_log_manager_in_errored_state_WHEN_restarted_THEN_logs_upload_is_reatt
...
+        List<PutLogEventsRequest> putLogEventsRequests = captor.getAllValues();
+        assertEquals(1, putLogEventsRequests.size());
+
+        for (PutLogEventsRequest request : putLogEventsRequests) {
+            assertEquals(calculateLogStreamName(THING_NAME, LOCAL_DEPLOYMENT_GROUP_NAME), request.logStreamName());
+            assertEquals("/aws/greengrass/GreengrassSystemComponent/" + AWS_REGION + "/System",
+                    request.logGroupName());
+            assertNotNull(request.logEvents());
+            assertTrue(request.logEvents().size() >= DEFAULT_LOG_LINE_IN_FILE * 6);
+            assertTrue(request.logEvents().stream().mapToLong(value -> value.message().length()).sum()
```
Nelsonochoam, line 438: Extract this `request.logEvents().stream().mapToLong(value -> value.message().length()).sum()` into a variable called `totalBytesUploaded` also why is the assertion not a `>=`?

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -350,4 +353,105 @@ void GIVEN_log_manager_in_errored_state_WHEN_restarted_THEN_logs_upload_is_reatt
...
+                    >= DEFAULT_FILE_SIZE * 6);
+        }
+        File folder = tempDirectoryPath.toFile();
+        Pattern logFileNamePattern = Pattern.compile(String.format(DEFAULT_FILE_REGEX, fileName));
+        List<File> allFiles = new ArrayList<>();
+        File[] files = folder.listFiles();
```
Nelsonochoam, line 444: There is a lot of overlap between this test and the one above it is really hard to understand what this is testing.  Try using better naming

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -630,9 +643,11 @@ private void processLogsAndUpload() throws InterruptedException {
                 } catch (SecurityException e) {
                     logger.atError().cause(e).log("Unable to get log files for {} from {}",
                             componentName, componentLogConfiguration.getDirectoryPath());
+                } catch (InvalidLogGroupException e) {
+                    logger.atError().cause(e).log("Unable to read the directory");
                 }
             }
-            if (componentLogFileInformation.get().isPresent()) {
+            if (!isActiveFileCompleted.get() && componentLogFileInformation.get().isPresent()) {
```
Nelsonochoam, line 650: Do we need this new check?  Would the active file be in the componentLogFileInformation already?

src/test/java/com/aws/greengrass/logmanager/model/LogFileGroupTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Test
+    void GIVEN_log_files_THEN_find_the_active_file() throws IOException, InterruptedException,
+            InvalidLogGroupException {
+        ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG.set(true);
+        LogFile file = new LogFile(directoryPath.resolve("greengrass_test.log_1").toUri());
+        byte[] bytesArray = givenAStringOfSize(1024).getBytes(StandardCharsets.UTF_8);
+        writeFile(file, bytesArray);
+
+        TimeUnit.SECONDS.sleep(1);
```
Nelsonochoam, line 37: Why this sleep?

src/test/java/com/aws/greengrass/logmanager/util/UnitTestLogFileHelper.java
```diff
@@ -0,0 +1,30 @@
+package com.aws.greengrass.logmanager.util;
+
+import com.aws.greengrass.logmanager.model.LogFile;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.util.Random;
+
+public final class UnitTestLogFileHelper {
```
Nelsonochoam, line 10:
```suggestion
public final class TestUtils {
```

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -560,50 +577,37 @@ private void processLogsAndUpload() throws InterruptedException {
             }
             AtomicReference<Optional<ComponentLogFileInformation>> componentLogFileInformation =
                     new AtomicReference<>(Optional.empty());
+            AtomicBoolean isActiveFileCompleted = new AtomicBoolean(false);
```
Nelsonochoam, line 580: If this is all running on the same thread there is no need to use an `AtomicBoolean`

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -350,4 +353,105 @@ void GIVEN_log_manager_in_errored_state_WHEN_restarted_THEN_logs_upload_is_reatt
...
+
+    //TODO: this test is only for getting some certain level of knowledge of current change uploading active log file
+    // . It will be eventually removed.
+    @Test
+    void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_THEN_logs_are_uploaded_to_cloud(
+            ExtensionContext ec) throws Exception {
+        when(cloudWatchLogsClient.putLogEvents(any(PutLogEventsRequest.class)))
+                .thenReturn(PutLogEventsResponse.builder().nextSequenceToken("nextToken").build());
+        logManagerService.ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG.set(true);
+        ignoreExceptionOfType(ec, NoSuchFileException.class);
```
dgggIoT, line 414: This boolean value is only for giving a chance for this small PR to perform some tests without changing the current logic, such that all tests can pass and this small PR can be merged. This is only for serving the purpose of avoiding massive PR, and this static boolean constant will be removed in the end.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -350,4 +353,105 @@ void GIVEN_log_manager_in_errored_state_WHEN_restarted_THEN_logs_upload_is_reatt
...
+    }
+
+    //TODO: this test is only for getting some certain level of knowledge of current change uploading active log file
+    // . It will be eventually removed.
+    @Test
+    void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_THEN_logs_are_uploaded_to_cloud(
+            ExtensionContext ec) throws Exception {
+        when(cloudWatchLogsClient.putLogEvents(any(PutLogEventsRequest.class)))
+                .thenReturn(PutLogEventsResponse.builder().nextSequenceToken("nextToken").build());
+        logManagerService.ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG.set(true);
```
dgggIoT, line 413: This boolean value is only for giving a chance for this small PR to perform some tests without changing the current logic, such that all tests can pass and this small PR can be merged. This is only for serving the purpose of avoiding massive PR, and this static boolean constant will be removed in the end.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -350,4 +353,105 @@ void GIVEN_log_manager_in_errored_state_WHEN_restarted_THEN_logs_upload_is_reatt
...
+            for (File file : files) {
+                if (file.isFile()
+                        && logFileNamePattern.matcher(file.getName()).find()
+                        && file.length() > 0) {
+                    allFiles.add(file);
+                }
+            }
+        }
+        assertEquals(1, allFiles.size());
+        logManagerService.ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG.set(false);
```
dgggIoT, line 403: This boolean value is only for giving a chance for this small PR to perform some tests without changing the current logic, such that all tests can pass and this small PR can be merged. This is only for serving the purpose of avoiding massive PR, and this static boolean constant will be removed in the end.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -101,6 +103,9 @@ public class LogManagerService extends PluginService {
     public static final String MULTILINE_PATTERN_CONFIG_TOPIC_NAME = "multiLineStartPattern";
     public static final int DEFAULT_PERIODIC_UPDATE_INTERVAL_SEC = 300;
     private final Object spaceManagementLock = new Object();
+    // TODO: this is the flag to marking the code that used for the feature development, in order to maintain PR
+    //  small and we can modify tests in the following PR.
+    public static final AtomicBoolean ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG = new AtomicBoolean(false);
```
dgggIoT, line 108: This boolean value is only for giving a chance for this small PR to perform some tests without changing the current logic, such that all tests can pass and this small PR can be merged. This is only for serving the purpose of avoiding massive PR, and this static boolean constant will be removed in the end.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
dgggIoT, line 580: This does not necessarily to be a AtomicBoolean, this type only helps changing value in the temporary `ACTIVE_FILE_FEATURE_ENABLED_FLAG`.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -618,6 +622,15 @@ private void processLogsAndUpload() throws InterruptedException {
                                     startPosition = processingFileInformation.startPosition;
                                 }
                             }
+
+                            //TODO: setting this flag is only to develop incrementally without having to changed all
+                            // tests yet, so that we can avoid a massive PR. This will be removed in the end.
+                            if (ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG.get()) {
```
dgggIoT, line 628: This boolean value is only for giving a chance for this small PR to perform some tests without changing the current logic, such that all tests can pass and this small PR can be merged. This is only for serving the purpose of avoiding massive PR, and this static boolean constant will be removed in the end.

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.File;
+import java.net.URI;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.function.Consumer;
+import java.util.regex.Pattern;
+
+import static com.aws.greengrass.logmanager.LogManagerService.ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG;
```
dgggIoT, line 15: This boolean value is only for giving a chance for this small PR to perform some tests without changing the current logic, such that all tests can pass and this small PR can be merged. This is only for serving the purpose of avoiding massive PR, and this static boolean constant will be removed in the end.

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -0,0 +1,85 @@
...
+                        && filePattern.matcher(file.getName()).find()
+                        && file.length() > 0) {
+                    allFiles.add(file);
+                }
+            }
+        }
+        allFiles.sort(Comparator.comparingLong(LogFile::lastModified));
+        //TODO: setting this flag is only to develop incrementally without having to changed all tests yet, so that
+        // we can avoid a massive PR. This will be removed in the end.
+        if (!ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG.get()) {
```
dgggIoT, line 56: This boolean value is only for giving a chance for this small PR to perform some tests without changing the current logic, such that all tests can pass and this small PR can be merged. This is only for serving the purpose of avoiding massive PR, and this static boolean constant will be removed in the end.

src/test/java/com/aws/greengrass/logmanager/model/LogFileGroupTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.time.Instant;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
+
+import static com.aws.greengrass.logmanager.LogManagerService.ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG;
```
dgggIoT, line 17: This boolean value is only for giving a chance for this small PR to perform some tests without changing the current logic, such that all tests can pass and this small PR can be merged. This is only for serving the purpose of avoiding massive PR, and this static boolean constant will be removed in the end.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
MikeDombo: What are we trying to wait for?

Why is this sleep here?

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
MikeDombo, line 414: why are we throwing this exception to begin with? We should consider not throwing it, or other remedies. There's nothing a user can do about this error anyway.

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -122,15 +124,22 @@ public CloudWatchAttempt processLogFiles(ComponentLogFileInformation componentLo
...
-
+        LogFileGroup logFileGroup = componentLogFileInformation.getLogFileGroup();
         // Run the loop until all the log files from the component have been read or the max message
         // size has been reached.
         while (!componentLogFileInformation.getLogFileInformationList().isEmpty() && !reachedMaxSize.get()) {
-            LogFile logFile = componentLogFileInformation.getLogFileInformationList().get(0).getLogFile();
-            long startPosition = componentLogFileInformation.getLogFileInformationList().get(0).getStartPosition();
-            //TODO Optional<String> fileHash = componentLogFileInformation.getLogFileInformationList().get(0)
-            // .getFileHash();
+            LogFileInformation tempLogFileInfomation = componentLogFileInformation.getLogFileInformationList().get(0);
```
MikeDombo, line 131: `temp` is not necessary. Just call it the `logFileInformation`
Also typo: `Infomation` --> `Information`

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -102,6 +102,9 @@ public String hashString() {
         }
 
         return fileHash;
+    }
 
+    public boolean isEmpty() {
+        return "".equals(this.hashString());
```
MikeDombo, line 108: `Utils.isEmpty(hashString())`

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -0,0 +1,85 @@
...
+     * @param lastUpdated the saved updated time of the last uploaded log of a component.
+     * @return list of logFile.
+     * @throws InvalidLogGroupException the exception if this is not a valid directory.
+     */
+    public static LogFileGroup create(Pattern filePattern, URI path, Instant lastUpdated)
+            throws InvalidLogGroupException {
+        File folder = new File(path);
+
+        if (!folder.isDirectory()) {
+            throw new InvalidLogGroupException("Must be a folder.");
```
MikeDombo, line 38: `String.format("%s must be a directory", path)`

src/test/java/com/aws/greengrass/logmanager/model/LogFileGroupTest.java
MikeDombo: This shouldn't be needed. Nothing here is running in different threads.

src/test/java/com/aws/greengrass/logmanager/util/UnitTestLogFileHelper.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private UnitTestLogFileHelper() { }
+
+    public static String givenAStringOfSize(int bytesNeeded) {
+        StringBuilder testStrings = new StringBuilder();
+        Random rnd = new Random();
```
MikeDombo, line 16: save Random as a static field in this class, it is expensive to construct random.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
dgggIoT: The sleep is from the existing code, my understanding for it is to lazily wait for Kernel finish entire cycle then we can evaluate the result. This will be refactored, I will add a TODO.

src/test/java/com/aws/greengrass/logmanager/model/LogFileGroupTest.java
dgggIoT: This is only for lazily differ the creation time of two files, then we know the last file is the active file.

MikeDombo: Add a comment that says why we're doing this then.

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -35,7 +35,7 @@ public static LogFileGroup create(Pattern filePattern, URI path, Instant lastUpd
         File folder = new File(path);
 
         if (!folder.isDirectory()) {
-            throw new InvalidLogGroupException("Must be a folder.");
+            throw new InvalidLogGroupException(String.format("%s must be a directory", path));
```
MikeDombo, line 38: what if it doesn't exist, but it _will_ exist in the future and will be a directory? What would we do then?

dgggIoT: The creation of logFileGroup is like a snapshot-mechanism to be created in every cycle. 
If the directory doesn't exist now, we could just skip it for this cycle. 
But if it exist in the next round, the logFileGroup will be newly created, so it will become a valid directory.

MikeDombo: OK, that's good. What do we do with the exception? We shouldn't be continuously logging the same issue. Definitely not at error level.

dgggIoT: Agree that it should not be at error Level, I will put it in the debug level.

Nelsonochoam (approved):
Approving this - as we have discussed, in the next PR we will need to

* Remove the old code
* Update the old tests
* Remove the feature flag

The one after that one we will need to refactor and clean this up a bit and improve the tests

MikeDombo (approved):

PR #105, opened 2022-10-14
feat: enable logFileGroup update with latest directory

dgggIoT:
**Issue #, if available:**
#51 
**Description of changes:**
Add the logFileGroup update to the latest directory, thus grab information even log rotation happens during uploading.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -513,15 +513,20 @@ public class LogManagerService extends PluginService {
                                                                 cloudWatchAttemptLogFileInformation) {
         //Todo: eventually the file and fileName should be obtained by the fileHash
         LogFile file = new LogFile(fileName);
-        String componentName = attemptLogInformation.getComponentName();
         boolean isActiveFile = false;
         //TODO: setting this flag is only to develop incrementally without having to changed all tests yet, so that
         // we can avoid a massive PR. This will be removed in the end.
         if (ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG.get()) {
             try {
                 String fileHash = cloudWatchAttemptLogFileInformation.getFileHash();
-                LogFileGroup logFileGroup = attemptLogInformation.getLogFileGroup().update();
-                file = logFileGroup.getFile(fileHash);
+                LogFileGroup logFileGroup = attemptLogInformation.getLogFileGroup().syncDirectory();
+                Optional<LogFile> file1 = logFileGroup.getFile(fileHash);
+                // TODO: this is temporary for not changing the original logic, will be well handled.
+                if (!file1.isPresent()) {
+                    logger.atDebug().log("The fileHash does not exist in directory");
+                    return;
+                }
+                file = file1.get();
                 fileName = file.getAbsolutePath();
                 // TODO: the following logic is only for passing this small PR while not changing the current context.
                 //  We will grab the fileHash in the future.
@@ -532,6 +537,7 @@ public class LogManagerService extends PluginService {
         }
         // If we have completely read the file, then we need add it to the completed files list and remove it
         // it (if necessary) for the current processing list.
+        String componentName = attemptLogInformation.getComponentName();
         if (!isActiveFile && file.length() == cloudWatchAttemptLogFileInformation.getBytesRead()
                 + cloudWatchAttemptLogFileInformation.getStartPosition()) {
             Set<String> completedFileNames = completedLogFilePerComponent.getOrDefault(componentName, new HashSet<>());
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -530,9 +530,13 @@ public class LogManagerService extends PluginService {
                 fileName = file.getAbsolutePath();
                 // TODO: the following logic is only for passing this small PR while not changing the current context.
                 //  We will grab the fileHash in the future.
-                isActiveFile = logFileGroup.isActiveFile(fileHash);
+                Optional<LogFile> activeFile = logFileGroup.getActiveFile();
+                if (activeFile.isPresent()) {
+                    isActiveFile = activeFile.get().fileEquals(file);
+                }
             } catch (InvalidLogGroupException e) {
                 logger.atDebug().cause(e).log();
+                return;
             }
         }
         // If we have completely read the file, then we need add it to the completed files list and remove it
@@ -638,7 +642,8 @@ public class LogManagerService extends PluginService {
                             //TODO: setting this flag is only to develop incrementally without having to changed all
                             // tests yet, so that we can avoid a massive PR. This will be removed in the end.
                             if (ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG.get()) {
-                                if (logFileGroup.isActiveFile(fileHash) && startPosition == file.length()) {
+                                if (file.fileEquals(logFileGroup.getActiveFile().get())
+                                        && startPosition == file.length()) {
                                     isActiveFileCompleted.set(true);
                                 }
                             }
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -518,11 +518,17 @@ public class LogManagerService extends PluginService {
         //TODO: setting this flag is only to develop incrementally without having to changed all tests yet, so that
         // we can avoid a massive PR. This will be removed in the end.
         if (ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG.get()) {
-            String fileHash = cloudWatchAttemptLogFileInformation.getFileHash();
-            LogFileGroup logFileGroup = attemptLogInformation.getLogFileGroup();
-            // TODO: the following logic is only for passing this small PR while not changing the current context.
-            //  We will grab the fileHash in the future.
-            isActiveFile = logFileGroup.isActiveFile(fileHash);
+            try {
+                String fileHash = cloudWatchAttemptLogFileInformation.getFileHash();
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -108,4 +108,8 @@ public String hashString() {
     public boolean isEmpty() {
         return Utils.isEmpty(this.hashString());
     }
+
+    public boolean isEmpty(String hash) {
```
Nelsonochoam, line 112: I don't think passing a hash to isEmpty makes much sense.

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -32,7 +42,11 @@ private LogFileGroup(List<LogFile> files) {
      */
     public static LogFileGroup create(Pattern filePattern, URI path, Instant lastUpdated)
             throws InvalidLogGroupException {
+        savedFilePattern = filePattern;
```
Nelsonochoam, line 45: Call this filePattern remove the saved prefix all variables are saved in memory

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -82,4 +98,21 @@ public void forEach(Consumer<LogFile> callback) {
...
+     * @return the logFile.
+     */
+    public LogFile getFile(String fileHash) {
+        if (!fileHashToFileMap.containsKey(fileHash)) {
+            logger.atDebug().log("FileHash does not exist");
+        }
+        return fileHashToFileMap.get(fileHash);
+    }
+
+    public LogFileGroup update() throws InvalidLogGroupException {
```
Nelsonochoam, line 114: Update is not representative of what it is doing. Call is something like snapshot. Also add some doc to the method that explains what it does and maybe why it is useful

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -82,4 +98,21 @@ public void forEach(Consumer<LogFile> callback) {
     public boolean isEmpty() {
         return this.logFiles.isEmpty();
     }
+
+    /**
+     * Get the LogFile object from the fileHash.
+     * @param fileHash the fileHash obtained from uploader.
+     * @return the logFile.
+     */
+    public LogFile getFile(String fileHash) {
```
Nelsonochoam, line 107: Return an Optional<Logfile> that represents something might not be there. If you do that you dont need the log

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -42,11 +56,13 @@ public static LogFileGroup create(Pattern filePattern, URI path, Instant lastUpd
...
         if (files.length != 0) {
             for (LogFile file: files) {
+                String fileHash = file.hashString();
                 if (file.isFile()
                         && lastUpdated.isBefore(Instant.ofEpochMilli(file.lastModified()))
                         && filePattern.matcher(file.getName()).find()
-                        && file.length() > 0) {
+                        && !file.isEmpty(fileHash)) {
                     allFiles.add(file);
+                    fileHashToFileMap.put(fileHash, file);
```
Nelsonochoam, line 65: Why have 2 data structures holding the same stuff?. You could do away with just the map and keep track of the active file on a separate variable basically while we loop through the files get check the lastModified and store the one with the greatest value.

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -17,6 +22,11 @@
 public final class LogFileGroup {
     @Getter
     private List<LogFile> logFiles;
+    private static Map<String, LogFile> fileHashToFileMap;
```
Nelsonochoam, line 25:
```suggestion
    private static Map<String, LogFile> fileHashToLogFile
```

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -518,11 +518,17 @@ private void processCloudWatchAttemptLogInformation(Map<String, Set<String>> com
         //TODO: setting this flag is only to develop incrementally without having to changed all tests yet, so that
         // we can avoid a massive PR. This will be removed in the end.
         if (ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG.get()) {
-            String fileHash = cloudWatchAttemptLogFileInformation.getFileHash();
-            LogFileGroup logFileGroup = attemptLogInformation.getLogFileGroup();
-            // TODO: the following logic is only for passing this small PR while not changing the current context.
-            //  We will grab the fileHash in the future.
-            isActiveFile = logFileGroup.isActiveFile(fileHash);
+            try {
+                String fileHash = cloudWatchAttemptLogFileInformation.getFileHash();
```
Nelsonochoam, line 522: Where is the test for this?

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
dgggIoT: This map is not for active file. It is for getting the file to be deleted.

Nelsonochoam: Yeah, I know I am saying we don't need 2 data structures we could do away with that map and a variable that stores the active file

dgggIoT: Actually, besides the reason of getting the active file from sorted files, in case of some edge cases, the LM wants to process the oldest file first. E.g. when processing new file, the older files get deleted due to the disk space limit, so LM need to process older files first to avoid data loss.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -513,19 +513,31 @@ private void processCloudWatchAttemptLogInformation(Map<String, Set<String>> com
...
+                    logger.atDebug().log("The fileHash does not exist in directory");
+                    return;
+                }
+                file = file1.get();
+                fileName = file.getAbsolutePath();
+                // TODO: the following logic is only for passing this small PR while not changing the current context.
+                //  We will grab the fileHash in the future.
+                isActiveFile = logFileGroup.isActiveFile(fileHash);
+            } catch (InvalidLogGroupException e) {
+                logger.atDebug().cause(e).log();
```
Nelsonochoam, line 535: Wouldn't we want to skip running the part below if this fails

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -513,19 +513,31 @@ private void processCloudWatchAttemptLogInformation(Map<String, Set<String>> com
...
+                // TODO: this is temporary for not changing the original logic, will be well handled.
+                if (!file1.isPresent()) {
+                    logger.atDebug().log("The fileHash does not exist in directory");
+                    return;
+                }
+                file = file1.get();
+                fileName = file.getAbsolutePath();
+                // TODO: the following logic is only for passing this small PR while not changing the current context.
+                //  We will grab the fileHash in the future.
+                isActiveFile = logFileGroup.isActiveFile(fileHash);
```
Nelsonochoam, line 533: Using `isActiveFile` and passing a hash doesn't seem like a good api. Why don't we

```java
LogFile activeFile = logGroup.getActiveFile();
isActiveFile = activeFile.equals(file)
```

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -82,4 +96,29 @@ public void forEach(Consumer<LogFile> callback) {
...
     }
+
+    /**
+     * Get the LogFile object from the fileHash.
+     * @param fileHash the fileHash obtained from uploader.
+     * @return the logFile.
+     */
+    public Optional<LogFile> getFile(String fileHash) {
+        if (!fileHashToLogFile.containsKey(fileHash)) {
+            return Optional.ofNullable(null);
```
Nelsonochoam, line 107:
```suggestion
            return Optional.empty();
```

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -82,4 +96,29 @@ public void forEach(Consumer<LogFile> callback) {
...
+        }
+        return Optional.ofNullable(fileHashToLogFile.get(fileHash));
```
Nelsonochoam, line 109:
```suggestion
        return Optional.of(fileHashToLogFile.get(fileHash));
```

src/test/java/com/aws/greengrass/logmanager/LogManagerServiceTest.java
```diff
@@ -587,28 +593,49 @@ void GIVEN_cloud_watch_attempt_handler_WHEN_attempt_completes_THEN_successfully_
 
         CloudWatchAttempt attempt = new CloudWatchAttempt();
         Map<String, CloudWatchAttemptLogInformation> logStreamsToLogInformationMap = new HashMap<>();
-        File file1 = new File(getClass().getResource("testlogs2.log").toURI());
-        File file2 = new File(getClass().getResource("testlogs1.log").toURI());
-        Map<String, CloudWatchAttemptLogFileInformation> attemptLogFileInformationMap1 = new HashMap<>();
-        attemptLogFileInformationMap1.put(file1.getAbsolutePath(), CloudWatchAttemptLogFileInformation.builder()
+
+        LogFile lastProcessedFile = new LogFile(directoryPath.resolve("testlogs2.log").toUri());
```
Nelsonochoam, line 597: This test can be refactored. See if you can remove the duplication and extract what is similar.

Nelsonochoam (approved):
Minor comment but it looks good

MikeDombo (approved):

PR #106, opened 2022-10-20
feat: remove fileName and add randomly-named rotation tests

dgggIoT:
**Issue #, if available:**
#51 
**Description of changes:**
Totally remove the fileName, and only use fileHash
Now we can handle active log file uploading
Existing tests have been modified, and new tests for log rotations are added.

**Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
@@ -16,6 +16,7 @@ import com.aws.greengrass.logging.impl.config.LogConfig;
 import com.aws.greengrass.logging.impl.config.LogStore;
 import com.aws.greengrass.logging.impl.config.model.LogConfigUpdate;
 import com.aws.greengrass.logmanager.LogManagerService;
+import com.aws.greengrass.logmanager.model.LogFile;
 import com.aws.greengrass.testcommons.testutilities.GGExtension;
 import com.aws.greengrass.util.exceptions.TLSAuthException;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -46,11 +47,13 @@ import java.nio.file.Paths;
 import java.text.SimpleDateFormat;
 import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.TimeZone;
+import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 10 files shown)


src/test/java/com/aws/greengrass/logmanager/LogManagerServiceTest.java
```diff
@@ -829,25 +850,35 @@ void GIVEN_user_component_logs_delete_file_after_upload_set_WHEN_upload_logs_THE
...
         CloudWatchAttempt attempt = new CloudWatchAttempt();
         Map<String, CloudWatchAttemptLogInformation> logStreamsToLogInformationMap = new HashMap<>();
-        File file1 = new File(directoryPath.resolve(fileNames.get(0)).toUri());
-        File file2 = new File(directoryPath.resolve(fileNames.get(1)).toUri());
+        LogFile file1 = new LogFile(directoryPath.resolve(fileNames.get(0)).toUri());
+        LogFile file2 = new LogFile(directoryPath.resolve(fileNames.get(1)).toUri());
+        //TODO
+        //System.out.println(file1.getName());
+        //System.out.println(file2.getName());
+        //System.out.println("test create logFileGroup");
```
Nelsonochoam, line 861: 🔪 remove this

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -225,14 +224,14 @@ private boolean processLogLine(AtomicInteger totalBytesRead,
...
                                    StringBuilder data,
-                                   String fileName,
                                    String fileHash,
                                    long startPosition,
                                    String componentName,
                                    long currentPosition,
                                    long lastModified,
                                    LogFileGroup logFileGroup) {
         String dataStr = data.toString();
+        //System.out.println(dataStr);
```
MikeDombo, line 234: remove

src/main/java/com/aws/greengrass/logmanager/model/ComponentLogConfiguration.java
```diff
@@ -25,6 +25,5 @@ public class ComponentLogConfiguration {
     private Level minimumLogLevel = Level.INFO;
     private Long diskSpaceLimit;
     private boolean deleteLogFileAfterCloudUpload;
-    private boolean uploadToCloudWatch;
```
MikeDombo, line 28: why are we removing this option?

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -55,24 +54,21 @@ public static LogFileGroup create(Pattern filePattern, URI directoryURI, Instant
         if (files.length != 0) {
             for (LogFile file: files) {
                 String fileHash = file.hashString();
+                boolean isModifiedAfterLastUpdatedFile =
```
MikeDombo, line 57: why do we care about the timestamp? What is the intention behind checking this?

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -504,68 +501,57 @@ private void handleCloudWatchAttemptStatus(CloudWatchAttempt cloudWatchAttempt)
...
-                Optional<LogFile> activeFile = logFileGroup.getActiveFile();
-                if (activeFile.isPresent()) {
-                    isActiveFile = activeFile.get().fileEquals(file);
-                }
-            } catch (InvalidLogGroupException e) {
-                logger.atDebug().cause(e).log();
+        try {
+            LogFileGroup logFileGroup = attemptLogInformation.getLogFileGroup().syncDirectory();
+            if (!logFileGroup.getFile(fileHash).isPresent()) {
+                logger.atDebug().log("The fileHash does not exist in directory");
```
MikeDombo, line 514: I don't think we really care about this. Maybe drop it to trace level.

src/main/java/com/aws/greengrass/logmanager/model/ComponentLogConfiguration.java
dgggIoT: This is wrongly added before, which should be in the next PR to allow userComponent has the boolean value.

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
dgggIoT: This is to filter out files that has been processed before, if they are modified before the lastUpdated file

MikeDombo:
In general, this change is too large, I think it can be broken up into at least 2 PRs. 1 for adding new tests, and 1 for removing the fileName and using fileHash instead. If it can be broken up even more, then that would be helpful for reviewers.

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
MikeDombo: We shouldn't be relying on timestamps for that. We should use the file size. If the file is longer than what we've already read, then we need to read it more and upload the rest.

dgggIoT: You are referring the case that rotated files may also still be adding some contents, right? But currently we don`t have a logic to record the bytesRead for each file, we only have it for the current processing file. This will be a big change, we can leave it as future refactoring.

dgggIoT (commented):
Deprecate this PR, because this PR will be divided into two small PRs to help reviewing.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -177,22 +181,19 @@ void GIVEN_user_component_config_with_small_periodic_interval_WHEN_interval_elap
...
         for (PutLogEventsRequest request : putLogEventsRequests) {
             assertEquals(calculateLogStreamName(THING_NAME, LOCAL_DEPLOYMENT_GROUP_NAME), request.logStreamName());
-            assertEquals("/aws/greengrass/UserComponent/" + AWS_REGION + "/UserComponentA",
-                    request.logGroupName());
+            assertEquals("/aws/greengrass/UserComponent/" + AWS_REGION + "/UserComponentA", request.logGroupName());
             assertNotNull(request.logEvents());
-            assertEquals(50, request.logEvents().size());
-            assertEquals(51200, request.logEvents().stream().mapToLong(value -> value.message().length())
-                    .sum());
+            assertEquals(DEFAULT_LOG_LINE_IN_FILE * 6, request.logEvents().size());
```
Nelsonochoam, line 186: It is not completely obvious what 6 means here. Is there a way it could be more descriptive? - like 

```java
// At the top of the test
LogFileGroup group = LogFileGroup.create(
   Pattern.compile("^integTestRandomLogFiles.log\\w*"), tempDirectoryPath.toUri(), Instant.now())

// Create the log file etc etc

// Inside the for loop
assertEquals(DEFAULT_LOG_LINE_IN_FILE * group.size(), request.logEvents().size());
```

Also this will remove the need for that weird for loop we have everywhere

PR #107, opened 2022-10-21
feat: add new tests for randomly-named rotation

dgggIoT:
**Issue #, if available:**
#51 
**Description of changes:**
Add new tests to handle the randomly-named rotation scenario, which should be well-handled when the LogManager does not rely on fileNames. 

Not tested yet, will be tested in another PR.

**Checklist:**


src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/util/LogFileHelper.java
```diff
@@ -68,4 +70,15 @@ public static void createFileAndWriteData(Path tempDirectoryPath, String fileNam
             addDataToFile(messageBytes, file.toPath());
         }
     }
+
+    public static LogFile createTempFileAndWriteDataAndReturnFile(Path tempDirectoryPath, String fileNamePrefix,
+                                                                  String fileNameSuffix) throws IOException {
+        Path filePath = Files.createTempFile(tempDirectoryPath, fileNamePrefix, fileNameSuffix);
```
MikeDombo, line 76: passing a `suffix` isn't required, you can just give it a random prefix.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -453,4 +461,174 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_
...
+            String randomName2 = UUID.randomUUID().toString();
+            LogFile logFile = createTempFileAndWriteDataAndReturnFile(tempDirectoryPath,
+                    randomName1 + "UserComponentB.", randomName2);
+            logFiles.add(logFile);
+        }
+
+        setupKernel(tempDirectoryPath, "randomlyNamedFilesWithoutComponentsConfig.yaml");
+        //TODO: this lazy sleeping inherits from existing tests, but should be replaced by some better mechanism in
+        // future refactoring
+        TimeUnit.SECONDS.sleep(15);
```
MikeDombo, line 494: what are we sleeping for? In this todo, you need to assume that you aren't the one who will fix it. What are we trying to wait for?

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -453,4 +461,174 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_
...
+            assertEquals(renameFile.hashString(), tempFileHash);
+            assertFalse(logFile.exists());
+        }
+        //TODO: this lazy sleeping inherits from existing tests, but should be replaced by some better mechanism in
+        // future refactoring
+        TimeUnit.SECONDS.sleep(15);
+
+        // Even the file are all renamed, the cloudWatchClient should only try to upload events once, and the total
+        // size should be contents of 4 files.
+        verify(cloudWatchLogsClient, times(1)).putLogEvents(captor.capture());
```
MikeDombo, line 513: Also verify that our map of filehashes has a size of 4. I want to make sure that we aren't growing that map forever.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
MikeDombo: What does "finish the entire cycle" mean. What, specifically, are we waiting to happen? If we could put a condition anywhere in the code to tell this test that it is ready to continue, where would that condition be? What is the trigger for us to continue?

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -453,4 +461,174 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_
...
+                if (file.isFile() && logFileNamePattern.matcher(file.getName()).find() && file.length() > 0) {
+                    allFiles.add(file);
+                }
+            }
+        }
+        assertEquals(testFileNumber, allFiles.size());
+    }
+
+    @Test
+    void GIVEN_files_randomly_named_WHEN_all_files_renamed_randomly_and_new_data_in_active_file_THEN_all_files_only_send_once()
```
MikeDombo, line 541: `new_data_in_active_file`, what about new data in some other file? In theory, we should be handling that too

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -453,4 +461,174 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_
         assertEquals(1, allFiles.size());
         logManagerService.ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG.set(false);
     }
+
+    @Test
+    void GIVEN_files_randomly_named_WHEN_all_files_renamed_randomly_THEN_all_files_only_send_once()
+            throws Exception {
+        //TODO: Intentionally pass the test now, since it will fail for current code.
+        if (!ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG.get()) {
```
Nelsonochoam, line 469: This is weird. Now you are letting this test be controlled from the outside. You should keep doing what you are doing with the other ones where the test sets the flag to a value

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
dgggIoT: We do not keep the global map of fileHash anymore. The new structure is introduced to scan and filter all files in directory, which is a kinda snapshot mechanism of the current directory.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
dgggIoT: Are you referring to the scenario that new data can be written in rotated file? 
Our current logic will skip rotated and uploaded files when scanning.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
MikeDombo: There is some map somewhere though, right? otherwise you'll lose the information about what we've uploaded and then we'll be duplicating the logs in the upload.

So, where are we tracking the hash now?

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
dgggIoT: This is for skipping this tests, because the PR just try to show the newly tests, but this test is not ran until next PR.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
MikeDombo: Yes. I don't love that, but OK.

dgggIoT: Agree. This will be included in future refactoring.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
dgggIoT: Avoiding duplicate logs is done by the lastModifiedTime and fileHash of the file that lastly uploaded, which are stored in runtime config and can be seen in next PR. 
For every scanning, we will create a logFileGroup structure and filter out uploaded and rotated file, so the internal map of fileHash is newly created. 
Here is the code: https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/main/src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java#L56

The logic is: 
1. Get the timestamp of the moment of last upload, and the fileHash of the partially reading file. 
2. Create the logFIleGroup, which will scan all files in the directory, filter files that is before the timestamp of last upload. Then store the Map <fileHash, file> as internal map. 
3. get the startPosition based on matched fileHash. 
4. After uploading, re-create the logFileGroup to get the fresh snap-shot of the directory, then classify file status based on fileHash.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
Nelsonochoam: Still a bad idea. you are basically having something external that is not being set by this test execute it. If this is related to the active log file why cant you just set the `logManagerService.ACTIVE_LOG_FILE_FEATURE_ENABLED_FLAG.set(true);` and then back to false

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -453,4 +464,242 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_
...
+
+        tempDirectoryPath = Files.createDirectory(tempRootDir.resolve("logs"));
+        LogConfig.getRootLogConfig().setLevel(Level.TRACE);
+        LogConfig.getRootLogConfig().setStore(LogStore.FILE);
+        LogManager.getLogConfigurations().putIfAbsent("UserComponentB",
+                newLogConfigFromRootConfig(LogConfigUpdate.builder().fileName("UserComponentB.log").build()));
+
+        List<LogFile> logFiles = new ArrayList<>();
+        int testFileNumber = 4;
+        for (int i = 0; i < testFileNumber; i++) {
```
Nelsonochoam, line 486: Why not update the helper to have `createTempFileAndWriteData(Path tempDirectoryPath, Supplier<String> fileNamePrefix, Supplier<String> fileNameSuffix)`. That way you avoid duplicate functionality

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -453,4 +464,242 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_
...
+                     @Override
+                     public Object answer(InvocationOnMock invocation)
+                     {
+                         latch1.countDown();
+                         return null;
+                     }
+                 }
+        ).when(logManagerService).isActiveFileCompleted.equals(true);
+
+        try {
```
Nelsonochoam, line 507: Remove this try catch just add it to the test method signature

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -122,6 +122,8 @@ public class LogManagerService extends PluginService {
     @Getter
     private int periodicUpdateIntervalSec;
     private Future<?> spaceManagementThread;
+    // public only for integ tests
+    public final AtomicBoolean isActiveFileCompleted = new AtomicBoolean(false);
```
Nelsonochoam, line 126: Exposing variables like this only for tests is not recommended. If you wanted to you could allow this service to have listeners that can get called when something happens. A very simple implementation for this:

```java
    private final List<Consumer<Event>> listeners = new ArrayList<>();

    enum Event {
          ACTIVE_FILE_COMPLETED
     }

     public registerlistener(Consumer<Event> listener) {
          listeners.add(listener);
     }

    private dispatchEvent(Event event) {
          for (Consumer<Event> listener : listeners) {
                listener(event);
           }
     }
```

even though for now it is only used by tests, listeners are opaque and do not know about the fact we store thing internally in a variable called `isActiveFileCompleted`. Additionally it allows you to do whatever you want and react to events.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -453,4 +464,242 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_
...
+        doReturn(new Answer<Object>()
+                 {
+                     @Override
+                     public Object answer(InvocationOnMock invocation)
+                     {
+                         latch1.countDown();
+                         return null;
+                     }
+                 }
+        ).when(logManagerService).isActiveFileCompleted.equals(true);
```
Nelsonochoam, line 505: I added a comment about how would can avoid having to do this weird mock. You might still need the countdown latch though.

Nelsonochoam (dismissed):

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -453,4 +464,242 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_
...
+        } catch (InterruptedException e) {
+            // do nothing
+        }
+
+        //TODO: this lazy sleeping inherits from existing tests, but should be replaced by some better mechanism in
+        // future refactoring
+        //TimeUnit.SECONDS.sleep(15);
+
+        // randomly rename file and see if uploader trying to process more file.
+        for (LogFile logFile : logFiles) {
```
Nelsonochoam, line 518: This could be a method. 

```java
private renameRotatedFiles(List<logFile> logFile, Function<LogFile, String> rename)
....
```

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -453,4 +464,242 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_
...
+        for (PutLogEventsRequest request : putLogEventsRequests) {
+            assertEquals(calculateLogStreamName(THING_NAME, LOCAL_DEPLOYMENT_GROUP_NAME), request.logStreamName());
+            assertEquals("/aws/greengrass/UserComponent/" + AWS_REGION + "/UserComponentB", request.logGroupName());
+            assertNotNull(request.logEvents());
+            assertEquals(DEFAULT_LOG_LINE_IN_FILE * testFileNumber, request.logEvents().size());
+            assertEquals(DEFAULT_FILE_SIZE * testFileNumber,
+                    request.logEvents().stream().mapToLong(value -> value.message().length()).sum());
+        }
+        // confirm no file is deleted
+        File folder = tempDirectoryPath.toFile();
```
Nelsonochoam, line 564: Create a LogGroup here and check the size

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/util/LogFileHelper.java
```diff
@@ -68,4 +70,15 @@ public static void createFileAndWriteData(Path tempDirectoryPath, String fileNam
             addDataToFile(messageBytes, file.toPath());
         }
     }
+
+    public static LogFile createTempFileAndWriteDataAndReturnFile(Path tempDirectoryPath, String fileNamePrefix)
```
Nelsonochoam, line 74: Why do we need this? - This is doing the same as the one above. Just change the one above to return the file it should not break anything since existing callers will just ignore the returned result

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -453,4 +464,242 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_
...
+            logFiles.add(logFile);
+        }
+
+        setupKernel(tempDirectoryPath, "randomlyNamedFilesWithoutComponentsConfig.yaml");
+        //TODO: this lazy sleeping inherits from existing tests, but should be replaced by some better mechanism in
+        // future refactoring.
+        //TimeUnit.SECONDS.sleep(15);
+
+        final CountDownLatch latch1 = new CountDownLatch(1);
+        doReturn(new Answer<Object>()
```
Nelsonochoam, line 609: Same comment as above, there is tons of duplication here and the essence of the test gets lost with all the imperative steps. Extract things into named methods and compose functionality. See all the comments I left on the one above

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
dgggIoT: This is not just related to active log file, it is also related to the log rotation, which the current logic will not handle. So main idea is that skip this test.

dgggIoT (commented):
Note: All tests have passed locally when running `mvn clean test`, but fail the git PR build. 
Reproduce it locally with `mvn -ntp -U clean verify`, tests failed in local. 
However, I have rolled back several commits, even the one from last year, the failure still exists. I suspect that this failure is related to the recent Nucleus release.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
Nelsonochoam: I you set the active log file flag to true on another test this test would run. I don't see how that is skipping this test. You are operating under the assumption that this is false. Additionally what is the value of a test if you don't want to run it? - Regardless if you really want to skip the test for now there is an annotation called @Disabled that you can use

src/main/java/com/aws/greengrass/logmanager/util/ActiveFileCompletedListener.java
```diff
@@ -0,0 +1,22 @@
+package com.aws.greengrass.logmanager.util;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+
+public class ActiveFileCompletedListener {
```
Nelsonochoam, line 7: The listeners are actually just the consumers listening to events.

Also this should be directly part of the LogManagerService and it would just expose the registerListenerMethod as public

```suggestion
public class LogManagerService {
   private final List<Consumer<Event>> listeners = new ArrayList<>();

   static enum Event {
     LOG_GROUP_PROCESSED
   }

    public Runnable registerListener(Consumer<Event> c) {
         listeners.add(c);
         return () -> listeners.remove(c);
    }

   private void emit(Event e);
```

Also an event can ve anything, don't just make it be able to emit ActiveFile statuses. It doesn't work well of you then want to trigger other events.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -822,6 +829,10 @@ private void deleteFilesIfNecessary(ComponentLogConfiguration componentLogConfig
         }
     }
 
+    public void handleActiveFileStatus(ActiveFileCompletedListener.ActiveFileStatusType activeFileStatus) {
```
Nelsonochoam, line 832: Why do we need this here? I don't really think this is needed

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -122,6 +124,9 @@ public class LogManagerService extends PluginService {
     @Getter
     private int periodicUpdateIntervalSec;
     private Future<?> spaceManagementThread;
+    // public only for integ tests
```
Nelsonochoam, line 127: Remove this comment, also this should not be public. See the comment below, we should only expose the register listener

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -453,4 +466,162 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_
...
+                        request.logEvents().stream().mapToLong(value -> value.message().length()).sum());
+            }
+        }
+        // confirm no file is deleted
+        Pattern logFileNamePattern = Pattern.compile("\\w*UserComponentB\\w*.\\w*");
+        LogFileGroup logFileGroup = LogFileGroup.create(logFileNamePattern, tempDirectoryPath.toUri(), mockInstant);
+        assertEquals(testFileNumber, logFileGroup.getLogFiles().size());
+    }
+
+    boolean isActiveFileProcessedOrTimeout(LogManagerService logManagerService) throws InterruptedException {
```
Nelsonochoam, line 612: The idea of having that observer pattern that emit event was so that we avoid things like this, you could do something like


```suggestion
    private Runnable subscribeToActiveFileProcessed(LogManagerService service) throws InterruptedException {
       CountDownLatch latch = new CountDownLatch(1);
        Runnable deregister = service.registerListener((Event event) -> {
              if (event == Event.ACTIVE_FILE_PROCESSED) {
                     latch.countDown();
              }
        })
        
        return () -> {
             latch.await(15, TimeUnit.SECONDS);
             deregister();
        }
   }
   
   // Then on the test
   @Test
   void fooBar() {
       // this just registers a latch and a listener
       Runnable waitForActiveFileToBeProcessed = subscribeToActiveFileProcessed(service);
       
       // Do some stuff
       waitForActiveFileToBeProcessed(); // <-- calling this will block the thread until the latch is released
   }
```

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -453,4 +466,162 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_
...
+                    randomName1 + "UserComponentB.");
+            logFiles.add(logFile);
+        }
+
+        setupKernel(tempDirectoryPath, "randomlyNamedFilesWithoutComponentsConfig.yaml");
+        assertTrue(isActiveFileProcessedOrTimeout(logManagerService));
+
+        // randomly rename file and see if uploader trying to process more file.
+        logFiles.sort(Comparator.comparingLong(LogFile::lastModified));
+        for (int i = 0; i < logFiles.size(); i++) {
```
Nelsonochoam, line 562: This could be a method that renames the files and you could make it more declarative and refactor it. It is hard to figure out what the intent of this is by reading it.

Also the loop is doing 2 things renaming and writing to the active file. This could be split into

```java
rotateLogFiles(logFiles);
LogFile activeLogFile = logFiles.get(logFiles.size() - 1);
writeLogsToFile(activeLogFile)
```

src/main/java/com/aws/greengrass/logmanager/model/EventStatusType.java
```diff
@@ -0,0 +1,5 @@
+package com.aws.greengrass.logmanager.model;
+
+public enum EventStatusType {
```
Nelsonochoam, line 3:
```suggestion
public enum EventType {
```

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -675,6 +680,19 @@ private void processLogsAndUpload() throws InterruptedException {
         }
     }
 
+    /**
+     * This method is used for tests monitoring the event status in the service.
```
Nelsonochoam, line 684: Don't mention it is used for tests. Yes, right now it is used for tests but it could be used internally as well for things like sending metrics when something happens etc etc.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java
```diff
@@ -453,4 +463,163 @@ void GIVEN_system_config_with_small_periodic_interval_WHEN_active_logs_included_
...
+        verify(cloudWatchLogsClient, times(2)).putLogEvents(captor.capture());
+
+        List<PutLogEventsRequest> putLogEventsRequests = captor.getAllValues();
+        assertEquals(2, putLogEventsRequests.size());
+        for (int i=0; i < 2; i++) {
+            PutLogEventsRequest request = putLogEventsRequests.get(i);
+            assertEquals(calculateLogStreamName(THING_NAME, LOCAL_DEPLOYMENT_GROUP_NAME), request.logStreamName());
+            assertEquals("/aws/greengrass/UserComponent/" + AWS_REGION + "/UserComponentB", request.logGroupName());
+            assertNotNull(request.logEvents());
+            if (i == 1) {
```
Nelsonochoam, line 579: What does this condition mean?. Think about someone reading this test

PR #112, opened 2022-10-27
refactor: add unitOfWork

dgggIoT:
**Issue #, if available:**
#51 
**Description of changes:**
Refactor the current change so that we will iterate all components before processing each work
This will be used for future feature of uploading active file to avoid endless running into the same component.
Pass all current tests

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -584,23 +584,16 @@ public class LogManagerService extends PluginService {
     @SuppressWarnings("PMD.CollapsibleIfStatements")
     private void processLogsAndUpload() throws InterruptedException {
         while (true) {
-            // If there is already an upload ongoing, don't do anything. Wait for the next schedule to trigger to
-            // upload new logs.
+            //TODO: this is only done for passing the current text. But in practise, we don`t need to intentionally
+            // sleep here.
             if (!isCurrentlyUploading.compareAndSet(false, true)) {
-                //TODO: Use a CountDownLatch so that we could await here instead of sleep?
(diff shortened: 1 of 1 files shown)


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -584,23 +584,16 @@ private void processCloudWatchAttemptLogInformation(Map<String, Set<String>> com
...
                 continue;
             }
-            AtomicReference<Optional<ComponentLogFileInformation>> componentLogFileInformation =
-                    new AtomicReference<>(Optional.empty());
-            AtomicBoolean isActiveFileCompleted = new AtomicBoolean(false);
+            List<ComponentLogFileInformation> unitsOfWork = new ArrayList<>();
             // Get the latest known configurations because the componentLogConfigurations can change if a new
             // configuration is received from the customer.
             for (ComponentLogConfiguration componentLogConfiguration : componentLogConfigurations.values()) {
-                if (componentLogFileInformation.get().isPresent()) {
```
Nelsonochoam, line 600: **Here is the reason why these changes are needed**

The current implementation builds a single `componentLogFileInformation` breaks the for loop on line 596 as soon as it processes 1 item from the `componentLogConfigurations`. This makes some assumptions about the logs and when adding the code for processing the active file.

The current implementation works because we are not processing the active log file so line 607 was returning empty and it would process the next component configuration on the next run of the while loop. When processing the active log file we run into a scenario where the line 607 never returns true causing other component logs to not be processed.

This is just a steps towards the right direction and it will be further refactored but without this uploading the active file for multiple components won't work

Nelsonochoam (approved):

junfuchen99 (approved):

PR #114, opened 2022-10-31
feat: handle the upgrading from older version

dgggIoT:
**Issue #, if available:**
#51 
**Description of changes:**
Handle the upgrade scenarios for customer from older version. 
The fileHash does not exist in older version. And fileName is deprecated from older version. 

This will be tested in UAT
**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -413,11 +413,6 @@ public class LogManagerService extends PluginService {
                     CurrentProcessingFileInformation.builder().build();
             currentProcessingComponentTopics.iterator().forEachRemaining(node ->
                     currentProcessingFileInformation.updateFromTopic((Topic) node));
-            // If upgrade from older version, then the fileHash does not exist but fileName exists, then
-            // transfer from fileName to fileHash. If no fileName, then keep the fileHash as empty.
-            if (Utils.isEmpty(currentProcessingFileInformation.getFileHash())) {
-                currentProcessingFileInformation.convertNameToHash();
-            }
             // Only store the processing information when the fileHash is not empty or null.
             if (Utils.isNotEmpty(currentProcessingFileInformation.getFileHash())) {
                 componentCurrentProcessingLogFile.put(componentName, currentProcessingFileInformation);
@@ -873,6 +868,9 @@ public class LogManagerService extends PluginService {
                 default:
                     break;
             }
+            if (Utils.isEmpty(fileHash) && Utils.isNotEmpty(fileName)) {
+                convertNameToHash();
+            }
         }
 
         public static CurrentProcessingFileInformation convertFromMapOfObjects(
@@ -889,20 +887,18 @@ public class LogManagerService extends PluginService {
                     .build();
         }
 
-        public void convertNameToHash() {
-            if (Utils.isEmpty(fileHash) && Utils.isNotEmpty(fileName)) {
-                try {
-                    File file = new File(Paths.get(fileName).toUri());
-                    if (file.exists()) {
-                        fileHash = LogFile.of(file).hashString();
-                    } else {
-                        logger.atWarn().log("File {} does not exist.", file.getAbsolutePath());
-                    }
-                } catch (InvalidPathException e) {
-                    logger.atError().cause(e).log("Path string cannot be converted to a path.");
-                } catch (IllegalArgumentException e) {
-                    logger.atError().cause(e).log("URI of the file is invalid.");
+        private void convertNameToHash() {
+            try {
+                File file = new File(Paths.get(fileName).toUri());
+                if (file.exists()) {
+                    fileHash = LogFile.of(file).hashString();
+                } else {
+                    logger.atWarn().log("File {} does not exist.", file.getAbsolutePath());
                 }
+            } catch (InvalidPathException e) {
+                logger.atError().cause(e).log("Path string cannot be converted to a path.");
+            } catch (IllegalArgumentException e) {
+                logger.atError().cause(e).log("URI of the file is invalid.");
             }
         }
     }
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -12,6 +12,7 @@ import com.aws.greengrass.config.UpdateBehaviorTree;
 import com.aws.greengrass.config.WhatHappened;
 import com.aws.greengrass.dependency.ImplementsService;
 import com.aws.greengrass.lifecyclemanager.PluginService;
+import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.logging.impl.config.LogConfig;
 import com.aws.greengrass.logmanager.exceptions.InvalidLogGroupException;
@@ -39,6 +40,7 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
+import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardWatchEventKinds;
@@ -55,7 +57,6 @@ import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
@@ -412,6 +413,11 @@ public class LogManagerService extends PluginService {
                     CurrentProcessingFileInformation.builder().build();
             currentProcessingComponentTopics.iterator().forEachRemaining(node ->
                     currentProcessingFileInformation.updateFromTopic((Topic) node));
+            // If upgrade from older version, then the fileHash does not exist but fileName exists, then
+            // transfer from fileName to fileHash. If no fileName, then keep the fileHash as empty.
+            if (Utils.isEmpty(currentProcessingFileInformation.getFileHash())) {
+                currentProcessingFileInformation.convertNameToHash();
+            }
             // Only store the processing information when the fileHash is not empty or null.
             if (Utils.isNotEmpty(currentProcessingFileInformation.getFileHash())) {
                 componentCurrentProcessingLogFile.put(componentName, currentProcessingFileInformation);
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -860,8 +860,15 @@ public void updateFromTopic(Topic topic) {
                     lastModifiedTime = Coerce.toLong(topic);
                     break;
                 case PERSISTED_CURRENT_PROCESSING_FILE_HASH:
-                    //TODO: the scenario of upgrading from older version needs to be handled in next PR.
+                    // If upgrade from older version, then the fileHash does not exist but fileName exists, then
+                    // transfer from fileName to fileHash.
+                    String savedName = Coerce.toString(PERSISTED_CURRENT_PROCESSING_FILE_NAME);
+                    String savedHash = Coerce.toString(topic);
```
Nelsonochoam, line 866: This line and the line below are doing the same thing

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -860,8 +860,15 @@ public void updateFromTopic(Topic topic) {
...
                     fileHash = Coerce.toString(topic);
+                    if (savedHash.isEmpty() && !savedName.isEmpty()) {
```
Nelsonochoam, line 868: Better to check for null given there might be no value for that topic `Objects.isNull`

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -860,8 +861,14 @@ public void updateFromTopic(Topic topic) {
...
                 case PERSISTED_CURRENT_PROCESSING_FILE_HASH:
-                    //TODO: the scenario of upgrading from older version needs to be handled in next PR.
+                    // If upgrade from older version, then the fileHash does not exist but fileName exists, then
+                    // transfer from fileName to fileHash.
+                    String savedName = Coerce.toString(PERSISTED_CURRENT_PROCESSING_FILE_NAME);
                     fileHash = Coerce.toString(topic);
+                    if (Objects.isNull(fileHash) && !Objects.isNull(savedName)) {
+                        LogFile savedFile = new LogFile(Paths.get(savedName).toUri());
+                        fileHash = savedFile.hashString();
+                    }
```
MikeDombo, line 871: this can throw an exception. Make sure that we're handling it correctly.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -860,8 +861,14 @@ public void updateFromTopic(Topic topic) {
                     lastModifiedTime = Coerce.toLong(topic);
                     break;
                 case PERSISTED_CURRENT_PROCESSING_FILE_HASH:
-                    //TODO: the scenario of upgrading from older version needs to be handled in next PR.
+                    // If upgrade from older version, then the fileHash does not exist but fileName exists, then
+                    // transfer from fileName to fileHash.
+                    String savedName = Coerce.toString(PERSISTED_CURRENT_PROCESSING_FILE_NAME);
```
MikeDombo, line 866: This isn't doing what you want. This will give you the string "currentProcessingFileName", which will always be non-null.

MikeDombo (changes requested):
This pull request has a bug which means that it would never work properly. This shows that it has not been appropriately tested. Please add a test to cover this scenario and verify that it 1) finds the bug and then 2) you resolve the bug and the test passes.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
dgggIoT: Sorry, what exception? Is that something expected?

MikeDombo: `LogFile` is a `File`. `File`'s constructor can throw exceptions. `Paths.get()` can also throw exceptions, such as `IllegalArgumentException` these must be handled appropriately so we don't kill log manager.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -881,5 +888,22 @@ public static CurrentProcessingFileInformation convertFromMapOfObjects(
                             .get(PERSISTED_CURRENT_PROCESSING_FILE_HASH)))
                     .build();
         }
+
+        public void convertNameToHash() {
```
Nelsonochoam, line 892: I don't think this is a good idea. Now callers have to remember to call this in order for it to work as intended. I liked how it was before given it has self contained inside the `CurrentProcessingFileInformation` now this has to be called outside of this class and if it is not, it breaks the internal functionality of this (breaking encapsulation) and opening the door for bugs 🐛

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -860,7 +868,6 @@ public void updateFromTopic(Topic topic) {
                     lastModifiedTime = Coerce.toLong(topic);
                     break;
                 case PERSISTED_CURRENT_PROCESSING_FILE_HASH:
-                    //TODO: the scenario of upgrading from older version needs to be handled in next PR.
                     fileHash = Coerce.toString(topic);
```
Nelsonochoam, line 871: By not calling the logic you added below here you are opening the door for this not setting the correct file hash. For example if on the previous file there was a `PERSISTED_CURRENT_PROCESSING_FILE_NAME` and no hash, when you call this it won't change that file name to a hash if (possible)

dgggIoT: The reason I put it out of the switch is that there is no guarantee for the order of the topic. 
1. If the fileName topic is called before the fileHash, then the embedded conversion works.
2. But if fileHash topic is called before the fileName, then there is no fileName to be used for locate the file.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
dgggIoT: Agree. I can put it back inside the original place, but move the order a little bit. Please check next commit.

Nelsonochoam (dismissed):

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -881,5 +881,30 @@ public static CurrentProcessingFileInformation convertFromMapOfObjects(
...
+            }
+
+            Topic nameTopic = topics.find(PERSISTED_CURRENT_PROCESSING_FILE_NAME);
+
+            if (nameTopic == null) {
+                return null;
+            }
+
+            Path filePath = Paths.get(Coerce.toString(nameTopic));
+            File file = filePath.toFile();
```
MikeDombo, line 900: this can throw, we must handle that. Even though the topic itself is not null, the topic's value might be null.

`Paths.get` can throw `InvalidFilePathException` which must be handled.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -881,5 +881,30 @@ public static CurrentProcessingFileInformation convertFromMapOfObjects(
                             .get(PERSISTED_CURRENT_PROCESSING_FILE_HASH)))
                     .build();
         }
+
+        private String getFileHashFromTopic(Topic topic) {
+            Topics topics = topic.parent;
+            Topic hashTopic = topics.find(PERSISTED_CURRENT_PROCESSING_FILE_HASH);
+
+            if (hashTopic != null) {
+                return Coerce.toString(hashTopic);
```
MikeDombo, line 890: this may return `null` is that OK and handled correctly by consumers?

dgggIoT: https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/main/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java#L415
It is handled here.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
dgggIoT: Agree, will add the exception.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -833,6 +835,7 @@ static class CurrentProcessingFileInformation {
         private long lastModifiedTime;
         @JsonProperty(PERSISTED_CURRENT_PROCESSING_FILE_HASH)
         private String fileHash;
+        private static final Logger logger = LogManager.getLogger(LogManagerService.class);
```
MikeDombo, line 838: do not create this. A logger already exists. `logger`

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -881,5 +884,34 @@ public static CurrentProcessingFileInformation convertFromMapOfObjects(
...
+                File file = filePath.toFile();
+
+                if (!file.exists() || !file.isFile()) {
+                    return null;
+                }
+
+                LogFile logFile = LogFile.of(file);
+                return logFile.hashString();
+            } catch (InvalidPathException e) {
+                logger.atWarn().cause(e).log("File name is not a valid path.");
```
MikeDombo, line 912: no period at the end of logs, we add a period automatically.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -881,5 +884,34 @@ public static CurrentProcessingFileInformation convertFromMapOfObjects(
...
+                return Coerce.toString(hashTopic);
+            }
+
+            Topic nameTopic = topics.find(PERSISTED_CURRENT_PROCESSING_FILE_NAME);
+
+            if (nameTopic == null) {
+                return null;
+            }
+            try {
+                Path filePath = Paths.get(Coerce.toString(nameTopic));
```
MikeDombo, line 902: this will cause a null pointer exception if the value of `nameTopic` is null

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
dgggIoT: The class is static, so it cannot use the existing logger.

MikeDombo: If we want to use a logger, use the right logger, which comes from the log manager service class. In order to use it, make this class non-static, then fix the issues that that makes (looks like only a test is impacted).

Nelsonochoam, line 838:
```suggestion
        private static final Logger logger = LogManager.getLogger(CurrentProcessingFileInformation.class);
```

dgggIoT: I just tried to make the class non-static, it raise another issue "@Builder is not supported on non-static nested classes", and this builder is used in several places. 
Can we use the logger as @Nelsonochoam  suggested?

MikeDombo: yes

MikeDombo (approved):

Nelsonochoam (approved):

PR #115, opened 2022-11-02
fix: fix flaky test

Nelsonochoam:
**Description of changes:**

Fix flaky test that was just randomly sleeping and occasionally failing

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/SpaceManagementTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/SpaceManagementTest.java
@@ -36,6 +36,7 @@ import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.time.Duration;
 import java.time.Instant;
 import java.time.format.DateTimeParseException;
 import java.time.temporal.ChronoUnit;
@@ -44,8 +45,10 @@ import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
 import static com.aws.greengrass.integrationtests.logmanager.util.LogFileHelper.createTempFileAndWriteData;
-import static com.aws.greengrass.logmanager.util.TestUtils.eventuallyTrue;
 import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector.ignoreExceptionOfType;
+import static com.github.grantwest.eventually.EventuallyLambdaMatcher.eventuallyEval;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.lenient;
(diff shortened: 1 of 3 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/SpaceManagementTest.java
```diff
@@ -120,27 +118,34 @@ void setupKernel(Path storeDirectory)
     @Test
     void GIVEN_user_component_config_with_space_management_WHEN_space_exceeds_THEN_excess_log_files_are_deleted()
             throws Exception {
-        tempDirectoryPath = Files.createDirectory(tempRootDir.resolve("IntegrationTestsTemporaryLogFiles"));
```
Nelsonochoam, line 123: I agree this is not the best test, but my goal was just to make what we had not flaky

src/test/java/com/aws/greengrass/logmanager/util/TestUtils.java
```diff
@@ -35,4 +43,23 @@ public static LogFile createLogFileWithSize(URI uri, int bytesNeeded) throws IOE
         writeFile(file, bytesArray);
         return file;
     }
+
```
Nelsonochoam, line 46: We have this in a bunch of other codebases. Eventually we should move it to the nucleus

MikeDombo: Can't we use the eventually matcher? https://github.com/grantwest/eventually-matchers

MikeDombo (approved):

src/test/java/com/aws/greengrass/logmanager/util/TestUtils.java
Nelsonochoam: Did not know that was a thing. 😂 I will check it out

dgggIoT (dismissed):

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/SpaceManagementTest.java
```diff
@@ -120,27 +121,33 @@ void setupKernel(Path storeDirectory)
...
+        for (int i = 0; i < 15; i++) {
+            createTempFileAndWriteData(tempDirectoryPath, "integTestRandomLogFiles.log_", "");
         }
 
         Pattern logFileNamePattern = Pattern.compile("^integTestRandomLogFiles.log\\w*");
         LogFileGroup logFileGroup = LogFileGroup.create(logFileNamePattern, tempDirectoryPath.toUri(), mockInstant);
-        assertEquals(10, logFileGroup.getLogFiles().size());
+        // Then
+
+        assertThat("Log group size not below 105 kb",() -> {
```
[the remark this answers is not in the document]
Nelsonochoam: I don't want to use exact numbers because there is a lot of magic here. If we were able to pass the exact number of bytes we are writing into each file then I'd say sure but for now checking that the setting does what it does is good enough

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -110,6 +110,17 @@ public LogFileGroup syncDirectory() throws InvalidLogGroupException {
         return create(this.filePattern, this.directoryURI, this.lastUpdated);
     }
 
+    /**
+     * Returns the size in bytes of all the contents being tracked on by the log group.
+     */
+    public long byteSize() {
```
MikeDombo, line 116: `totalSizeBytes`/`totalSizeInBytes`

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/SpaceManagementTest.java
MikeDombo, line 143: "log group size should eventually be less than 105 KB" - you're checking for <=105 to be eventually true, so I don't see how this text is currently accurate.

MikeDombo (approved):

dgggIoT (approved):

PR #117, opened 2022-11-05
fix: the duplicate bug and the NPE bug for racing conditions

dgggIoT:
**Issue #, if available:**
#51 
**Description of changes:**
This is the fix for two bugs.

Do not merge, it is not cleaned up. 
**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
+++ b/src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
@@ -9,6 +9,7 @@ import com.aws.greengrass.deployment.DeviceConfiguration;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.GreengrassLogMessage;
 import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.logmanager.exceptions.InvalidLogGroupException;
 import com.aws.greengrass.logmanager.model.CloudWatchAttempt;
 import com.aws.greengrass.logmanager.model.CloudWatchAttemptLogFileInformation;
 import com.aws.greengrass.logmanager.model.CloudWatchAttemptLogInformation;
@@ -124,14 +125,24 @@ public class CloudWatchAttemptLogsProcessor {
                 .replace("{componentName}", componentLogFileInformation.getName());
         attempt.setLogGroupName(logGroupName);
         String logStreamName = getLogStreamName(thingName);
-        LogFileGroup logFileGroup = componentLogFileInformation.getLogFileGroup();
         // Run the loop until all the log files from the component have been read or the max message
         // size has been reached.
         while (!componentLogFileInformation.getLogFileInformationList().isEmpty() && !reachedMaxSize.get()) {
             LogFileInformation logFileInformation = componentLogFileInformation.getLogFileInformationList().get(0);
             LogFile logFile = logFileInformation.getLogFile();
-            long startPosition = logFileInformation.getStartPosition();
             String fileHash = logFileInformation.getFileHash();
+            long startPosition = logFileInformation.getStartPosition();
+            LogFileGroup logFileGroup = componentLogFileInformation.getLogFileGroup();
+            // If the race condition happen, meaning file get rotated between scanning file and reading file, then
+            // let us update logFileGroup and get the new file object.
+            if (!fileHash.equals(logFile.hashString())) {
+                try {
+                    logFileGroup = logFileGroup.syncDirectory();
+                    logFile = logFileGroup.getFile(fileHash);
+                } catch (InvalidLogGroupException e) {
+                    logger.atDebug().cause(e).log("Invalid log group");
+                }
+            }
             //This has been handled in the service, but leave here to prevent processor crash
             if (logFile.isEmpty() || startPosition == logFile.length()) {
                 componentLogFileInformation.getLogFileInformationList().remove(0);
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -458,12 +458,8 @@ private void handleCloudWatchAttemptStatus(CloudWatchAttempt cloudWatchAttempt)
...
         completedLogFilePerComponent.forEach((componentName, completedFiles) -> {
             completedFiles.forEach(file -> {
-                if (!lastComponentUploadedLogFileInstantMap.containsKey(componentName)
-                        || lastComponentUploadedLogFileInstantMap.get(componentName)
-                        .isBefore(Instant.ofEpochMilli(file.lastModified()))) {
-                    lastComponentUploadedLogFileInstantMap.put(componentName,
-                            Instant.ofEpochMilli(file.lastModified()));
-                }
+                updatelastComponentUploadedLogFile(lastComponentUploadedLogFileInstantMap, componentName,
+                        file.lastModified());
```
dgggIoT, line 462: Extract this method so that we can reuse in other place

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -527,6 +523,7 @@ private void processCloudWatchAttemptLogInformation(Map<String, Set<LogFile>> co
             String fileName = file.getAbsolutePath();
             // If we have completely read the file, then we need add it to the completed files list and remove it
             // it (if necessary) for the current processing list.
+            // .log, .log1, .log2, .log3
```
dgggIoT, line 526: Please ignore this, will delete.

MikeDombo:
We identified new issues on unchanged lines of code. Navigate to the Amazon CodeGuru Reviewer console to view the recommendations to fix them.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -634,6 +652,9 @@ private void processLogsAndUpload() throws InterruptedException {
 
                             if (startPosition < file.length()) {
                                 unitOfWork.getLogFileInformationList().add(logFileInformation);
+                            } else if (startPosition == file.length() && !logFileGroup.isActiveFile(file)) {
+                                updatelastComponentUploadedLogFile(lastComponentUploadedLogFileInstantMap,
+                                                                    componentName, file.lastModified());
```
dgggIoT, line 657: Here is the fix of the data duplicate bugs. Here is some summaries: 
1. In the first scan: When the active file (with hash1) is fully written and processed but not rotated, the startPos will be sat as the end of the file. After uploading, this file is saved as currentProcessing file
2. In the second scan: this file (with hash1) is rotated but still can be picked up the right startPos because it is marked as the currentProcessing file. But when we process the file, we will skip it because startPos == file.length(). After uploading, the currentProcessing file will be pointed to a new file.
3. In the third scan: the file (with hash1) will be picked up again, because the lastUploaded file didn`t update, but this time, we cannot sync the startPos. Then duplicate happens. 
 
Why this fix the bug: 
Once we figure out this file (with hash1) is rotated and startPos == file.length, we treat it as the lastUploaded file, so we try to update the global value. This is before we skip it in the processing steps.

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -148,10 +158,15 @@ public CloudWatchAttempt processLogFiles(ComponentLogFileInformation componentLo
                          StandardCharsets.UTF_8))) {
                 r.setInitialPosition(startPosition);
                 chan.position(startPosition);
+                //TODO: this only used for verifying the bugs
+                if (startPosition > logFile.length()) {
```
dgggIoT, line 162: Please ignore this, This will be cleaned up.

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -124,22 +125,31 @@ public CloudWatchAttempt processLogFiles(ComponentLogFileInformation componentLo
...
+            // If the race condition happen, meaning file get rotated between scanning file and reading file, then
+            // let us update logFileGroup and get the new file object.
+            if (!fileHash.equals(logFile.hashString())) {
+                try {
+                    logFileGroup = logFileGroup.syncDirectory();
+                    logFile = logFileGroup.getFile(fileHash);
+                } catch (InvalidLogGroupException e) {
+                    logger.atDebug().cause(e).log("Invalid log group");
+                }
+            }
```
dgggIoT, line 145: THis is the fix for the NPE bug causing the LM breaks: 
WHy it happen: 
The file object is passed to this class after scanning and creating the logFileGroup. However, it got rotated before we create the bufferReader for it. Although it depends on how customers rotate it, but for current rotating policy, the file object now points to another file, that is the newly-created active file. THen we have the bug of the startPos mismatch. 

How it fix:
If the fileHash we pass from the scanning steps match the current file object, then it is not rotated. We keep the same. 
Otherwise, we will sync the logFIleGroup, and use the fileHash as the identifier to links to the target file.

dgggIoT: Note: 
This change still allow the possibility of the NPE happen, even though it is super low but not 0. 
Another addition fix for it is to handle the NPE nicely so we just simply skip this file if the startPos mismatch happen. I didn`t apply this fix so far is to intentionally test if the current fix reduce dramtically the possibility of the issue. Will apply the second fix once we confirm the first fix is correct.

PR #119, opened 2022-11-05
Bug fixes

Nelsonochoam:
**Issue #, if available:**

**Description of changes:**

**Checklist:**


src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -104,13 +105,115 @@ String getLogStreamName(String thingName) {
...
+    }
+
     /**
      * Gets CW input log events from the component which processLogFiles need to be uploaded to CloudWatch.
      *
      * @param componentLogFileInformation log files information for a component to read logs from.
      * @return CloudWatch attempt containing information needed to upload logs from the component to the cloud.
      */
-    public CloudWatchAttempt processLogFiles(ComponentLogFileInformation componentLogFileInformation) {
+    public CloudWatchAttempt processLogFilesDeprecated(ComponentLogFileInformation componentLogFileInformation) {
```
Nelsonochoam, line 216: Tagged this as deprecated for now we need to go and update the tests that are directly calling this method

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -84,6 +111,28 @@ private String readBytesToString() {
         return "";
     }
 
+    /**
+     * Get the hash of the logfile with target lines.
+     * @param chan a nyte channel that allow us to read the file
+     * @return the calculated hash value of the logfile, empty string if not enough lines for digest.
+     */
+    public String hashStringWithChannel(SeekableByteChannel chan) {
```
Nelsonochoam, line 119: You could just call this `hashString` and overload the other method. Ideally we delete the other method maybe on a next PR

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
Nelsonochoam: We will delete this in the next PR

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -55,6 +57,31 @@ public static LogFile[] of(File... files) {
         return Arrays.stream(files).map(LogFile::of).toArray(LogFile[]::new);
     }
 
+    @SuppressWarnings("PMD.CloseResource")
+    private String readBytesToStringV2(SeekableByteChannel chan) {
```
Nelsonochoam, line 61:
```suggestion
    private String readBytesToString(SeekableByteChannel chan) {
```

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
Nelsonochoam, line 119:
```suggestion
    public String hashString(SeekableByteChannel chan) {
```

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -104,13 +105,115 @@ String getLogStreamName(String thingName) {
...
+     * @param chan a byte channel that allows to read a file from the specified position
+     *
+     * @throws IOException if it fails to read the file or the file has been deleted
+     */
+    public CloudWatchAttempt processLogFiles(
+            CloudWatchAttempt attempt, ComponentLogFileInformation cpInfo, LogFileInformation fileInfo,
+            SeekableByteChannel chan) throws IOException {
+        long startPosition = fileInfo.getStartPosition();
+        boolean reachedMaxSize = false;
+        AtomicInteger totalBytesRead = new AtomicInteger();
```
jbutler, line 138: do you know why this is an atomic integer? was the previous code being called in parallel?

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
jbutler: the new function shouldn't have this name, so do we need to rename it? it just makes the PR bigger for no reason

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -458,12 +460,8 @@ private void handleCloudWatchAttemptStatus(CloudWatchAttempt cloudWatchAttempt)
...
         completedLogFilePerComponent.forEach((componentName, completedFiles) -> {
             completedFiles.forEach(file -> {
-                if (!lastComponentUploadedLogFileInstantMap.containsKey(componentName)
-                        || lastComponentUploadedLogFileInstantMap.get(componentName)
-                        .isBefore(Instant.ofEpochMilli(file.lastModified()))) {
-                    lastComponentUploadedLogFileInstantMap.put(componentName,
-                            Instant.ofEpochMilli(file.lastModified()));
-                }
+                updatelastComponentUploadedLogFile(lastComponentUploadedLogFileInstantMap, componentName,
+                        file.lastModified());
```
jbutler, line 464: what's this? can this be a separate PR? I _really_ want us to get in the habit of not lumping a bunch of unrelated stuff together

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -601,21 +619,57 @@ private void processLogsAndUpload() throws InterruptedException {
...
+                    // 2. Collect the contents from the files that need to be uploaded in a CloudWatchAttempt
+
+                    CloudWatchAttempt groupAttempt = new CloudWatchAttempt();
+                    groupAttempt.setLogGroupName(
+                            logsProcessor.getLogGroupName(componentLogConfiguration.getComponentType(), componentName));
+                    HashSet<String> processedFileHashes = new HashSet<>();
+
+                    for (LogFile file : logFileGroup.getLogFiles()) {
+
+                        // Open the file while file before trying to process it given it might be rotated and we still
```
jbutler, line 639: nit: typo `open the file while file`

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -562,6 +560,24 @@ private void processCloudWatchAttemptLogInformation(Map<String, Set<LogFile>> co
...
     }
 
+    /**
+     * This updates the lastComponentUploadedLogFileInstantMap if the current component has a newly uploaded file,
+     * which the lastModified time is larger than saved value of lastModified time.
+     * @param lastComponentUploadedLogFileInstantMap The instant map of all components.
+     * @param componentName componentName.
+     * @param lastModified the lastModified time of the file.
+     */
+    private void updatelastComponentUploadedLogFile(Map<String, Instant> lastComponentUploadedLogFileInstantMap,
```
Nelsonochoam, line 570:
```suggestion
    private void updateLastComponentUploadedLogFile(Map<String, Instant> lastComponentUploadedLogFileInstantMap,
```

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -601,21 +619,57 @@ private void processLogsAndUpload() throws InterruptedException {
...
+                            //  display out of order in cloudwatch
+                            //
+                            if (processedFileHashes.contains(fileHash)) {
+                                break;
+                            }
+
+                            // The file must contain enough lines for digest hash, otherwise fileHash is empty string
+                            if (HASH_VALUE_OF_EMPTY_STRING.equals(fileHash)) {
+                                return;
+                            }
```
jbutler, line 669: This is pretty fragile. If there isn't enough data then the hashing function shouldn't return a hash

Can fix in a follow up since this was existing behavior

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -633,26 +687,33 @@ private void processLogsAndUpload() throws InterruptedException {
                                     .build();
 
                             if (startPosition < file.length()) {
-                                unitOfWork.getLogFileInformationList().add(logFileInformation);
+                                componentLogFileInformation.getLogFileInformationList().add(logFileInformation);
+                                logsProcessor.processLogFiles(
+                                        groupAttempt, componentLogFileInformation, logFileInformation, fileByteChannel);
+                            } else if (startPosition == file.length() && !logFileGroup.isActiveFile(file)) {
```
Nelsonochoam, line 693: why do we need this else if here?. This is being called after the upload occurs by the `handleCloudWatchAttemptStatus` handler. It seems fishy to be updating this here when the upload has not happened

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
dgggIoT: Shall we directly delete it now?

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
Nelsonochoam: It was an atomic integer before because it was being called inside a lambda and given it is being passed on `processLogLine` and then to other methods I didn't want to modify those methods yet and bloat this even more.

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
Nelsonochoam: @jbutler not sure I get your comment. Are you saying keep the old method being called `processLogFiles` and have the new one have a different method? I think the effect on the size of the PR is the same because we have to change the tests that directly called the old one to call the new one

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
dgggIoT: This is for the active file is processed completely in the last run, but rotated before the next run. So we will still get the startPosition to the end, but now this file is rotated and finished. So we need to handle the status. 
`handleCloudWatchAttemptStatus` only handle the file got processed, but the file mentioned above has no contents to be uploaded, so it will not get processed.  
This is the root cause of the duplicate logs.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
Nelsonochoam: I think this was a miss. Before it was inside of a lambda and calling return would skip the file this should be a `continue`

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
dgggIoT: K. Let me remove and leave if for a separate PR.

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
dgggIoT: Not sure. Theoratically, this is only one thread.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
Nelsonochoam: Mmmm if the active log file is process completely (meaning all the logs that are available in it) and it gets rotated on the next run. When finding the files and looping through them it computes the hashes and we would get the start position for the correct file. Maybe I am missing something but this shouldn't be required.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
dgggIoT: Yeah, good catch. It should be `continue`

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
Nelsonochoam: As I replied. The only reason for this to be an atomic integer was before before this was inside of lambda and all the methods that where called inside of a lambda expect an atomic integer on their signature. We can change it but I did not in the spirit of not adding more changes. Let's table changing this for the next PR unless you folks think otherwise

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
dgggIoT: We can discuss it in another PR. As @jbutler suggests, we will leave this change to a separate PR.

Nelsonochoam: This is new functionality you changed... I don't see how discussing it on another PR changes it. Could you provide an example on a scenario this would be required to be here?

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
jbutler: The new method isn't processing files. It's processing a single file. So there was no reason to rename this is what I was saying

Nelsonochoam: Got it, it was being called directly on test though the only real saving is not seeing that line change so from this what we can do is revert the naming of this and change the name of the one above to something like `processLogFile` or `addLogFileToAttempt`?

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -110,6 +112,29 @@ private String readBytesToString() {
...
+            }
+            // String stringToHash = readBytesToString();
+            String stringToHash = readBytesToString(chan);
+            if (!stringToHash.isEmpty()) {
+                fileHash = calculate(stringToHash);
+            }
+        }  catch (NoSuchAlgorithmException e) {
+            logger.atError().cause(e).log("The digest algorithm is invalid");
+        }
+        return fileHash;
```
MikeDombo, line 135: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

Similar code fragments were detected in the same file at the following lines: 122:135, 145:157.  
Refactoring can help improve code maintainability. Consider reducing duplicate code by extracting it into a separate method. You can then replace duplicated code with calls to this new method.

dgggIoT: Thanks, will have a separate PR to handle this refactor.

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -110,6 +111,28 @@ private String readBytesToString() {
...
```
MikeDombo, line 133: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

Similar code fragments were detected in the same file at the following lines: 120:133, 143:155.  
Refactoring can help improve code maintainability. Consider reducing duplicate code by extracting it into a separate method. You can then replace duplicated code with calls to this new method.

PR #130, opened 2022-11-08
feat: detect rotation when getting log group files

Nelsonochoam:
**Issue #, if available:**

**Description of changes:**
This PR sets the foundation for detecting if there was a rotation when reading the files that need to be processed from the file system.

* This PR doesn't add yet the generation of hard links but that would be encapsulated inside the creation of the log files via the LogFileGroup.
* The test needs some additional assertions to check that we got the correct file references but this will get added once the hard links are added.

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
+++ b/src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
@@ -25,16 +25,8 @@ public final class LogFileGroup {
     private final URI directoryURI;
     private final Instant lastUpdated;
 
-    /**
-     * Creates a log file group without initializing the logFiles. To initialize them you must call
-     * getLogFiles.
-     * @param logFiles - the log group log files
-     * @param filePattern - the pattern use to match files on the provided directoryURI
-     * @param directoryURI - the directory where log files are located
-     * @param lastUpdated - instant used to filter out log files last modified after its value
-     */
-    public LogFileGroup(List<LogFile> logFiles, Pattern filePattern, URI directoryURI, Instant lastUpdated) {
-        this.logFiles = logFiles;
+    private LogFileGroup(List<LogFile> files, Pattern filePattern, URI directoryURI, Instant lastUpdated) {
+        this.logFiles = files;
         this.filePattern = filePattern;
         this.directoryURI = directoryURI;
         this.lastUpdated = lastUpdated;
@@ -49,7 +41,7 @@ public final class LogFileGroup {
      * that do not match the log group pattern and have been last modified after the provided lastUpdated value.
      * @param allDirectoryLogFiles - files in a directory
      */
-    public List<LogFile> getLogFiles(File... allDirectoryLogFiles) {
+    List<LogFile> getLogFiles(File... allDirectoryLogFiles) {
         List<LogFile> files = getLogFilesInternal(allDirectoryLogFiles);
 
         if (haveRotated(files)) {
--- a/src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
+++ b/src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
@@ -8,27 +8,93 @@ import java.io.File;
 import java.net.URI;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -1,36 +1,105 @@
...
+    public List<LogFile> getLogFiles() {
+       return getLogFiles(listDirectory());
+    }
+
+    /**
+     * Gets the log files for the component, by listing all the files in the directory and filtering out the ones
+     * that do not match the log group pattern and have been last modified after the provided lastUpdated value.
+     * @param allDirectoryLogFiles - files in a directory
+     */
+    public List<LogFile> getLogFiles(File... allDirectoryLogFiles) {
```
Nelsonochoam, line 50: The reason I exposed this tests was so I could better simulate the issue we are trying to address or logs rotating as soon as we scan the directory for logs

src/test/java/com/aws/greengrass/logmanager/model/LogFileGroupTest.java
```diff
@@ -47,4 +56,43 @@ void GIVEN_log_files_THEN_find_the_active_file() throws IOException, Interrupted
...
+        assertEquals(directoryFiles.length, logFiles.size());
+
+        Path[] expectedPaths = {
+            directoryPath.resolve("test.log.2"),
+            directoryPath.resolve("test.log.1"),
+            directoryPath.resolve("test.log")
+        };
+
+        List<Path> filePaths = logFiles.stream().map(File::toPath).collect(Collectors.toList());
+        for (int i = 0; i < expectedPaths.length; i++) {
```
Nelsonochoam, line 91: This check is temporary I will update once logfiles have the hard links to ensure we are getting the expected files

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -8,29 +8,95 @@
...
+    /**
+     * Creates a log file group without initializing the logFiles. To initialize them you must call
+     * getLogFiles.
+     * @param logFiles - the log group log files
+     * @param filePattern - the pattern use to match files on the provided directoryURI
+     * @param directoryURI - the directory where log files are located
+     * @param lastUpdated - instant used to filter out log files last modified after its value
+     */
+    public LogFileGroup(List<LogFile> logFiles, Pattern filePattern, URI directoryURI, Instant lastUpdated) {
+        this.logFiles = logFiles;
```
jbutler, line 37: Not understanding this. I had thought the `LogFileGroup` would create the hardlinks which back the `List<LogFile>`?

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
jbutler: Making this package private instead of public may help to convey intent

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -8,29 +8,95 @@
...
+    private List<LogFile> getLogFilesInternal(File... allDirectoryLogFiles) {
+       return Arrays.stream(allDirectoryLogFiles)
+               .filter(logFile ->  filePattern.matcher(logFile.getName()).find())
+               .filter(logFile -> lastUpdated.isBefore(Instant.ofEpochMilli(logFile.lastModified())))
+               .map(LogFile::of)
+               .sorted(Comparator.comparingLong(File::lastModified))
+               .collect(Collectors.toList());
+    }
+
+    private boolean haveRotated(List<LogFile> computedFiles) {
```
jbutler, line 81: What happens if logs rotate but the set of filenames don't? I don't think comparing LogFiles will work

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
Nelsonochoam: Yes I mentioned on the description that will be the next PR

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
Nelsonochoam: Yeah this isn't fully there we would compare agains the hardlinks created. Next PR

PR #131, opened 2022-11-09
feat: log file v2 hardllinks

Nelsonochoam:
**Issue #, if available:**

**Description of changes:**

**Checklist:**


src/main/java/com/aws/greengrass/logmanager/model/LogFileV2.java
```diff
@@ -0,0 +1,203 @@
...
+     * regardless of the original file being rotated. One important consideration is that getting the original source
+     * path if the original file rotates is only possible if the file remains on the same folder it originally was
+     * created. Rotation mechanisms that rotate files to different folders will yield in an invalid source path
+     * and a IOException will be thrown.
+     *
+     * @param sourceFile - the source java File.
+     * @param hardLink - a File path to where the file hardlink should be stored
+     * @throws IOException - if we fail to get the source file path
+     */
+    public static LogFileV2 fromFile(File sourceFile, Path hardLink) throws IOException {
```
jbutler, line 48: Do we need this? I think we can reuse the original `LogFile` -- it doesn't need to change at all, actually. In fact we can revert the SeekableByteChannel change from the weekend and just open the file directly.

LogFileGroup will make sure LogFiles are hard links instead of pointing directly at the target

Thoughts?

Nelsonochoam: I think we both have the same thoughts my idea was first write something and prove it could work with tests and then replace it. We would reuse the original log file but I just wanted to first create the constructs and then replace it in places little by little

PR #136, opened 2022-11-10
feat: hardlink changes

Nelsonochoam:
**Description of changes:**
This changes replace the LogFile underlying implementation to work with hard links. There are a few things that still need to be addressed.

1) File deletion
2) Detect if the logs rotated just after creating the hard links
3) Adding a few more tests to ensure the core pieces work
  3.1) Setting the correct lastUploadedTime

Avoid getting wrong file data if the files rotate

Currently doing manual tests

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
+++ b/src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
@@ -1,13 +1,20 @@
 package com.aws.greengrass.logmanager.model;
 
+import com.aws.greengrass.logging.api.Logger;
+import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.logmanager.exceptions.InvalidLogGroupException;
 import com.aws.greengrass.util.Utils;
 import lombok.Getter;
 
 import java.io.File;
+import java.io.IOException;
 import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
@@ -17,6 +24,8 @@ import java.util.regex.Pattern;
 
 
 public final class LogFileGroup {
+    private static final Logger logger = LogManager.getLogger(LogFileGroup.class);
+    private static final String DEFAULT_HARDLINKS_PATH = "aws.greengrass.LogManager/hardlinks";
     @Getter
     private List<LogFile> logFiles;
     private static Map<String, LogFile> fileHashToLogFile;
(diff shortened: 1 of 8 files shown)


src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -49,29 +58,49 @@ public static LogFileGroup create(Pattern filePattern, URI directoryURI, Instant
             throw new InvalidLogGroupException(String.format("%s must be a directory", directoryURI));
         }
 
-        LogFile[] files = LogFile.of(folder.listFiles());
+        Path hardLinkPath = Paths.get(DEFAULT_HARDLINKS_PATH).resolve(folder.getName());
+        try {
+            Utils.createPaths(hardLinkPath);
+        } catch (IOException e) {
```
MikeDombo, line 64: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**: While wrapping the caught exception into a custom one, information about the caught exception is being lost, including information about the stack trace of the exception.

**Fix**: If the caught exception object does not contain sensitive information, consider passing it as the &#34;rootCause&#34; or inner exception parameter to the constructor of the new exception before throwing the new exception. (Note that not all exception constructors support inner exceptions. Use a wrapper exception that supports inner exceptions.) 
[Learn more](https://www.ibm.com/support/pages/best-practice-catching-and-re-throwing-java-exceptions)

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -1,75 +1,120 @@
...
             throws InvalidLogGroupException {
         File folder = new File(directoryURI);
-        fileHashToLogFile = new ConcurrentHashMap<>();
 
         if (!folder.isDirectory()) {
             throw new InvalidLogGroupException(String.format("%s must be a directory", directoryURI));
         }
 
-        LogFile[] files = LogFile.of(folder.listFiles());
+        Path componentHardlinksDirectory = hardLinksDirectory.resolve(filePattern.toString());
```
saranyailla, line 80: Is file pattern the right thing to use here as it  can be a regex? Also, why are we using a separate component hardlink directory for each component if we can do pattern matching there as well.

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -1,75 +1,120 @@
...
+        for (File file : files) {
+            boolean isModifiedAfterLastUpdatedFile =
+                    lastUpdated.isBefore(Instant.ofEpochMilli(file.lastModified()));
+            boolean isNameMatchPattern = filePattern.matcher(file.getName()).find();
+
+            if (file.isFile() && isModifiedAfterLastUpdatedFile && isNameMatchPattern) {
+                try {
+                    LogFile logfile = LogFile.of(file, componentHardlinksDirectory);
+                    allFiles.add(logfile);
+                    fileHashToLogFile.put(logfile.hashString(), logfile);
```
saranyailla, line 109: +1 for calculating hash after the other checks.

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
Nelsonochoam: It is not. That is why I mentioned the hard-link directory should be passed as a parameter but lets leave that for a separate pr because that would require test changes.

Using the file pattern here for now to make the hardlink dirs for each component unique for now.

The reason que are creating unique directories is because at the beginning we are deleting everything that is in that directory for simplicity and we wouldn't want to accidentally delete hard links for another component

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
Nelsonochoam: Don't know what this comment means but the hardlink gets memoized after the first time we calculate it to avoid additional reads

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
saranyailla: This makes having a separate component hardlink directory mandatory.

Nelsonochoam: Yeah, if we wanted to structure it that way. Alternatively we could look in a non unique hardlink directory for all the hardlinks that match the pattern and delete them before creating new ones

PR #145, opened 2022-11-12
feat: use hardlinks of log file during processing

saranyailla:
**Issue #, if available:**

**Description of changes:**
Created hardlinks for log file when collecting the log files to process and upload. 
This can prevent data loss due to rotation as the hardlink still persists even when the source log file is deleted. 

Taken mostly from: https://github.com/aws-greengrass/aws-greengrass-log-manager/pull/136

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -473,14 +473,14 @@ public class LogManagerService extends PluginService {
             }
             completedFiles.forEach(file -> {
                 try {
-                    boolean successfullyDeleted = Files.deleteIfExists(file.toPath());
+                    boolean successfullyDeleted = Files.deleteIfExists(Paths.get(file.getSourcePath()));
                     if (successfullyDeleted) {
-                        logger.atDebug().log("Successfully deleted file with name {}", file.getAbsolutePath());
+                        logger.atDebug().log("Successfully deleted file with name {}", file.getSourcePath());
                     } else {
-                        logger.atWarn().log("Unable to delete file with name {}", file.getAbsolutePath());
+                        logger.atWarn().log("Unable to delete file with name {}", file.getSourcePath());
                     }
                 } catch (IOException e) {
-                    logger.atError().cause(e).log("Unable to delete file with name: {}", file.getAbsolutePath());
+                    logger.atError().cause(e).log("Unable to delete file with name: {}", file.getSourcePath());
                 }
             });
         });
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -473,14 +473,14 @@ private void handleCloudWatchAttemptStatus(CloudWatchAttempt cloudWatchAttempt)
             }
             completedFiles.forEach(file -> {
                 try {
-                    boolean successfullyDeleted = Files.deleteIfExists(file.toPath());
+                    boolean successfullyDeleted = Files.deleteIfExists(Paths.get(file.getSourcePath()));
```
Nelsonochoam, line 476: We have to remember to avoid using the getSourcePath since it is not reliable with rotations. We chatted about this

Nelsonochoam (dismissed):

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -25,34 +25,42 @@ public class LogFile extends File {
...
-    public static LogFile[] of(File... files) {
-        if (files == null) {
-            return new LogFile[] {};
-        }
-        return Arrays.stream(files).map(LogFile::of).toArray(LogFile[]::new);
+    public static LogFile of(File sourceFile, Path hardLinkDirectory) throws IOException {
+        // Why do we need this? - Hardlinks can't get the path of their source file. We need to keep track of which path
+        // it was originally created with and in case the file on that path changes we need to scan the directory to
+        // check for the inode that matches the hardlink to get the correct path.
+        Path destinationPath = hardLinkDirectory.resolve(sourceFile.getName());
```
Nelsonochoam, line 60: In case someone asks why have this here. It is because we will need to add a method to check if the file it was originally tracking is not the same one as the one the hardlink is tracking

Nelsonochoam (dismissed):

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -107,4 +115,12 @@ public boolean isEmpty() {
         return Utils.isEmpty(this.hashString());
     }
 
+    /**
+     * Returns the source path of the file original file.
+     */
+    public String getSourcePath() {
```
jbutler, line 121: Let's not expose this. The fact that we're using hard links under the hood should be hidden as much as possible. Deletion is a special case. Let's handle that separately by exposing a delete method which can encapsulate the logic to handle rotation.

Nelsonochoam (approved):

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
Nelsonochoam: +1

Nelsonochoam: Could we fix this one a separate PR? - We still need to fix the delete anyways

dgggIoT (approved):

PR #148, opened 2022-11-14
feat: rollback to only uploading rotated files if we fail to create hardlinks

Nelsonochoam:
**Description of changes:**
We can not create hard links for log files being created on a volume separate to the one greengrass is running on. In the scenario where that happens we want to rollback and only upload the rotated log files and stop processing the active file. Effectively working how it used to work before

Ensure we are not breaking customers that have a setup where they track log files being written to a separate volume.

Manual tests, we will cover with unit and integration tests but we have to make more changes to make this scenarios more easily testable.

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
+++ b/src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
@@ -12,6 +12,7 @@ import java.net.URI;
 import java.nio.file.Path;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


dgggIoT (approved):

src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java
```diff
@@ -66,34 +71,88 @@ public static LogFileGroup create(ComponentLogConfiguration componentLogConfigur
...
+            logger.atDebug().kv("component", componentName)
                     .kv("directory", directoryURI)
                     .log("No component logs are found in the directory");
-            return new LogFileGroup(Collections.emptyList(), filePattern, new HashMap<>());
+            return new LogFileGroup(Collections.emptyList(), filePattern, new HashMap<>(), false);
         }
 
         Map<String, LogFile> fileHashToLogFileMap = new ConcurrentHashMap<>();
-        List<LogFile> allFiles = new ArrayList<>();
+        boolean isUsingHardlinks;
```
saranyailla, line 83: If we always try creating hardlinks first, initiate with True?

saranyailla (approved):

dgggIoT (approved):

PR #149, opened 2022-11-14
refactor: memoize hash string value

Nelsonochoam:
**Description of changes:**
Tiny change to avoid unnecessary  IO operations to compute the hash.

Manual tests. We will add more tests in a subsequent PR

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
+++ b/src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
@@ -16,6 +16,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.NoSuchAlgorithmException;
+import java.util.Objects;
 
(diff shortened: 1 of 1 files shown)


dgggIoT (approved):

src/main/java/com/aws/greengrass/logmanager/model/LogFile.java
```diff
@@ -96,19 +98,24 @@ private String readBytesToString() {
      * @return the calculated hash value of the logfile, empty string if not enough lines for digest.
      */
     public String hashString() {
-        String fileHash = HASH_VALUE_OF_EMPTY_STRING;
+        if (!Objects.equals(this.hash, HASH_VALUE_OF_EMPTY_STRING)) {
+            return this.hash;
+        }
```
saranyailla, line 103: Nit: rename this later to getHash()?

saranyailla (approved):

PR #150, opened 2022-11-15
feat: update delete after upload

Nelsonochoam:
**Description of changes:**
Hardlinks can't return the source path they are tracking. When we delete a file we want to ensure we are deleting the correct file and not the file path the hardlink got originally created with.

We might end up not deleting the files or deleting the wrong file.

Unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -472,15 +472,11 @@ public class LogManagerService extends PluginService {
                 return;
             }
             completedFiles.forEach(file -> {
-                try {
-                    boolean successfullyDeleted = Files.deleteIfExists(Paths.get(file.getSourcePath()));
-                    if (successfullyDeleted) {
-                        logger.atDebug().log("Successfully deleted file with name {}", file.getSourcePath());
-                    } else {
-                        logger.atWarn().log("Unable to delete file with name {}", file.getSourcePath());
-                    }
-                } catch (IOException e) {
-                    logger.atError().cause(e).log("Unable to delete file with name: {}", file.getSourcePath());
+                boolean successfullyDeleted = file.delete();
+                if (successfullyDeleted) {
+                    logger.atDebug().log("Successfully deleted file with name {}", file.getName());
+                } else {
+                    logger.atWarn().log("Unable to delete file with name {}", file.getName());
                 }
             });
         });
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -472,15 +472,11 @@ private void handleCloudWatchAttemptStatus(CloudWatchAttempt cloudWatchAttempt)
...
-                try {
-                    boolean successfullyDeleted = Files.deleteIfExists(Paths.get(file.getSourcePath()));
-                    if (successfullyDeleted) {
-                        logger.atDebug().log("Successfully deleted file with name {}", file.getSourcePath());
-                    } else {
-                        logger.atWarn().log("Unable to delete file with name {}", file.getSourcePath());
-                    }
-                } catch (IOException e) {
-                    logger.atError().cause(e).log("Unable to delete file with name: {}", file.getSourcePath());
+                boolean successfullyDeleted = file.delete();
```
dgggIoT, line 475: Is this `file` referring to the hardlink one or the original one?

Nelsonochoam: Could be any. It is referring to a log file. If we rolled back to the old functionality it refers to a regular java file if not it refers to the hardlink

dgggIoT: Ignore above comment, just see the `delete` method you defined.

src/test/java/com/aws/greengrass/logmanager/model/LogFileTest.java
```diff
@@ -97,6 +97,59 @@ void GIVEN_log_file_with_more_than_target_lines_but_two_lines_WHEN_calculate_fil
...
+                        .filter(File::isFile)
+                        .toArray(File[]::new);
+        assertEquals(directoryFiles.length, 1);
+
+        File remainingFile = directoryFiles[0];
+        assertEquals(remainingFile.toPath(), file.toPath());
+    }
+
+    @Test
+    void GIVEN_logFileTrackingRegularFile_WHEN_trackedFileRotates_THEN_itDeletesTheWrongFile() throws IOException {
```
saranyailla, line 130: Nit: Add a comment her that this happens in older versions of LM or if the hardlinks are not used.

saranyailla (approved):

dgggIoT (approved):

PR #151, opened 2022-11-15
refactor: cleanup and add some unit tests for log group and log file

Nelsonochoam:
**Description of changes:**
Add a few unit tests for the LogFile and the LogFileGroup and also made it so the log group does not return files that do not have a hash so we don't have to be sprinkling that logic around the codebase

Encapsulate logic related to which files should be processed.

Unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -73,7 +73,6 @@ import javax.inject.Inject;
 
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
 import static com.aws.greengrass.logmanager.LogManagerService.LOGS_UPLOADER_SERVICE_TOPICS;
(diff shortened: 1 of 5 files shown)


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -609,44 +608,42 @@ private void processLogsAndUpload() throws InterruptedException {
...
                             .build();
 
-                    unitsOfWork.add(unitOfWork);
+                    componentMetadata.add(logFileInfo);
 
                     logFileGroup.forEach(file -> {
                         long startPosition = 0;
                         String fileHash = file.hashString();
-                        // The file must contain enough lines for digest hash, otherwise fileHash is empty string
-                        if (!HASH_VALUE_OF_EMPTY_STRING.equals(fileHash)) {
```
Nelsonochoam, line 626: This red block is just removing this if statement. The log group will filter out files that are no ready to be processed

dgggIoT (approved):

saranyailla (approved):

PR #165, opened 2022-11-30
fix: add missing steps

Nelsonochoam:
**Description of changes:**
Now that we can reference the locally installed version of OTF there are a few steps that we were missing that are available. This PR just adds those steps in to the tests the were already ported.

Other changes:
* Fix checkstyle violations
* Remove bad step

Ensures we wait first for the LogManager to be running

### DIFF
--- a/uat/src/main/java/com.aws.greengrass/FileSteps.java
+++ b/uat/src/main/java/com.aws.greengrass/FileSteps.java
@@ -39,7 +39,7 @@ public class FileSteps {
     private final Platform platform;
     private final TestContext testContext;
     private final ScenarioContext scenarioContext;
-    private static final String activeFile = "ActiveFile";
+    private static final String ACTIVEFILE = "ActiveFile";
 
     /**
      * Arranges some log files with content on the /logs folder for a component
@@ -90,7 +90,7 @@ public class FileSteps {
             fileName = String.format("%s_%d.log", filePrefix, i);
             createFileAndWriteData(logsDirectory, fileName, false);
         }
-        scenarioContext.put(componentName + activeFile, logsDirectory.resolve(fileName).toAbsolutePath().toString());
+        scenarioContext.put(componentName + ACTIVEFILE, logsDirectory.resolve(fileName).toAbsolutePath().toString());
     }
 
     private void createFileAndWriteData(Path tempDirectoryPath, String fileNamePrefix, boolean isTemp)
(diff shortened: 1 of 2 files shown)


uat/src/main/resources/greengrass/features/log-manager-1.feature
```diff
@@ -40,9 +40,9 @@ Feature: Greengrass V2 LogManager
         """
         And I deploy the Greengrass deployment configuration
         Then the Greengrass deployment is COMPLETED on the device after 2 minutes
+        Then I verify the aws.greengrass.LogManager component is RUNNING using the greengrass-cli
         Then I verify that it created a log group for component type GreengrassSystemComponent for component System, with streams within 120 seconds in CloudWatch
         And I verify that it created a log group for component type UserComponent for component UserComponentA, with streams within 120 seconds in CloudWatch
-        And I verify the rotated files are not deleted except for the active log file for component UserComponentA
```
Nelsonochoam, line 45: This step did not make sense and it does not exist on EF

uat/src/main/java/com.aws.greengrass/FileSteps.java
```diff
@@ -39,8 +39,7 @@ public class FileSteps {
     private final Platform platform;
     private final TestContext testContext;
     private final ScenarioContext scenarioContext;
-
-    private final String ACTIVE_FILE = "ActiveFile";
+    private static final String activeFile = "ActiveFile";
```
dgggIoT, line 42:
```suggestion
    private static final String ACTIVE_FILE = "ActiveFile";
```
nit: I agree with the change from `final `to `static final String`. But `ACTIVE_FILE` may be still a good name, since it implies the static final String. By saying that, in the following code, when seeing `this.activeFile`, I thought it was the file object instead of a string.

Nelsonochoam: I changes it to that based on the checktyle rules... Let me try `ACTIVEFILE`

uat/src/main/resources/greengrass/features/log-manager-1.feature
saranyailla: we don't want to accidentally delete the files (bug in the code for any reason) when the delete files on upload is set to true. Are we only writing the steps that exist on EF?

Nelsonochoam: This step was not on EF + this step doesn't make sense in the context of what this test is testing, the log manager should **never** delete the active log file. We could improve this step by adding it and making it correct but I want us to focus first on porting what we have and then improving them

Nelsonochoam: There are other tests that verify files btw

saranyailla (approved):

PR #172, opened 2022-12-07
chore: update readme for release

Nelsonochoam:
**Description of changes:**
Update the README under uats to better describe what they do and how to run them

### DIFF
--- a/uat/README.md
+++ b/uat/README.md
@@ -1,33 +1,48 @@
-##User Acceptance Test
-UATs are defined under uat module. 
-UATs use aws-greengrass-testing-standalone as the test framework to run the tests.(OTF)
-aws-greengrass-testing-standalone is pulled as maven dependency from GG maven repo. 
-You can add/update UATs at uat source. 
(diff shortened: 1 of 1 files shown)


uat/README.md
```diff
@@ -1,33 +1,48 @@
...
 ```
 
 Command arguments:
 
-Dggc.archive - path to the nucleus zip
-Dtest.log.path - path where the test results are being stored
-Dtags - filter tests by a specific cucumber tags.
\ No newline at end of file
+Dggc.archive - path to the nucleus zip that was downloaded
+Dtest.log.path - path where you would like the test results to be stored
```
saranyailla, line 48: -Dtags?

Nelsonochoam: I did't want to put them there. I want to force it to always target LogManager. If we don't specify it it will actually run all the OTF tests + the LogManager tests

saranyailla (approved):

parrottsquawk (approved):

PR #178, opened 2023-01-06
test: add local deployments for custom component

dgggIoT:
**Issue #, if available:**

**Description of changes:**
This only serves for adding the basic flow and structure of local deployment of custom component. The custom component will be in next PR.

**Checklist:**


uat/testing-features/pom.xml
```diff
@@ -150,6 +165,16 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.7.0</version>
+                <configuration>
+                    <source>1.8</source>
+                    <target>1.8</target>
+                    <encoding>UTF-8</encoding>
+                </configuration>
+            </plugin>
```
saranyailla, line 177: You can revert this and also the lines 16 and 17.

uat/testing-features/src/main/java/com/aws/greengrass/LocalDeploymentSteps.java
```diff
@@ -0,0 +1,267 @@
...
+                "--artifactDir " + artifactPath.toString(),
+                "--recipeDir " + recipePath.toString()));
+        if (StringUtils.isNotBlank(componentVersion)) {
+            commandArgs.add("--merge " + componentName + "=" + componentVersion);
+        }
+        String updateConfigArgs = getCliUpdateConfigArgs(componentName, configuration);
+        if (!updateConfigArgs.isEmpty()) {
+            commandArgs.add("--update-config '" + updateConfigArgs + "'");
+        }
+        LOGGER.info("The cli command: {}", commandArgs);
```
saranyailla, line 134: We can remove this log.

uat/testing-features/src/main/java/com/aws/greengrass/LocalDeploymentSteps.java
```diff
@@ -0,0 +1,266 @@
...
+
+    /**
+     * implemented the step of installing a custom component with configuration.
+     *
+     * @param componentName         the name of the custom component
+     * @param configurationTable    the table which describes the configurations
+     * @throws InterruptedException InterruptedException could be throw out during the component deployment
+     * @throws IOException          IOException could be throw out during preparation of the CLI command
+     */
+    @When("I install the component {word} from local store with configuration")
```
MikeDombo, line 92: why is this here?

Local deployments should be implemented in OTF itself, this is not specific to Log Manager. Can we do this better?

saranyailla (dismissed):

uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
```diff
@@ -9,6 +9,18 @@ Feature: Greengrass V2 LogManager
         And 5 temporary rotated log files for component aws.greengrass.Nucleus have been created
         And 5 temporary rotated log files for component UserComponentA have been created
 
+    @local
+    Scenario: LogManager-1-T0: install the local custom component
+        Given I create a Greengrass deployment with components
+            | aws.greengrass.Cli        | LATEST |
```
MikeDombo, line 15: OTF doesn't have the CLI by default?

uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
```diff
@@ -9,6 +9,18 @@ Feature: Greengrass V2 LogManager
...
+        And I deploy the Greengrass deployment configuration
+        Then the Greengrass deployment is COMPLETED on the device after 2 minutes
+        When I install the component logGenerator from local store with configuration
```
MikeDombo, line 18: I don't want to repeat the same problem that we have with our internal tests.

I want this to use MERGE/RESET format directly.


```
"""json
{
"MERGE": {"key": "value", "FileSize": "10MB"}
}
"""
```

uat/testing-features/src/main/java/com/aws/greengrass/LocalDeploymentSteps.java
dgggIoT: OTF has local deployment but it does not take parameters. The LogManager has it here so then it can take some parameters for different scenarios, e.g. logging frequency, file size, rotation speed.

MikeDombo: OTF should have this logic.

dgggIoT: But for different component, it is hard to define the number of parameters. Will it be more component-owned thing, instead of general steps in OTF?  Here is the [link](https://github.com/aws-greengrass/aws-greengrass-testing/blob/1ce7ef461465fb11f701f2e1231caa817afcba04/aws-greengrass-testing-features/aws-greengrass-testing-features-api/src/main/java/com/aws/greengrass/testing/features/DeploymentSteps.java#L148) for what original OTF has:

uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
dgggIoT: OTF does not install CLI automatically.

uat/testing-features/src/main/java/com/aws/greengrass/LocalDeploymentSteps.java
dgggIoT: This will be moved into OTF eventually. Now, we are trying what we want for local deployments and will provide feedback to OTF team.

MikeDombo: We're all the same team. Why do you add the appropriate code into OTF and then consume it here? We don't need to wait for them to do it, you've already done it; just add it into OTF with a pull request.

dgggIoT: The codes have been merged into OTF. The new commit is letting LogManager directly consume methods from OTF.

MikeDombo (approved):

saranyailla (approved):

uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
```diff
@@ -9,6 +9,24 @@ Feature: Greengrass V2 LogManager
...
+        Given I create a Greengrass deployment with components
+            | aws.greengrass.Cli        | LATEST |
+        And I deploy the Greengrass deployment configuration
+        Then I wait 120 seconds
+        Then the Greengrass deployment is COMPLETED on the device after 2 minutes
+        When I install the component logGenerator from local store with configuration
+        """
+        {
+            "MERGE": {
+                "FileSize": "10MB",
```
Nelsonochoam, line 23: Maybe we should split the unit and size so parsing becomes easier. Also speed might not be the appropriate wording in this context (speed = distance traveled / time elapsed) and we are going to use it to how frequent we write a log into a file. Maybe try



```suggestion
                "FileSize": "10",
                "FileSizeUnit": "MB",
                "LogWriteFrequencySeconds": 1
```

uat/testing-features/pom.xml
```diff
@@ -103,6 +103,21 @@
                 <artifactId>maven-resources-plugin</artifactId>
                 <version>2.6</version>
                 <executions>
+                    <execution>
```
Nelsonochoam, line 106: It would be nice if OTF did this for us and we had a specific structure it could look for (just a thought for the future)

uat/custom-components/src/main/resources/recipes/logGenerator.yaml
```diff
@@ -0,0 +1,18 @@
...
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+
+---
+RecipeFormatVersion: '2020-01-25'
+ComponentName: logGenerator
+ComponentVersion: '0.0.0'
+ComponentDescription: Component to generate log files with random log messages.
+ComponentPublisher: AWS
```
Nelsonochoam, line 11:
```suggestion
ComponentConfiguration:
     DefaultConfiguration:
         LogWriteFrequencySeconds: 1
```

uat/custom-components/src/main/resources/recipes/logGenerator.yaml
```diff
@@ -0,0 +1,18 @@
...
+Manifests:
+  - Artifacts:
+      - URI: classpath:/local-store/artifacts/custom-components.jar
+    Lifecycle:
+      Run: >-
+        java -Dlog.level=INFO -DcomponentName="logGenerator" -jar {artifacts:path}/custom-components.jar 
+        "{configuration:/FileSize}" "{configuration:/LogSpeed}"
```
Nelsonochoam, line 18:
```suggestion
        "{configuration:/FileSize}" "configuration:/FileSizeUnit" "{configuration:/LogWriteFrequencySeconds}"
```

uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
dgggIoT: This is just temporary for testing if the current toy component can receive parameters. They should be well-defined once the custom component is developed.

uat/custom-components/src/main/resources/recipes/logGenerator.yaml
dgggIoT: Why do we want to remove the publisher?

uat/custom-components/src/main/resources/recipes/logGenerator.yaml
dgggIoT: This is just temporary for testing if the current toy component can receive parameters. They should be well-defined once the custom component is developed.

uat/custom-components/src/main/resources/recipes/logGenerator.yaml
Nelsonochoam: We don't that was a mistake

uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
Nelsonochoam: Why even add any value then?

dgggIoT: Good suggestion. I will apply them when developing custom component

PR #179, opened 2023-01-24
fix: handle multiple active files scenario

Nelsonochoam:
**Issue #, if available:**

Customers can describe on the component configuration a regex that picks multiple files that are constantly being written to. The current version of the Log Manager works under the assumption that in each configuration group there is at most only one file that is being actively modified and when users specify a regex that targets a group of files that has more than 1 file that has contents written to it causes the Log Manager to upload duplicates due to the fact that it only tracks the start position for a single non rotated file.

**Description of changes:**

This changes allow the Log Manager to track changes of multiple "active" files within a configuration group. The implementation uses an LRU 1) to track up to 5 active files per component configuration group 2) Avoid in a sensible way and OOM scenario.

Avoid customers from getting duplicate data to their AWS cloudwatch account when using configurations with a regex that targets multiple "active" files.

* New tests
* Manual Testing

**Checklist:**


src/main/java/com/aws/greengrass/logmanager/model/ProcessingFileInformation.java
```diff
@@ -0,0 +1,123 @@
...
+import static com.aws.greengrass.logmanager.LogManagerService.PERSISTED_CURRENT_PROCESSING_FILE_START_POSITION;
+
+/**
+ * Stores information about a file that the Log Manager worked on or previously processed, to know what was the last
+ * start position of the contents uploaded, the file hash, name and last modified.
+ */
+@Builder
+@Getter
+@Data
+public class ProcessingFileInformation {
```
Nelsonochoam, line 32: This is just extracting a static class that currently exists within the service

src/main/java/com/aws/greengrass/logmanager/model/ProcessingFileLRU.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+
+/**
+ * LRU cache that holds information about each file that is being processed. As long as a log file is in the file system
+ * and can be found, we can't know for sure if that file won't get new contents written to it in the future.
+ */
+public class ProcessingFileLRU {
```
Nelsonochoam, line 15: Still need to handle locking here. Although we only have 1 thread

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -524,35 +554,36 @@ private void processCloudWatchAttemptLogInformation(Map<String, Set<LogFile>> co
...
+                ProcessingFileInformation.builder()
+                        .fileName(file.getSourcePath()) // @deprecated
+                        .startPosition(bytesUploaded)
+                        .lastModifiedTime(cloudWatchAttemptLogFileInformation.getLastModifiedTime())
+                        .fileHash(fileHash)
+                        .build();
+
+        if (!processingFileInformationNELSON.containsKey(componentName)) {
+            // TODO: need to figure a smarter mechanism to adjust the LRU capacity. Add an adjust capacity method
+            processingFileInformationNELSON.put(componentName, new ProcessingFileLRU(
```
MikeDombo, line 581: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

Your code calls `ConcurrentHashMap.containsKey`. If `containsKey` returns `false`, your code then calls `ConcurrentHashMap.put`. Because these operations are not atomic, the key could already exist in the map when you call `put()`, even if `containsKey` returns `false`. This could lead to data loss. Instead of `put()`, call `putIfAbsent()`, which associates an unassociated key with a value atomically. See the following diffs on GitHub that make this code change: [diff1](https://github.com/ninjaframework/ninja/commit/921cbd8b05a4ea930a5a5ad7e451a53e15ec03b7#diff-f39d2fcbd1ca7bbafe398d2a9aaeaf5edd8a23661c40b2f82351ff79ac1b23aeL121), [diff2](https://github.com/kieker-monitoring/kieker/commit/43c9ebd98a4597ea71021c96d04e125ee589d7f8#diff-053788507d3e6fdb69e8ca94e5b2bd0a86b03e3cd2d53a452e18dda69ea071d0L152)

Learn more: [ConcurrentHashMap.putIfAbsent in the Java Platform, Standard Edition 8 API Specification](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#putIfAbsent-K-V-)

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -524,35 +553,36 @@ private void processCloudWatchAttemptLogInformation(Map<String, Set<LogFile>> co
...
-                            .build();
-            currentProcessingLogFilePerComponent.put(attemptLogInformation.getComponentName(),
-                    processingFileInformation);
         }
+
+
+        // Track each file that got uploaded. So that if modified in the future we can still track upload
+        // the new contents starting from the last upload position.
+
+        CurrentProcessingFileInformation processingFileInformation =
```
Nelsonochoam, line 570: We are now storing / tracking every single file per component type. Each component we track has an LRU whose capacity defaults to the max of 5 or the number of files that it processed. What this means is that we could track by default up 5 "active files per component/configuration". We can get smarter and dynamically resize the lru

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -410,18 +419,34 @@ private void setDeleteLogFileAfterCloudUpload(String deleteLogFileAfterCloudUplo
...
         if (!currentProcessingComponentTopics.isEmpty()) {
-            CurrentProcessingFileInformation currentProcessingFileInformation =
-                    CurrentProcessingFileInformation.builder().build();
-            currentProcessingComponentTopics.iterator().forEachRemaining(node ->
-                    currentProcessingFileInformation.updateFromTopic((Topic) node));
-            // Only store the processing information when the fileHash is not empty or null.
-            if (Utils.isNotEmpty(currentProcessingFileInformation.getFileHash())) {
-                componentCurrentProcessingLogFile.put(componentName, currentProcessingFileInformation);
-            }
+//            ProcessingFileLRU lru =  new ProcessingFileLRU(DEFAULT_MAX_FILES_TO_TRACK_PER_COMPONENT);
```
Nelsonochoam, line 424: Need to test and add this back to load the values from the config

src/test/java/com/aws/greengrass/logmanager/model/ProcessingFileLRUTest.java
```diff
@@ -0,0 +1,71 @@
...
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@ExtendWith({MockitoExtension.class, GGExtension.class})
```
Nelsonochoam, line 23: I need to add a few more tests here and also change a few now that we are tracking multiple files

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -524,35 +553,36 @@ private void processCloudWatchAttemptLogInformation(Map<String, Set<LogFile>> co
...
+                        .fileName(file.getSourcePath()) // @deprecated
+                        .startPosition(bytesUploaded)
+                        .lastModifiedTime(cloudWatchAttemptLogFileInformation.getLastModifiedTime())
+                        .fileHash(fileHash)
+                        .build();
+
+        if (!processingFilesInformation.containsKey(componentName)) {
+            // TODO: need to figure a smarter mechanism to adjust the LRU capacity. Add an adjust capacity method
+            processingFilesInformation.put(componentName, new ProcessingFileLRU(
+                    Math.max(logFileGroup.getLogFiles().size(), DEFAULT_MAX_FILES_TO_TRACK_PER_COMPONENT)));
```
jbutler, line 581: This only gives you the log files that have been modified since the last iteration, right? I think this would potentially remove hashes prematurely

src/main/java/com/aws/greengrass/logmanager/model/ProcessingFileLRU.java
jbutler: Did you look at using the built-in LinkedHashMap?

Here's an example of how we used to use it in CDA: https://github.com/aws-greengrass/aws-greengrass-client-device-auth/commit/982fe21e7b8171bd01f80e92458600482769c983#diff-dbfa22ec04fcd9b53a287244e4b4ce88dbd005eafefe4f3ba69e371812d3b552R28

Nelsonochoam: Nope ahaha I will update to use that

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
Nelsonochoam: This is just initializing the LRU, in case there is nothing in there. The only time at which they would get removed is on line 585 below if we have more than 5 files. Shouldn't be a problem unless there are more than 5 active files that match the regex

src/main/java/com/aws/greengrass/logmanager/model/ProcessingFileLRU.java
```diff
@@ -0,0 +1,73 @@
...
+        this.capacity = capacity;
+    }
+
+    @Override
+    protected boolean removeEldestEntry(Map.Entry eldest) {
+        return size() > capacity;
+    }
+
+    /**
+     * Converts the objects stored in the LRY into a map. Used serialize the LRU to be stored
```
dgggIoT, line 30: > Used serialize the LRU to be stored
Used to serialize the LRU to be stored

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -108,14 +114,18 @@ public class LogManagerService extends PluginService {
...
+    public static final int DEFAULT_MAX_FILES_TO_TRACK_PER_COMPONENT = 5;
     private final Object spaceManagementLock = new Object();
     private final List<Consumer<EventType>> serviceStatusListeners = new ArrayList<>();
 
     // public only for integ tests
     public final Map<String, Instant> lastComponentUploadedLogFileInstantMap =
             Collections.synchronizedMap(new LinkedHashMap<>());
-    final Map<String, CurrentProcessingFileInformation> componentCurrentProcessingLogFile =
+
+    public final Map<String, ProcessingFileLRU> processingFilesInformation =
```
dgggIoT, line 125: nit: processingFilesInformation -> processingFileLRUs. To do that just for differing from the following name [here](https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/cea1473c4dcc232447f1665f358dc4fd6f289de0/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java#L431)

Nelsonochoam: This is called processingFilesInformation (plural)

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -857,8 +913,12 @@ static class CurrentProcessingFileInformation {
         private long lastModifiedTime;
         @JsonProperty(PERSISTED_CURRENT_PROCESSING_FILE_HASH)
         private String fileHash;
+        private boolean isDirty;
```
dgggIoT, line 916: Is this value used somewhere?

dgggIoT (dismissed):

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -408,20 +418,53 @@ private void setDeleteLogFileAfterCloudUpload(String deleteLogFileAfterCloudUplo
     }
 
     private void loadStateFromConfiguration(String componentName) {
-        Topics currentProcessingComponentTopics = getRuntimeConfig()
+        ProcessingFileLRU lru =  new ProcessingFileLRU(DEFAULT_MAX_FILES_TO_TRACK_PER_COMPONENT);
+
+        // Versions 2.3.0 and below
```
jbutler, line 423: nit: put these into separate methods and try to get the V2 config first. Only fall back to reading V1 config if that doesn't exist

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -447,37 +490,53 @@ private void loadStateFromConfiguration(String componentName) {
...
-        currentProcessingLogFilePerComponent.forEach(componentCurrentProcessingLogFile::put);
+
+
+        // Update the runtime configuration and store the last processed file information
 
         context.runOnPublishQueueAndWait(() -> {
-            componentCurrentProcessingLogFile.forEach((componentName, currentProcessingFileInformation) -> {
-                Topics componentTopics =
+            processingFilesInformation.forEach((componentName, lru) -> {
+                // Update old config value to handle downgrade from v 2.3.1 to older ones
```
jbutler, line 523: I understand why you're doing this, but boy would it be nice not to.

Was the config change from 2.2 to 2.3 backward compatible?

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -108,14 +114,18 @@ public class LogManagerService extends PluginService {
     public static final String UPLOAD_TO_CW_CONFIG_TOPIC_NAME = "uploadToCloudWatch";
     public static final String MULTILINE_PATTERN_CONFIG_TOPIC_NAME = "multiLineStartPattern";
     public static final int DEFAULT_PERIODIC_UPDATE_INTERVAL_SEC = 300;
+    public static final int DEFAULT_MAX_FILES_TO_TRACK_PER_COMPONENT = 5;
```
jbutler, line 117: This is used in a max function below, which makes it the default MIN, not MAX.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
jbutler: 5 is very small if we're not dynamically resizing. I think that needs to be a P0 requirement.

Consider the case where the customer has one regex for all of their components, assuming the follow our recommended naming convention (`com.company.XYZ`)

src/main/java/com/aws/greengrass/logmanager/model/ProcessingFileLRU.java
```diff
@@ -0,0 +1,73 @@
...
+            map.put(key, value.convertToMapOfObjects());
+        });
+
+        return map;
+    }
+
+    /**
+     * Returns the head of the LRU.
+     */
+    public LogManagerService.CurrentProcessingFileInformation head() {
```
jbutler, line 46: I'm not sure if this is supposed to return the oldest or newest entry, but it's returning the tail, not the head :) Please rename to something along the lines of `getOldest`/`getNewest`

src/main/java/com/aws/greengrass/logmanager/model/ProcessingFileLRU.java
```diff
@@ -0,0 +1,73 @@
...
+        if (entry != null) {
+            return  entry.getValue();
+        }
+
+        return null;
+    }
+
+    @Override
+    @SuppressWarnings("PMD.CloneThrowsCloneNotSupportedException")
+    public ProcessingFileLRU clone() {
```
jbutler, line 63: I don't see this used anywhere

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
Nelsonochoam: We made it so that it was... we are still writing the file names into the config. I agree that this is just nasty but I don't think we have communicated to customers that downgrading from a higher to a lower version might result in issues (if we were to not handle it)

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
Nelsonochoam: I guess so, let me try to add something here to handle this with a few tests

src/main/java/com/aws/greengrass/logmanager/model/ProcessingFileLRU.java
Nelsonochoam: getMostRecentlyUsed?

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
jbutler: What happens if one of the active logs doesn't get written to between LM iterations? It won't get returned in the log group, which could result in the hash being removed. I think this approach still has some fundamental issues

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
Nelsonochoam: Done it is on a separate method and processed after v2 methods

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
Nelsonochoam: Added a method below called adjust capacity which will increase the LR capacity such that it always stores the max amount of logFileGroup it has recorded. I don't have enough time but we can modify the adjust capacity method later to clear entries if the capacity number is lower after and a record has not been updated in say 2 days to clear some memory

src/main/java/com/aws/greengrass/logmanager/model/ProcessingFileLRU.java
Nelsonochoam: Just added it because the code scanner complained if I did not overwrite it

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
Nelsonochoam: PS: I know that it won't be perfect unless we store every single file in existence for example if one cycle we get 6 files and then the next cycle we get other 6 completely different files the sure, we will have duplicates.

We could make it so it just keep adding elements forever and when a new element get added it evicts anything that has not been updated in say 2 days as well, but the time to eviction should be something the user should be able to configure

Nelsonochoam: I made some changes to the LRU to work based on time. File information that gets stored that has not been modified in the last MAX_CACHE_INACTIVE_TIME_SECONDS gets removed when a new record gets added. We could make it configurable later on but we could start with a sensible MAX_CACHE_INACTIVE_TIME_SECONDS number.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -447,37 +502,53 @@ private void loadStateFromConfiguration(String componentName) {
...
         });
-        currentProcessingLogFilePerComponent.forEach(componentCurrentProcessingLogFile::put);
+
+
+        // Update the runtime configuration and store the last processed file information
 
         context.runOnPublishQueueAndWait(() -> {
-            componentCurrentProcessingLogFile.forEach((componentName, currentProcessingFileInformation) -> {
-                Topics componentTopics =
+            processingFilesInformation.forEach((componentName, lru) -> {
```
dgggIoT, line 534: There is no `lru` anymore, right?

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -447,37 +502,53 @@ private void loadStateFromConfiguration(String componentName) {
...
+                // Update old config value to handle downgrade from v 2.3.1 to older ones
+                Topics componentTopicsDeprecated =
                         getRuntimeConfig().lookupTopics(PERSISTED_COMPONENT_CURRENT_PROCESSING_FILE_INFORMATION,
                                 componentName);
-                componentTopics.updateFromMap(currentProcessingFileInformation.convertToMapOfObjects(),
+
+                if (lru.getMostRecentlyUsed() != null) {
```
dgggIoT, line 540: Same, no `lru` anymore.

src/test/java/com/aws/greengrass/logmanager/model/ProcessingFilesTest.java
```diff
@@ -0,0 +1,130 @@
...
+        Clock spyClock = spy(Clock.class);
+        MockedStatic<Clock> clockMock;
+        clockMock = mockStatic(Clock.class);
+        clockMock.when(Clock::systemUTC).thenReturn(spyClock);
+        when(spyClock.instant()).thenReturn(Instant.ofEpochMilli(expected));
+        this.clockMock = Optional.of(clockMock);
+    }
+
+    @Test
+    void GIVEN_filledLRU_WHEN_toMap_returnsMapRepresentation() {
```
dgggIoT, line 51: No `lru` now.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
Nelsonochoam: Chatted with Joe about this offline

dgggIoT (approved):

saranyailla (approved):

PR #185, opened 2023-02-08
test: cover scenario where diskSpaceLimit config is removed

Nelsonochoam:
**Issue #, if available:**
There was an issue where someone removed the diskSpaceLimit configuration from their configuration and the Log Manager was still deleting files. I wrote an integration test to validate if there was a real issue (which is all this PR has, a new integ test)

After some research and noticing it was not an issue with the log manager code (hence the integ test). I pinpointed the root cause to it being that removing a configuration value using merge from the AWS console doesn't cause the configuration from being removed in the component. Customers must use reset when wanting to delete values from their configuration

**Description of changes:**

Just an integration test

Covers a scenario not previously tested

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/SpaceManagementTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/SpaceManagementTest.java
@@ -92,8 +92,8 @@ class SpaceManagementTest extends BaseITCase {
         kernel.shutdown();
     }
 
-    private int writeNLogFiles(Path path, int numberOfFiles, String pattern) throws IOException {
-        int totalLength = 0;
+    private long writeNLogFiles(Path path, int numberOfFiles, String pattern) throws IOException {
+        long totalLength = 0;
 
         for (int i = 0; i < numberOfFiles; i++) {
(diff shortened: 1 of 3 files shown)


avipinku (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/SpaceManagementTest.java
```diff
@@ -75,13 +85,51 @@ void beforeEach(ExtensionContext context) {
...
+    }
+
+    private void assertLogFilesAreNotDeleted(ComponentLogConfiguration logConfiguration) throws
+            InvalidLogGroupException, InterruptedException {
+        long originalBytes = getTotalLogFilesBytesFor(logConfiguration);
+        Instant deadline = Instant.now().plusSeconds(20);
+
+        while (Instant.now().isBefore(deadline)) {
+            long currentBytes = getTotalLogFilesBytesFor(logConfiguration);
+            System.out.println(currentBytes);
```
saranyailla, line 129: Nit: Remove sout.

avipinku (approved):

saranyailla (approved):

PR #190, opened 2023-03-07
chore: run lm uats on codebuild pipeline

Nelsonochoam:
**Description of changes:**
Adds workflow to run the uats on codebuild

Extra testing layer to perform e2e tests before merging

Codebuild flow passing

### DIFF
--- a/.github/workflows/uat.yaml
+++ b/.github/workflows/uat.yaml
@@ -1,0 +1,36 @@
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+name: OTF UATS
+
+on:
+  pull_request:
+    branches: 'main'
+
+env:
+  AWS_REGION : "us-west-2"
+  CODE_BUILD_PROJECT_LINUX: "LogManagerUatCodeBuildLinux"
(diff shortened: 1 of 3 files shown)


uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
```diff
@@ -9,25 +9,6 @@ Feature: Greengrass V2 LogManager
         And 5 temporary rotated log files for component aws.greengrass.Nucleus have been created
         And 5 temporary rotated log files for component UserComponentA have been created
 
-    @local
-    Scenario: LogManager-1-T0: install the local custom component
```
Nelsonochoam, line 13: Deleted this test because it is not actually testing an acceptance test. It was just running and testing the test component that will be used for testing the log manager

.github/workflows/uat.yaml
```diff
@@ -0,0 +1,36 @@
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+name: OTF UATS
+
+on:
+  pull_request:
+    branches: 'main'
```
Nelsonochoam, line 9: Running only on PRs to main for now. We can decide if we want to change this later

dgggIoT (approved):

PR #194, opened 2023-03-10
test: check cw logs where applicable

Nelsonochoam:
**Description of changes:**
Modify most UATs to verify the contents of the log files that were uploaded by using the log generator component

Ensure that making code changes do not break the Log Manager functionality

UATs

### DIFF
--- a/uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
+++ b/uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
@@ -19,7 +19,7 @@ Feature: Greengrass V2 LogManager
                 "logsUploaderConfiguration": {
                      "componentLogsConfigurationMap": {
                         "UserComponentM": {
-                            "logFileRegex": "UserComponentM\\w*.log",
+                            "logFileRegex": "UserComponentM(.*).log",
                             "logFileDirectoryPath": "${UserComponentMLogDirectory}",
                             "deleteLogFileAfterCloudUpload": "false"
                         }
@@ -67,7 +67,7 @@ Feature: Greengrass V2 LogManager
                 "logsUploaderConfiguration": {
                      "componentLogsConfigurationMap": {
                         "UserComponentX": {
-                            "logFileRegex": "UserComponentX\\w*.log",
+                            "logFileRegex": "UserComponentX(.*).log",
                             "logFileDirectoryPath": "${UserComponentXLogDirectory}",
                             "deleteLogFileAfterCloudUpload": "true"
                         }
@@ -150,7 +150,7 @@ Feature: Greengrass V2 LogManager
                 "logsUploaderConfiguration": {
                      "componentLogsConfigurationMap": {
                         "UserComponentY": {
-                            "logFileRegex": "UserComponentY\\w*.log",
+                            "logFileRegex": "UserComponentY(.*).log",
                             "logFileDirectoryPath": "${UserComponentYLogDirectory}"
                         }
                     },
--- a/uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
+++ b/uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
@@ -7,8 +7,9 @@ Feature: Greengrass V2 LogManager
         Given my device is registered as a Thing
         And my device is running Greengrass
 
-    Scenario: LogManager-1-T1: component logs are uploaded to cloudwatch
-        Given I create a log directory for component UserComponentM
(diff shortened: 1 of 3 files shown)


uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
```diff
@@ -56,80 +55,51 @@ Feature: Greengrass V2 LogManager
         And I verify that it created a log group of type UserComponent for component UserComponentM, with streams within 60 seconds in CloudWatch
         And I verify 20 logs for UserComponentM of type UserComponent have been uploaded to Cloudwatch within 60 seconds
 
-    Scenario: LogManager-1-T1-b: As a customer I can configure the logs uploader component using a componentLogsConfigurationMap and logs are uploaded to CloudWatch
+    Scenario: LogManager-1-T2: Files are deleted after successfully being uploaded to cloudwatch
+        Given I create a log directory for component UserComponentX
         Given I create a Greengrass deployment with components
             | aws.greengrass.Cli        | LATEST                                    |
             | aws.greengrass.LogManager | classpath:/greengrass/recipes/recipe.yaml |
-        And 5 temporary rotated log files for component aws.greengrass.Nucleus have been created
```
Nelsonochoam, line 63: Deleted this test because it was the exact same one as the one above with the exception of testing an old configuration key that was used before

dgggIoT (dismissed):

uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
saranyailla: I don't think we should delete this test unless we've at least one test already using these keys.

uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
```diff
@@ -1,15 +1,14 @@
...
 
     Background:
         Given my device is registered as a Thing
         And my device is running Greengrass
 
-    Scenario: LogManager-1-T1: configure the log manager component using a componentLogsConfiguration list and logs are uploaded to
-    CloudWatch
-        Given I create a log directory for component UserComponentM 
+    Scenario: LogManager-1-T1: component logs are uploaded to cloudwatch
+        Given I create a log directory for component UserComponentM
```
saranyailla, line 11: Nit: Should we specify the log directory name here (UserComponentMLogDirectory)?

uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
Nelsonochoam: I disagree. Not everything should be tested with UATs. We already have integration and unit tests covering this. It is just redundant and not providing any additional level of confidence that a code change would break the functionality

uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
Nelsonochoam: The log directory name is randomly created to avoid any sort of collision with another test wanting to use the same name

uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
saranyailla: We deprecated `componentLogsConfiguration` for map. There should be a test with `componentLogsConfiguration` as there can be a customer out there using that key.

Nelsonochoam: Yeah. Still, look at the test, that is not what this one is testing. We could ad a new one testing that but this one isn't testing that it is just duplicating the same above. We have integration tests that cover working with that old configuration

Nelsonochoam: I can create a follow up one with just that configuration

saranyailla: My comment may not be relevant for this change in the PR. Just calling that out in general.

uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
saranyailla: It can still be random. I'm saying more in terms of interpolation <log-directory>. 

Because, down it the log manager component configuration, test authors must know that the log directory created i the test will be component_name + "LogDirectory" in the scenario context. It's an improvement not a blocker for this PR.

saranyailla (approved):

dgggIoT (approved):

PR #198, opened 2023-04-14
fix: free up cache space when files get deleted

Nelsonochoam:
**Issue #, if available:**
To fix an issue where customer wrote to multiple active files within a same group we started tracking all the files on a cache that is written into disk. Items get evicted based on time, if the entry has not been modified in more that one day we evict it.

We are not removing entries from the cache once files get deleted though. This PR makes it so that entries for files that have been deleted get removed from the cache, resulting in less memory and disk space consumption.

Avoid memory consumption spikes and potential OOM errors for high write frequency use cases.

Integration tests. Will follow up with a UAT

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerConfigTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerConfigTest.java
@@ -76,7 +76,7 @@ class LogManagerConfigTest extends BaseITCase {
     private static LogManagerService logManagerService;
     private static Path tempDirectoryPath;
     private static Path tempRootDir;
-    private static final  ObjectMapper YAML_OBJECT_MAPPER = new ObjectMapper(new YAMLFactory());
+    private static final ObjectMapper YAML_OBJECT_MAPPER = new ObjectMapper(new YAMLFactory());
     private static final String componentName = "UserComponentA";
     private static final Level minimumLogLevelDefault = Level.INFO;
 
@@ -104,7 +104,8 @@ class LogManagerConfigTest extends BaseITCase {
                 logManagerService = (LogManagerService) service;
             }
         });
-        deviceConfiguration = new DeviceConfiguration(kernel, "ThingName", "xxxxxx-ats.iot.us-east-1.amazonaws.com", "xxxxxx.credentials.iot.us-east-1.amazonaws.com", "privKeyFilePath",
+        deviceConfiguration = new DeviceConfiguration(kernel, "ThingName", "xxxxxx-ats.iot.us-east-1.amazonaws.com",
+                "xxxxxx.credentials.iot.us-east-1.amazonaws.com", "privKeyFilePath",
                 "certFilePath", "caFilePath", "us-east-1", "roleAliasName");
 
         kernel.getContext().put(DeviceConfiguration.class, deviceConfiguration);
@@ -134,16 +135,16 @@ class LogManagerConfigTest extends BaseITCase {
             Exception {
         setupKernel();
 
-        assertThat(()-> logManagerService.getPeriodicUpdateIntervalSec(), eventuallyEval(is(60),
+        assertThat(() -> logManagerService.getPeriodicUpdateIntervalSec(), eventuallyEval(is(60),
                 Duration.ofSeconds(30)));
 
-            logManagerService.getConfig().find(CONFIGURATION_CONFIG_KEY, LOGS_UPLOADER_PERIODIC_UPDATE_INTERVAL_SEC).remove();
-            assertThat(()-> logManagerService.getPeriodicUpdateIntervalSec(),
-                    eventuallyEval(is(LogManagerService.DEFAULT_PERIODIC_UPDATE_INTERVAL_SEC), Duration.ofSeconds(30)));
+        logManagerService.getConfig().find(CONFIGURATION_CONFIG_KEY, LOGS_UPLOADER_PERIODIC_UPDATE_INTERVAL_SEC).remove();
+        assertThat(() -> logManagerService.getPeriodicUpdateIntervalSec(),
+                eventuallyEval(is(LogManagerService.DEFAULT_PERIODIC_UPDATE_INTERVAL_SEC), Duration.ofSeconds(30)));
 
-            logManagerService.getConfig().lookup(CONFIGURATION_CONFIG_KEY, LOGS_UPLOADER_PERIODIC_UPDATE_INTERVAL_SEC).withValue(600);
-            assertThat(()-> logManagerService.getPeriodicUpdateIntervalSec(), eventuallyEval(is(600),
-                    Duration.ofSeconds(30)));
+        logManagerService.getConfig().lookup(CONFIGURATION_CONFIG_KEY, LOGS_UPLOADER_PERIODIC_UPDATE_INTERVAL_SEC).withValue(600);
+        assertThat(() -> logManagerService.getPeriodicUpdateIntervalSec(), eventuallyEval(is(600),
+                Duration.ofSeconds(30)));
     }
 
     @Test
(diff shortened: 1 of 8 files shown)


MikeDombo (commented):
What if the file is deleted because the logger (the writer of the file) deleted it due to max size/max file count?

This seems to only handle when _we_ delete the files right?

MikeDombo (commented):
We have some piece of code in here which is finding log files to upload (ie. we are listing files), can we use that list of files to remove from the cache when the files no longer exist?

src/main/java/com/aws/greengrass/logmanager/model/ProcessingFiles.java
```diff
@@ -45,6 +48,28 @@ public void put(LogManagerService.CurrentProcessingFileInformation value) {
         evictStaleEntries();
     }
 
+    /**
+     * Removes an entry from the cache for a provided file hash.
+     * @param fileHash - A file hash.
+     */
+    public void remove(String fileHash) {
+        Node node = cache.remove(fileHash);
```
MikeDombo, line 56: handle null filehash?

Nelsonochoam:
> What if the file is deleted because the logger (the writer of the file) deleted it due to max size/max file count?
> 
> This seems to only handle when _we_ delete the files right?

Even if they delete it (not us) as long as we had processed it we would clear it from the cache

src/main/java/com/aws/greengrass/logmanager/model/ProcessingFiles.java
Nelsonochoam: Yeah we should. Theoretically that would not happen because we filter and not process empty file hashes but this calls should defensively handle it.

MikeDombo:
> Even if they delete it (not us) as long as we had processed it we would clear it from the cache

Processed in what way? uploaded? What if we never uploaded it, but it was added to our cache?

Nelsonochoam:
> > Even if they delete it (not us) as long as we had processed it we would clear it from the cache
> 
> 
> 
> Processed in what way? uploaded? What if we never uploaded it, but it was added to our cache?

We only add files to the cache only after we have uploaded contents from it. If we have fully uploaded a file, the pointer we track will be equal to the file size at that point we check if the delete after upload is enabled we try to delete the file, at which point we remove the file from the cache regardless of the LM succeeding to remove the file from the FS (due to an external process having deleted it already for example)

MikeDombo (approved):

MikeDombo:
Basically I'm happy with this change as-is, but I think we need to go further to make sure our runtime config doesn't keep around useless stuff.

src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerConfigTest.java
```diff
@@ -463,12 +464,15 @@ void GIVEN_runtimeConfiguration_withOldProcessingFileFormat_WHEN_onStartUp_THEN_
                 PERSISTED_COMPONENT_CURRENT_PROCESSING_FILE_INFORMATION,
                 componentName);
 
+        long now = Instant.now().toEpochMilli();
```
Nelsonochoam, line 467: This is the only thing that changed on this file the rest are just auto formatting.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -237,6 +240,15 @@ private synchronized void processConfiguration(Map<String, Object> logsUploaderC
         });
 
         this.componentLogConfigurations = newComponentLogConfigurations;
+
+        // Clear cached entries for components that have been removed
+
+        processingFilesInformation.keySet().forEach(componentName -> {
```
Nelsonochoam, line 246: Clear the cache entries for a component if it is removed from the configuration

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -489,8 +501,13 @@ private void handleCloudWatchAttemptStatus(CloudWatchAttempt cloudWatchAttempt)
...
             }
 
+            ProcessingFiles processingFiles = processingFilesInformation.get(componentName);
+
             // Delete files based on diskspace
-            this.diskSpaceManagementService.freeDiskSpace(attemptLogInformation.getLogFileGroup());
+            List<String> deletedHashes =
+                    this.diskSpaceManagementService.freeDiskSpace(attemptLogInformation.getLogFileGroup());
+
+            processingFiles.remove(deletedHashes); // Stop tracking files already uploaded
```
Nelsonochoam, line 510: Remove entries from the cache for files that have been removed due to disk space being freed

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -543,6 +583,13 @@ private void deleteFile(ComponentLogConfiguration config, LogFile file) {
         } else {
             logger.atWarn().log("Unable to delete file with name {}", file.getName());
         }
+
+        // Stop tracking file currently processing information once deleted
+        ProcessingFiles processingFiles = processingFilesInformation.get(config.getName());
```
Nelsonochoam, line 588: Remove entries from cache when deleted after upload

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -768,6 +817,8 @@ public Map<String, Object> convertToMapOfObjects() {
             currentProcessingFileInformationMap.put(PERSISTED_CURRENT_PROCESSING_FILE_START_POSITION, startPosition);
             currentProcessingFileInformationMap.put(PERSISTED_CURRENT_PROCESSING_FILE_LAST_MODIFIED_TIME,
                     lastModifiedTime);
+            currentProcessingFileInformationMap.put(PERSISTED_CURRENT_PROCESSING_FILE_LAST_ACCESSED_TIME,
```
Nelsonochoam, line 820: Store the lastAccessed time here in case the customer restarts greengrass or the LM component

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -753,6 +800,8 @@ public static class CurrentProcessingFileInformation {
         private long startPosition;
         @JsonProperty(PERSISTED_CURRENT_PROCESSING_FILE_LAST_MODIFIED_TIME)
         private long lastModifiedTime;
+        @JsonProperty(PERSISTED_CURRENT_PROCESSING_FILE_LAST_ACCESSED_TIME)
+        private long lastAccessedTime = Instant.EPOCH.toEpochMilli();
```
MikeDombo, line 804: default should be now right? customers upgrading to this version won't have this key, so the default should be something more conservative

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -753,6 +800,8 @@ public static class CurrentProcessingFileInformation {
         private long startPosition;
         @JsonProperty(PERSISTED_CURRENT_PROCESSING_FILE_LAST_MODIFIED_TIME)
         private long lastModifiedTime;
+        @JsonProperty(PERSISTED_CURRENT_PROCESSING_FILE_LAST_ACCESSED_TIME)
```
MikeDombo, line 803: are we even using Jackson on this object?

If we are, make sure that the jackson we use is configured to allow for unknown properties otherwise this will break if someone downgrades.

src/main/java/com/aws/greengrass/logmanager/model/ProcessingFiles.java
```diff
@@ -35,25 +39,72 @@ public ProcessingFiles(int maxInactiveTimeSeconds) {
         this.maxInactiveTimeSeconds = maxInactiveTimeSeconds;
     }
 
+    /**
+     * Stores the currently processing file information only if the entry does not exist.
+     *
+     * @param value - currently processing file information
+     */
+    public void putIfAbsent(LogManagerService.CurrentProcessingFileInformation value) {
+        if (cache.get(value.getFileHash()) == null) {
```
MikeDombo, line 48: use computeIfAbsent

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
Nelsonochoam: Great point I wasn't too familiar with Jackson, but it seems this might be leftovers from some legacy code. Might be able to remove them

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
Nelsonochoam: I was setting the value on the ProcessingFiles but doing it here is actually simpler. Will take that suggestion.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -397,7 +408,7 @@ private void loadProcessingFilesConfigDeprecated(String componentName) {
             currentProcessingComponentTopicsDeprecated.iterator().forEachRemaining(node ->
                     processingFileInformation.updateFromTopic((Topic) node));
 
-            processingFiles.put(processingFileInformation);
+            processingFiles.putIfAbsent(processingFileInformation);
```
saranyailla, line 411: I think we should update the last accessed time so we should still use `put` instead of `putIfAbsent`?

Nelsonochoam: In this case I don't want it here to be put again if it is already there. This method gets called when the configuration gets loaded. It will only be empty when we are booting up GG or we get restarted but for the rest we just want to put the file info if absent.

Since this values come from the config we don't want to update the lastAccessed time here.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -89,6 +88,8 @@ public class LogManagerService extends PluginService {
             "lastFileProcessedTimeStamp";
     public static final String PERSISTED_CURRENT_PROCESSING_FILE_LAST_MODIFIED_TIME =
             "currentProcessingFileLastModified";
+    public static final String PERSISTED_CURRENT_PROCESSING_FILE_LAST_ACCESSED_TIME =
+            "currentProcessingFileLastAccessed";
```
MikeDombo, line 92: no reason for this key to be so long, everything takes us space. Just `lastAccessed` is fine.

src/main/java/com/aws/greengrass/logmanager/model/ProcessingFiles.java
MikeDombo: that's because you're calling put, don't call put from here.

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/logmanager/model/ProcessingFiles.java
Nelsonochoam: Yeah I know, but on put I am also setting the mostRecentlyUsed which is something that I would still want. Instead of doing it in to places I just do it in put and have this just delegate.

MikeDombo (approved):

saranyailla (approved):

PR #199, opened 2023-04-18
Release 2.3.x

Nelsonochoam:
**Issue #, if available:**

**Description of changes:**

**Checklist:**


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -622,17 +704,18 @@ private void processLogsAndUpload() throws InterruptedException {
...
 
                         // If the file was partially read in the previous run, then get the starting position for
                         // new log lines.
-                        if (componentCurrentProcessingLogFile.containsKey(componentName)) {
-                            CurrentProcessingFileInformation processingFileInformation =
-                                    componentCurrentProcessingLogFile.get(componentName);
-                            if (processingFileInformation.fileHash.equals(fileHash)) {
-                                startPosition = processingFileInformation.startPosition;
+                        if (processingFilesInformation.containsKey(componentName)) {
+                            ProcessingFiles processingFiles = processingFilesInformation.get(componentName);
```
MikeDombo, line 714: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.



**Problem**  
 You are using a `ConcurrentHashMap`, but your usage of `containsKey()` and `get()` may not be thread-safe at lines: **713 and 714**. In between the check and the `get()` another thread can remove the key and the `get()` will return `null`. The remove that can remove the key is at line: **557**.

**Fix**  
 Consider calling `get()`, checking instead of your current check if the returned object is `null`, and then using that object only, without calling `get()` again.

**More info**  
 [View an example on GitHub](https://github.com/apache/hadoop/blob/f16cf877e565084c66bc63605659b157c4394dc8/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/s3guard/S3Guard.java#L302-L304) (external link).

PR #200, opened 2023-05-25
test: add a UAT to verify max log event limit

saranyailla:
**Issue #, if available:**

**Description of changes:**
When there are more than 10000 logs to upload, then log manager uploads those logs in multiple attempts where each attempt uploads a maximum of 10000 logs. (CW limit for max log events per API is 10000)

**Checklist:**

### DIFF
--- a/uat/custom-components/src/main/java/com/aws/greengrass/artifacts/LogGenerator.java
+++ b/uat/custom-components/src/main/java/com/aws/greengrass/artifacts/LogGenerator.java
@@ -59,8 +59,8 @@ public class LogGenerator implements Consumer<String[]> {
         Logger logger = configureLogger();
 
         for (int i = 1; i <= numberOfLogs; i++) {
-            String logLine = String.format("(seq: %d, name: %s)", i, logFileName);
-            logger.info(logLine);
+            String logLine = String.format("(seq: %d)", i);
+            logger.info(logLine); // INFO LogGenerator (seq: 1)
             TimeUnit.MILLISECONDS.sleep(writeFreqMs);
         }
     }
--- a/uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
+++ b/uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
@@ -38,7 +38,7 @@ Feature: Greengrass V2 LogManager
         }
         """
         And I deploy the Greengrass deployment configuration
-        And the Greengrass deployment is COMPLETED on the device after 2 minutes
+        And the Greengrass deployment is COMPLETED on the device after 3 minutes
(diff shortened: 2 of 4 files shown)


uat/custom-components/src/main/java/com/aws/greengrass/artifacts/LogGenerator.java
```diff
@@ -59,8 +59,8 @@ private void generateLogs() throws InterruptedException {
         Logger logger = configureLogger();
 
         for (int i = 1; i <= numberOfLogs; i++) {
-            String logLine = String.format("(seq: %d, name: %s)", i, logFileName);
-            logger.info(logLine);
+            String logLine = String.format("(seq: %d)", i);
```
saranyailla, line 62: This is to make sure that the log message length is less than 70 bytes (34 more bytes are added per log event) and  so LM will try to upload 10,000+ log events at once (10000*104 = ~1,040,000). Otherwise, max limit is reached (1024*1024 = ~1,048,576) for <10,000 logs and LM will upload log events  less than 10,000.

MikeDombo (dismissed):

jcosentino11 (dismissed):

MikeDombo (approved):

jcosentino11 (approved):

uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
```diff
@@ -228,9 +228,57 @@ Feature: Greengrass V2 LogManager
...
+                    ],
+                    "systemLogsConfiguration": {
+                        "uploadToCloudWatch": "true",
+                        "minimumLogLevel": "INFO",
+                        "diskSpaceLimit": "25",
+                        "diskSpaceLimitUnit": "MB",
+                        "deleteLogFileAfterCloudUpload": "true"
+                    }
+                },
+                "periodicUploadIntervalSec": "300"
```
MikeDombo, line 264: why is this set to 300? Shouldn't you set this very very low so the test can execute more quickly?

Adding this UAT seems to have increased the OTF time from 7 minutes to over 20!

saranyailla: As soon as the custom component writing logs is deployed, LM starts uploading active logs (which will be less than 10K) because it is deployed first. This is to force LM to wait and upload 10K+ logs at once by giving enough time for the custom component to be deployed (~3 min) and write 10K+ logs. 240 seconds may also work.

MikeDombo: why don't you just deploy the custom component first then?

PR #201, opened 2023-05-25
fix: validate cw log events size  before uploading

saranyailla:
**Issue #, if available:**

**Description of changes:**
AWS CloudWatch PutLogEvents API has a limit of 10,000 log events per request.
Documentation link: [Array Members: Minimum number of 1 item. Maximum number of 10000 items](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html#:~:text=array%20members%3A%20minimum%20number%20of%201%20item.%20maximum%20number%20of%2010000%20items.).

Currently, log manager component tries to upload all the log events without validating their size. Adding an extra check to validate the log events size before adding a new log event. 

This is an edge case where there are more than 10,000 logs to upload, with each log event size less than the 1024*256 - 8-26 bytes limit and the total batch size less than 1024*1024 bytes. 

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
+++ b/src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
@@ -67,6 +67,7 @@ public class CloudWatchAttemptLogsProcessor {
     private static final int TIMESTAMP_BYTES = 8;
     private static final int MAX_BATCH_SIZE = 1024 * 1024;
     private static final int MAX_EVENT_LENGTH = 1024 * 256 - TIMESTAMP_BYTES - EVENT_STORAGE_OVERHEAD;
+    private static final int MAX_NUM_OF_LOG_EVENTS = 10_000;
     private static final ObjectMapper DESERIALIZER = new ObjectMapper()
             .configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true);
     private static final ThreadLocal<SimpleDateFormat> DATE_FORMATTER =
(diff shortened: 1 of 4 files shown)


MikeDombo (approved):
Please merge UAT first and have it execute successfully with this PR

src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java
```diff
@@ -379,6 +380,10 @@ private Pair<Boolean, AtomicInteger> addNewLogEvent(AtomicInteger totalBytesRead
             return new Pair<>(true, new AtomicInteger());
         }
 
+        if (attemptLogInformation.getLogEvents().size() >= MAX_NUM_OF_LOG_EVENTS) {
```
jcosentino11, line 383:
```suggestion
        if (attemptLogInformation.getLogEvents().size() > MAX_NUM_OF_LOG_EVENTS) {
```
From linked docs it seems like 10000 is allowed? 
> The maximum number of log events in a batch is 10,000

saranyailla: If `attemptLogInformation.getLogEvents()` is already 10000, it should return that the max limit is reached. In fact, this check need not compare if it is greater than 10,000 as we only add one event at a time after this check. 

https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/fd906cb0c1dc2f23d0a054a98965dee73b2ed710/src/main/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessor.java#L420

jcosentino11: Ah makes sense-- I scanned the code too quickly 😅

jcosentino11 (approved):

uat/testing-features/src/main/resources/greengrass/features/log-manager-1.feature
```diff
@@ -233,7 +233,7 @@ Feature: Greengrass V2 LogManager
         Then I verify that it created a log group of type GreengrassSystemComponent for component System, with streams within 60 seconds in CloudWatch
         Then I verify that it created a log group of type UserComponent for component UserComponentY, with streams within 60 seconds in CloudWatch
         And I verify 20 logs for UserComponentY of type UserComponent have been uploaded to Cloudwatch within 80 seconds
-
+    @saranya
```
MikeDombo, line 236: remove

MikeDombo (approved):

vabhasin (approved):

PR #207, opened 2023-06-17
perf: only write actually changed values to config

MikeDombo:
**Issue #, if available:**

**Description of changes:**
1. Add new config option "deprecatedVersionSupport" which will prevent log manager from loading and writing the deprecated currently processing file information. This helps with performance because we don't need to write the same information to both the V2 section and the deprecated section. The default is to keep supporting the old version, set it to "false" to stop it.
2. Add a new ConfigUtil which allows updating config from a map, but ignoring when nothing actually changes. Currently a large portion of config updates are just "timestampUpdated" which we really don't care about.

Utilizing both of these changes, we're up to ~1.8MBps upload rate.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
+++ b/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
@@ -62,6 +62,7 @@ import javax.inject.Inject;
 
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
 import static com.aws.greengrass.logmanager.LogManagerService.LOGS_UPLOADER_SERVICE_TOPICS;
+import static com.aws.greengrass.logmanager.util.ConfigUtil.updateFromMapWhenChanged;
 
 
 @ImplementsService(name = LOGS_UPLOADER_SERVICE_TOPICS, version = "2.0.0")
@@ -404,7 +405,8 @@ public class LogManagerService extends PluginService {
                 .lookupTopics(PERSISTED_COMPONENT_CURRENT_PROCESSING_FILE_INFORMATION, componentName);
 
 
-        if (currentProcessingComponentTopicsDeprecated != null
+        if (isDeprecatedVersionSupported()
+                && currentProcessingComponentTopicsDeprecated != null
                 && !currentProcessingComponentTopicsDeprecated.isEmpty()) {
             CurrentProcessingFileInformation processingFileInformation =
                     CurrentProcessingFileInformation.builder().build();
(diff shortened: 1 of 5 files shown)


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -520,15 +522,18 @@ private void handleCloudWatchAttemptStatus(CloudWatchAttempt cloudWatchAttempt)
...
-                // Update old config value to handle downgrade from v 2.3.1 to older ones
-                Topics componentTopicsDeprecated =
-                        getRuntimeConfig().lookupTopics(PERSISTED_COMPONENT_CURRENT_PROCESSING_FILE_INFORMATION,
-                                componentName);
-
-                if (processingFiles.getMostRecentlyUsed() != null) {
-                    componentTopicsDeprecated.updateFromMap(
-                            processingFiles.getMostRecentlyUsed().convertToMapOfObjects(),
-                        new UpdateBehaviorTree(UpdateBehaviorTree.UpdateBehavior.REPLACE, System.currentTimeMillis()));
+                if (isDeprecatedVersionSupported()) {
```
parrottsquawk, line 525: Didn't we check this in the parent if statement?

MikeDombo: No, that's a completely different method.

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -404,7 +405,8 @@ private void loadProcessingFilesConfigDeprecated(String componentName) {
                 .lookupTopics(PERSISTED_COMPONENT_CURRENT_PROCESSING_FILE_INFORMATION, componentName);
 
 
-        if (currentProcessingComponentTopicsDeprecated != null
+        if (isDeprecatedVersionSupported()
+                && currentProcessingComponentTopicsDeprecated != null
```
parrottsquawk, line 409: Nit, can we make this more readable with a dedicated variable?
```
bool weShouldUpdateTheSprocketEngine = isDeprecatedVersionSupported() && currentProcessingComponentTopicsDeprecated != null && !currentProcessingComponentTopicsDeprecated.isEmpty();

if (weShouldUpdateTheSprocketEngine) {
...
```

src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
parrottsquawk: it's confusing that they have the same signature then `isDeprecatedVersionSupported()`

robcmann (approved):

Issue #213, opened 2023-10-25
logManager: Error while deploying


JoseOrtega17:
**Describe the bug**
Error while deploying logManager 

**To Reproduce**
Running a new deployment

**Expected behavior**
Deployment success

**Actual behavior**
• DEPLOYMENT_FAILURE
• COMPONENT_UPDATE_ERROR
• COMPONENT_BROKEN

**Environment**
 - OS: Docker container jetson nano Jetpack 32.1
 - JDK version:
```
 openjdk version "11.0.19" 2023-04-18
OpenJDK Runtime Environment (build 11.0.19+7-post-Ubuntu-0ubuntu118.04.1)
OpenJDK 64-Bit Server VM (build 11.0.19+7-post-Ubuntu-0ubuntu118.04.1, mixed mode)
```
 - Nucleus version: 2.11.2
 - Log Manager version: 2.3.5

**Additional context**
Error when checking GG log:
```

2023-10-25T18:42:54.277Z [ERROR] (pool-2-thread-33) com.aws.greengrass.logmanager.LogManagerService: service-errored. {serviceName=aws.greengrass.LogManager, currentState=RUNNING}
java.lang.NullPointerException: Name is null
	at java.base/java.lang.Enum.valueOf(Enum.java:238)
	at org.slf4j.event.Level.valueOf(Level.java:14)
	at com.aws.greengrass.logmanager.CloudWatchAttemptLogsProcessor.checkAndAddNewLogEvent(CloudWatchAttemptLogsProcessor.java:350)
	at com.aws.greengrass.logmanager.CloudWatchAttemptLogsProcessor.processLogLine(CloudWatchAttemptLogsProcessor.java:251)
	at com.aws.greengrass.logmanager.CloudWatchAttemptLogsProcessor.processLogFiles(CloudWatchAttemptLogsProcessor.java:181)
	at com.aws.greengrass.logmanager.LogManagerService.lambda$processLogsAndUpload$22(LogManagerService.java:769)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at com.aws.greengrass.logmanager.LogManagerService.processLogsAndUpload(LogManagerService.java:768)
	at com.aws.greengrass.logmanager.LogManagerService.startup(LogManagerService.java:797)
	at com.aws.greengrass.lifecyclemanager.Lifecycle.lambda$handleStateTransitionStartingToRunningAsync$9(Lifecycle.java:591)
[...]

PR #214, opened 2023-10-25
fix: handle valid JSON messages which are not GreengrassLogMessages

MikeDombo:
**Issue #, if available:**
#213

**Description of changes:**
Handle the case where log level is null in deserialized JSON format logs. Log level is never supposed to be null, so this should only be possible if someone is not using the Greengrass logger (waiting for more customer info to determine this). Regardless, we should handle the case properly. Here we default to INFO level if the log level parsed from JSON is null or empty.

Updated existing test file with a null log level and verified that test fails without the fix and that it passes with the fix.

**Checklist:**

### DIFF
--- a/src/test/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessorTest.java
+++ b/src/test/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessorTest.java
@@ -502,7 +502,6 @@ class CloudWatchAttemptLogsProcessorTest extends GGServiceTestUtil {
         assertTrue(logEventsForStream2.getAttemptLogFileInformationMap().containsKey(fileHash2));
         assertEquals(0, logEventsForStream2.getAttemptLogFileInformationMap().get(fileHash2).getStartPosition());
         assertEquals(1237, logEventsForStream2.getAttemptLogFileInformationMap().get(fileHash2).getBytesRead());
-        System.out.println(logEventsForStream2.getAttemptLogFileInformationMap().get(fileHash2));
         assertEquals("TestComponent", logEventsForStream2.getComponentName());
     }
 
--- a/src/test/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessorTest.java
+++ b/src/test/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessorTest.java
@@ -476,8 +476,8 @@ class CloudWatchAttemptLogsProcessorTest extends GGServiceTestUtil {
         String logStream2 = "/2020/12/18/thing/testThing";
         assertTrue(attempt.getLogStreamsToLogEventsMap().containsKey(logStream));
         assertTrue(attempt.getLogStreamsToLogEventsMap().containsKey(logStream2));
-        CloudWatchAttemptLogInformation logEventsForStream1 = attempt.getLogStreamsToLogEventsMap().get(logStream);
-        CloudWatchAttemptLogInformation logEventsForStream2 = attempt.getLogStreamsToLogEventsMap().get(logStream2);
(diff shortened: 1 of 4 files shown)


uat/custom-components/src/test/java/com/aws/greengrass/LogGeneratorTest.java
```diff
@@ -17,19 +16,18 @@ public class LogGeneratorTest {
...
     String logFileName = "localtest";
-    String logFileExtention = "log";
-    String logWriteFreqSeconds = "0.1";
+    String logWriteFreqMs = "100";
     String totalLogNumbers = "50";
     String fileSizeBytes = "1024";
     String fileSizeUnit = "KB";
     String componentName = "com.aws.greengrass.artifacts.LogGenerator";
-    String activeFileName = logFileName + "." + logFileExtention;
+    String activeFileName = logFileName + ".log";
```
K-AWS, line 24: Why are we moving from a variable to a hard coded string?

K-AWS (approved):
One nit about the ".log" string

uat/custom-components/src/test/java/com/aws/greengrass/LogGeneratorTest.java
MikeDombo: This is a test, variable wasn't a variable but a constant.

src/test/java/com/aws/greengrass/logmanager/CloudWatchAttemptLogsProcessorTest.java
```diff
@@ -494,10 +494,15 @@ void GIVEN_one_components_two_file_less_than_max_WHEN_merge_THEN_reads_and_merge
...
+        assertEquals(3, logEventsForStream2.getLogEvents().size());
+        // Read the 1 remaining cloudwatch log stream which will have today's date because the log lines
+        // had no parsed timestamp.
+        assertEquals(2, attempt.getLogStreamsToLogEventsMap()
+                .get(attempt.getLogStreamsToLogEventsMap().keySet().stream().findFirst().get()).getLogEvents().size());
         assertTrue(logEventsForStream2.getAttemptLogFileInformationMap().containsKey(fileHash2));
         assertEquals(0, logEventsForStream2.getAttemptLogFileInformationMap().get(fileHash2).getStartPosition());
-        assertEquals(1239, logEventsForStream2.getAttemptLogFileInformationMap().get(fileHash2).getBytesRead());
+        assertEquals(1237, logEventsForStream2.getAttemptLogFileInformationMap().get(fileHash2).getBytesRead());
+        System.out.println(logEventsForStream2.getAttemptLogFileInformationMap().get(fileHash2));
```
saranyailla, line 505: Nit: remove this line.

MikeDombo: updated.

saranyailla (approved):

uat/custom-components/src/test/java/com/aws/greengrass/LogGeneratorTest.java
```diff
@@ -17,19 +16,18 @@ public class LogGeneratorTest {
     @TempDir
     static Path tempPath;
     String logFileName = "localtest";
-    String logFileExtention = "log";
-    String logWriteFreqSeconds = "0.1";
+    String logWriteFreqMs = "100";
```
K-AWS, line 19: Is someone going to think this is milliseconds and not messages per second?

K-AWS (approved):

uat/custom-components/src/test/java/com/aws/greengrass/LogGeneratorTest.java
MikeDombo: It is milliseconds. 0.1 seconds is 100 milliseconds.

PR #238, opened 2026-04-09
feat: create Rust package with Cargo workspace and module layout

utsavvyas:
*Issue #, if available:**

N/A — Initial project setup for Rust v2 implementation of LogManager

**Description of changes:**

Sets up the Rust project structure for `gg-log-manager` — a Generic Type component reimplementing `aws.greengrass.LogManager` in Rust for Greengrass Nucleus and Nucleus Lite.

- `Cargo.toml` with dependencies: `tracing` (structured logging), `tracing-subscriber`
- `LICENSE` — Apache-2.0
- Module layout: `config/`, `scanner/`, `uploader/`, `disk/`, `credentials/`
- `main.rs` — imports from lib crate, tracing init
- `recipe.yaml` — GG Generic type, multi-arch (amd64 + arm64, Linux + Windows)
- `.cargo/config.toml` — cross-compilation targets for musl
- Release profile: `strip = true`, `lto = true`, `opt-level = "z"` for minimal binary size

Foundation for the Rust v2 LogManager component. The current Java implementation runs as a Nucleus plugin (~30MB JVM footprint). The Rust implementation targets <2MB statically linked binary, suitable for Nucleus Lite and resource-constrained devices.

- `cargo zigbuild --target aarch64-unknown-linux-musl` ✅
- `cargo zigbuild --target x86_64-unknown-linux-musl` ✅

**Checklist:**


.gitignore
```diff
@@ -1,2 +1,2 @@
-target/
-Cargo.lock
+/target
+tarpaulin-report.html
```
archigup, line 3: What is this for?

.cargo/config.toml
```diff
@@ -0,0 +1,8 @@
+[target.x86_64-unknown-linux-musl]
+rustflags = ["-C", "target-feature=+crt-static"]
```
archigup: Why are we statically linking libc?

archigup: default should be building dynamically, unless explicitly building statically, so it works on musl distros

src/credentials/mod.rs
```diff
@@ -0,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Credential provider and environment detection for GG Classic and Lite
```
archigup:
```suggestion
//! Credential provider and environment detection for Greengrass Nucleus and Nucleus Lite
```

README.md
```diff
@@ -1 +1,44 @@
...
+
+```bash
+# ARM64 Linux (aarch64)
+cargo zigbuild --release --target aarch64-unknown-linux-musl
+
+# x86_64 Linux
+cargo zigbuild --release --target x86_64-unknown-linux-musl
+```
+
+The GG Component SDK requires Linux (`epoll`), so native macOS/Windows builds are not supported. Use cross-compilation.
```
archigup: Doesn't this mean cross-compilation to macOS/Windows wouldn't work either? epoll is a runtime requirement

Cargo.lock
```diff
@@ -0,0 +1,401 @@
...
+checksum = "ba73ea9cf16a25df0c8caa16c51acb937d5712a8429db78a3ee29d5dcacd3a65"
+
+[[package]]
+name = "windows-link"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f0805222e57f7521d6a62e36fa9163bc891acd422f971defe97d64e70d0a4fe5"
+
+[[package]]
+name = "windows-sys"
```
archigup, line 452: Why are windows crates being pulled in?

.gitignore
utsavvyas: It is used for the code coverage report generation.

.cargo/config.toml
utsavvyas: Fair callout. Changed it.

README.md
utsavvyas: Is epoll needed for runtime updates on the configuration right ? If we remove that requirement we dont need it right ?

Cargo.lock
utsavvyas: I believe tokio by default fetch, I have removed it in the initial pull but it will come eventually. Current log manager do support windows.

README.md
archigup: Epoll is used to implement the IPC protocol

utsavvyas: I believe we dont even need IPC protocol as well.

archigup (approved):

Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+[package]
+name = "gg-log-manager"
+version = "0.1.0"
+edition = "2021"
+description = "Greengrass LogManager - Generic type component for log and EMF file upload to CloudWatch"
+license = "Apache-2.0"
+
+[dependencies]
+tracing = "0.1"
+tracing-subscriber = "0.3"
```
alter-mage, line 15: The PR description and commit message mention `aws-greengrass-component-sdk` and `tokio` as dependencies, but I only see `tracing` and `tracing-subscriber` here. Were those removed at some point? Can you update the description and commit message to reflect what's actually in the PR?

src/main.rs
```diff
@@ -0,0 +1,23 @@
...
+    tracing::info!("gg-log-manager starting");
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn all_modules_compile() {
+        assert!(true);
+    }
+}
```
alter-mage: What is this test verifying? `assert!(true)` can never fail, and module compilation is already guaranteed by `cargo build` / `cargo test`. Is there a reason to keep this rather than removing it?

src/main.rs
```diff
@@ -0,0 +1,23 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! gg-log-manager: Greengrass LogManager Generic Type Component.
+
+mod config;
+mod credentials;
+mod disk;
+mod scanner;
+mod uploader;
```
alter-mage: Have you considered importing from the library crate instead of re-declaring the modules here?

Something like:
```rust
use gg_log_manager::{config, credentials, disk, scanner, uploader};
```
Would that work here?

recipe.yaml
```diff
@@ -0,0 +1,41 @@
...
+      install: chmod +x {artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-arm64
+      run: "{artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-arm64"
+      shutdown: SIGTERM
+  - Platform:
+      os: windows
+      architecture: amd64
+    Artifacts:
+      - URI: s3://gg-components/gg-log-manager-{COMPONENT_VERSION}-windows-amd64.exe
+    Lifecycle:
+      run: "{artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-windows-amd64.exe"
```
alter-mage, line 41: Is this intended to be the actual production recipe, or more of a reference/example? And is there a reason the Windows manifest doesn't have `install` and `shutdown` lifecycle steps like the Linux ones do?

Cargo.toml
utsavvyas: Good catch. The PR description was stale from an earlier iteration. aws-greengrass-component-sdk was removed since we don't need IPC for v1 (config comes via CLI args, credentials via TES env vars). tokio is introduced in PR 2 (config + main entry point). Updated the PR description and commit message to reflect the actual contents

src/main.rs
utsavvyas: Agreed — removed. Module compilation is already verified by cargo build/cargo test. No tests in this scaffolding PR; unit and integration tests start in PR 2.

src/main.rs
utsavvyas: Good suggestion — changed to use gg_log_manager::{config, credentials, disk, scanner, uploader}. This avoids double-compilation and keeps the binary depending on the lib crate as the single source of truth.

recipe.yaml
utsavvyas: Good callout. We've already deployed and validated this component E2E on EC2 (AL2023 + GG Nucleus). The binary starts, reads config, and handles SIGTERM with a 4-step graceful shutdown. The Windows manifest doesn't need install (no chmod needed) or explicit shutdown (GG Nucleus handles Windows process termination via TerminateProcess). Added a comment clarifying this. The Linux manifests already have both install and shutdown lifecycle steps.

alter-mage (approved):

PR #240, opened 2026-04-20
feat(scanner): directory scanner, file reader, content hashing

paabbs:
**Issue #:** N/A (Task 1.3 — LogManager Rust rewrite)

**Description of changes:**
File scanner subsystem — directory scanning, offset-tracked file reading, and content hashing for log rotation detection.

- `scan_directory()`: list files matching regex, sort by mtime, mark newest as active
- `compute_content_hash()`: SHA-256 Base64-encoded (backward compatible with Java LogManager checkpoints)
- `read_file_from_offset()`: offset-tracked reader with timestamp extraction (ISO-8601 + epoch-millis)
- Invalid UTF-8 replaced with U+FFFD (matching Java behavior)
- Truncation detection: reset to offset 0 if file shrinks
- Event size truncation at CW Logs 256KB limit
- Clippy/rustfmt fixes for merged #239 code

Core file scanning infrastructure for the Rust LogManager. Required by the upload pipeline to discover, read, and track log/EMF files across rotations.

- `cargo build` ✅
- `cargo clippy -- -D warnings` ✅
- `cargo fmt --check` ✅
- `cargo test` ✅ — 95 tests, 0 failures

**Checklist:**

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -120,6 +120,7 @@ fn persist_checkpoints() {
 /// Subscribes to configuration updates via GG Component SDK IPC.
 /// Re-validates and applies new configuration without restart.
 #[cfg(target_os = "linux")]
+#[allow(dead_code, reason = "wired in when GG SDK IPC is integrated")]
 fn subscribe_to_config_updates() {
     // TODO: Wire up with gg_sdk::Sdk::subscribe_to_configuration_update()
 }
--- a/src/scanner/reader.rs
+++ b/src/scanner/reader.rs
@@ -2,3 +2,411 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! File reader with offset tracking and content hashing
+
+use base64::{engine::general_purpose::STANDARD, Engine};
+use sha2::{Digest, Sha256};
+use std::fs::File;
+use std::io::{Read, Seek, SeekFrom};
+use std::path::Path;
+
+/// Maximum event size in bytes (CloudWatch Logs limit: 256KB - 8 timestamp - 26 overhead)
+const MAX_EVENT_SIZE: usize = 262_110;
+
+/// A parsed log event with timestamp and message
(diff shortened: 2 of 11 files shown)


src/main.rs
```diff
@@ -7,9 +7,13 @@
 //! Tails log files and EMF JSON files, uploads to CloudWatch Logs.
 
 mod config;
+#[allow(unused)]
```
archigup: Probably should use expect instead of allow to ensure it gets removed later

paabbs: I've tried it with `#[expect(unused)]` but it doesn't work as the lint fires on items inside the module (in the binary crate), not on the mod declaration itself. Putting `expect(dead_code)` on individual functions also doesn't work because they're reachable from the lib crate (tests use them). So went ahead with this as a backup

src/scanner/reader.rs
```diff
@@ -2,3 +2,421 @@
...
+
+    loop {
+        let mut line = String::new();
+        match reader.read_line(&mut line) {
+            Ok(0) => break,
+            Ok(bytes_read) => {
+                current_offset += bytes_read as u64;
+            }
+            Err(e) if e.kind() == std::io::ErrorKind::InvalidData => {
+                tracing::warn!(path = %path.display(), offset = current_offset, "Invalid UTF-8, skipping byte");
```
limingyaoo: this could be slow on files with many invalid UTF-8, consider using string::from_utf8_lossy instead

paabbs: Makes sense, will send a revision acknowledging this

paabbs:
Rebased the incoming changes from #239 and switched from byte-by-byte skip to from_utf8_lossy as in invalid bytes are now replaced with U+FFFD instead of being dropped

limingyaoo (approved):

archigup (approved):

PR #241, opened 2026-04-22
feat(scanner): multi-line log assembly

paabbs:
## Multi-line log assembly (task 1.4)

Implements multiline log event grouping for the Rust LogManager, matching Java LogManager behavior.

### Changes
- `assemble_multiline()`: buffers lines until start pattern match, emits concatenated (no separator, matching Java's `StringBuilder.append`)
- Default pattern `^[\[(]?\d{4}-\d\d-\d\d|^\{` applied when `multiLineStartPattern` is null (matching Java's `DEFAULT_MULTILINE_PATTERN`)
- Timestamp fallback uses `SystemTime::now()` (matching Java's `Instant.now()`)
- Accepts `Vec<LogEvent>` directly, eliminating double timestamp parsing
- Oversized multiline truncation at 256KB CW limit with first-line context in warning

### Java behavioral parity
| Behavior | Java | Rust | Match |
|---|---|---|---|
| Default pattern when null | `DEFAULT_MULTILINE_PATTERN` | Same via `LazyLock` | ✅ |
| Join separator | `StringBuilder.append()` — none | `buffer.join("")` — none | ✅ |
| Timestamp fallback | `Instant.now()` | `SystemTime::now()` | ✅ |
| MAX_EVENT_LENGTH | 262,110 | 262,110 | ✅ |

### Known gaps (tracked separately in Asana)
- Oversized events: Java chunks, Rust truncates
- Log level filtering: not yet implemented (batcher scope)

### Tests
14 unit tests + 2 integration tests, 0 failures

### DIFF
--- a/src/scanner/multiline.rs
+++ b/src/scanner/multiline.rs
@@ -2,3 +2,263 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! Multi-line log assembly
+
+use super::reader::{extract_timestamp, LogEvent};
+use super::MAX_EVENT_SIZE;
+use regex::Regex;
+use std::sync::LazyLock;
+use std::time::SystemTime;
+
+/// Default multiline start pattern matching Java LogManager behavior.
+/// Matches lines starting with a date (optionally in brackets/parens) or JSON object.
+static DEFAULT_MULTILINE_PATTERN: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^[\[(]?\d{4}-\d\d-\d\d|^\{").unwrap());
+
+/// Assemble lines into LogEvents based on multi-line start pattern.
+/// When start_pattern is None, the default pattern is applied (matching Java behavior).
+/// When start_pattern is Some, that pattern is used.
+/// Lines are buffered until a new match, then emitted concatenated (no separator, matching Java).
+#[must_use = "assembled events must be consumed"]
+pub fn assemble_multiline(lines: Vec<LogEvent>, start_pattern: Option<&Regex>) -> Vec<LogEvent> {
+    let pattern = start_pattern.unwrap_or(&DEFAULT_MULTILINE_PATTERN);
+    let mut events = Vec::with_capacity(lines.len());
+    let mut buffer: Vec<String> = Vec::new();
+    let mut first_timestamp: Option<i64> = None;
+
+    for event in lines {
+        if pattern.is_match(&event.message) && !buffer.is_empty() {
+            events.push(emit_buffered(&buffer, first_timestamp));
+            buffer.clear();
+            first_timestamp = None;
+        }
+        if !event.message.trim().is_empty() {
+            if buffer.is_empty() {
+                first_timestamp = Some(event.timestamp);
+            }
+            buffer.push(event.message);
+        }
+    }
+    if !buffer.is_empty() {
+        events.push(emit_buffered(&buffer, first_timestamp));
+    }
+    tracing::debug!(event_count = events.len(), "Multiline assembly complete");
+    events
+}
+
+fn emit_buffered(buffer: &[String], first_event_timestamp: Option<i64>) -> LogEvent {
+    debug_assert!(!buffer.is_empty(), "emit_buffered called with empty buffer");
+    let timestamp = first_event_timestamp
+        .and_then(|ts| if ts != 0 { Some(ts) } else { None })
+        .or_else(|| extract_timestamp(&buffer[0]))
+        .unwrap_or_else(|| {
+            SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)
+                .map(|d| d.as_millis() as i64)
+                .unwrap_or(0)
+        });
+    // Join without separator to match Java's StringBuilder.append(partialLogLine) behavior.
+    let joined = buffer.join("");
+    let message = if joined.len() <= MAX_EVENT_SIZE {
+        joined
+    } else {
+        let first_line = &buffer[0][..80.min(buffer[0].len())];
+        tracing::warn!(
+            original_size = joined.len(),
+            max_size = MAX_EVENT_SIZE,
+            first_line,
+            "Truncating oversized multiline event"
+        );
+        let mut end = MAX_EVENT_SIZE;
+        while !joined.is_char_boundary(end) {
+            end -= 1;
+        }
+        joined[..end].to_string()
+    };
+    LogEvent { timestamp, message }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn ev(ts: i64, msg: &str) -> LogEvent {
+        LogEvent {
+            timestamp: ts,
+            message: msg.into(),
+        }
+    }
+
+    #[test]
+    fn test_default_pattern_groups_non_matching_lines() {
+        // With None pattern, the default Java pattern is applied.
+        // Lines without date/JSON prefix are grouped as continuations.
+        let lines = vec![ev(1000, "line1"), ev(1000, "line2"), ev(1000, "line3")];
+        let events = assemble_multiline(lines, None);
+        assert_eq!(events.len(), 1);
+        assert_eq!(events[0].message, "line1line2line3");
+    }
+
+    #[test]
+    fn test_default_pattern_splits_on_date() {
+        let lines = vec![
+            ev(1000, "2024-01-15 first event"),
+            ev(1000, "  continuation"),
+            ev(1000, "2024-01-16 second event"),
+        ];
+        let events = assemble_multiline(lines, None);
+        assert_eq!(events.len(), 2);
+        assert_eq!(events[0].message, "2024-01-15 first event  continuation");
+        assert_eq!(events[1].message, "2024-01-16 second event");
+    }
+
+    #[test]
+    fn test_default_pattern_splits_on_json() {
+        let lines = vec![ev(1000, r#"{"key":"val1"}"#), ev(1000, r#"{"key":"val2"}"#)];
+        let events = assemble_multiline(lines, None);
+        assert_eq!(events.len(), 2);
+    }
+
+    #[test]
+    fn test_multiline_assembly() {
+        let pattern = Regex::new(r"^\d{4}-\d{2}-\d{2}").unwrap();
+        let lines = vec![
+            ev(1705276800000, "2024-01-15 ERROR something"),
+            ev(0, "  stack trace line 1"),
+            ev(0, "  stack trace line 2"),
+            ev(1705363200000, "2024-01-16 INFO next event"),
+        ];
+        let events = assemble_multiline(lines, Some(&pattern));
+        assert_eq!(events.len(), 2);
+        assert_eq!(
+            events[0].message,
+            "2024-01-15 ERROR something  stack trace line 1  stack trace line 2"
+        );
+        assert_eq!(events[1].message, "2024-01-16 INFO next event");
+    }
+
+    #[test]
+    fn test_eof_flush() {
+        let pattern = Regex::new(r"^START").unwrap();
+        let lines = vec![ev(1000, "START event1"), ev(0, "continuation")];
+        let events = assemble_multiline(lines, Some(&pattern));
+        assert_eq!(events.len(), 1);
+        assert_eq!(events[0].message, "START event1continuation");
+    }
+
+    #[test]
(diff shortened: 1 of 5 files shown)


limingyaoo (approved):

src/scanner/multiline.rs
```diff
@@ -2,3 +2,220 @@
...
+
+use super::reader::{extract_timestamp, LogEvent};
+use super::MAX_EVENT_SIZE;
+use regex::Regex;
+
+/// Assemble lines into LogEvents based on multi-line start pattern.
+/// When start_pattern is None, each line becomes its own LogEvent.
+/// When start_pattern is Some, buffer lines until a new match, emit buffered joined by '\n'.
+pub fn assemble_multiline(
+    lines: Vec<String>,
```
alter-mage: **Intentional change: no default multiline pattern?**

The Java implementation applies a default pattern (`^[\\[(]?\\d{4}-\\d\\d-\\d\\d|^\\{`) when `multiLineStartPattern` is null, so multiline grouping is always active even without explicit configuration. This treats `None` as single-line mode — every line becomes its own event.

Is this intentional? Most components don't explicitly configure `multiLineStartPattern`, so they'd go from getting grouped stack traces to individual lines per event after upgrading. If the default pattern is being moved elsewhere in the pipeline (e.g., the caller always provides a pattern), that makes sense — just want to confirm.

src/scanner/multiline.rs
```diff
@@ -2,3 +2,220 @@
...
+    start_pattern: Option<&Regex>,
+    default_timestamp: i64,
```
alter-mage: Nit: consider adding `#[must_use = "assembled events must be consumed"]` to `assemble_multiline`. This is a pure function with no side effects — `#[must_use]` would let the compiler catch accidental unused results.

src/scanner/multiline.rs
```diff
@@ -2,3 +2,220 @@
...
+            events
+        }
+    };
+    tracing::debug!(event_count = events.len(), "Multiline assembly complete");
+    events
+}
+
+fn emit_buffered(buffer: &[String], default_timestamp: i64) -> LogEvent {
+    let timestamp = extract_timestamp(&buffer[0]).unwrap_or(default_timestamp);
+    let joined = buffer.join("\n");
```
alter-mage: **Intentional change: newline separator in multiline joining?**

The Java implementation joins continuation lines without any separator (`data.append(partialLogLine)`), so a stack trace ends up as one long concatenated string. This joins with `"\n"`, which preserves readability in CloudWatch.

Is this an intentional improvement? The Rust output is arguably more correct, but it's a behavioral change — any customer tooling (metric filters, Insights queries, alarms) that pattern-matches on the old concatenated format would see different output after upgrading. Just want to confirm this is a deliberate decision.

src/scanner/multiline.rs
```diff
@@ -2,3 +2,220 @@
...
+            }
+            events
+        }
+    };
+    tracing::debug!(event_count = events.len(), "Multiline assembly complete");
+    events
+}
+
+fn emit_buffered(buffer: &[String], default_timestamp: i64) -> LogEvent {
+    let timestamp = extract_timestamp(&buffer[0]).unwrap_or(default_timestamp);
```
alter-mage: **Timestamp fallback behavior differs from Java LogManager**

The Java implementation falls back to `Instant.now()` when no timestamp can be parsed from the assembled event. This falls back to a caller-provided `default_timestamp`, which has no guarantee of being reasonable.

If `default_timestamp` is stale or `0`, CloudWatch Logs will reject events older than 14 days — resulting in silent data loss. Consider using `SystemTime::now()` as the fallback here (matching Java behavior) or at minimum emitting a `tracing::warn!` when the fallback is used, so it's observable.

src/scanner/multiline.rs
```diff
@@ -2,3 +2,220 @@
...
+    let joined = buffer.join("\n");
+    let message = if joined.len() <= MAX_EVENT_SIZE {
```
alter-mage, line 61: **Oversized events are truncated instead of split — data loss regression**

The Java implementation splits oversized events into multiple CloudWatch log events by chunking at `MAX_EVENT_LENGTH` byte boundaries, preserving all data. This truncates at `MAX_EVENT_SIZE` and discards the remainder.

For a 400KB multiline event:
- Java: 2 CW events (256KB + 144KB) — no data loss
- Rust: 1 CW event (256KB) — 144KB silently dropped

This could affect components that produce large log entries (verbose stack traces, serialized payloads). Consider matching the Java chunking behavior to avoid data loss.

alter-mage (commented):
Thorough review covering behavioral differences vs Java LogManager, Rust idioms, and pipeline architecture questions. 11 inline comments below.

src/scanner/multiline.rs
```diff
@@ -2,3 +2,220 @@
...
+        joined
+    } else {
+        tracing::warn!(
```
alter-mage, line 65: **Truncation warning needs more context for debugging**

This logs `original_size` and `max_size`, but nothing to identify which event triggered it. On a device with many components, this isn't enough to diagnose the source. Consider including the first ~80 chars of the first line:
```rust
first_line = &buffer[0][..80.min(buffer[0].len())],
```
Ideally the source file path or component name would be threaded through too, but even just the first line snippet would make this actionable.

src/scanner/multiline.rs
```diff
@@ -2,3 +2,220 @@
...
+            original_size = joined.len(),
+            max_size = MAX_EVENT_SIZE,
+            "Truncating oversized multiline event"
+        );
+        let mut end = MAX_EVENT_SIZE;
+        while !joined.is_char_boundary(end) {
```
alter-mage, line 72: **Double allocation on oversized event truncation**

`emit_buffered` first allocates the full joined string (potentially megabytes), then `to_string()` allocates a second copy for the truncated slice. For a 400KB+ event, that's ~660KB of transient heap usage on resource-constrained IoT devices.

Consider building the string incrementally with a size cap:
```rust
let mut message = String::with_capacity(MAX_EVENT_SIZE.min(estimated_size));
for (i, line) in buffer.iter().enumerate() {
    if i > 0 {
        if message.len() + 1 > MAX_EVENT_SIZE { break; }
        message.push('\n');
    }
    let remaining = MAX_EVENT_SIZE - message.len();
    if line.len() <= remaining {
        message.push_str(line);
    } else {
        let mut end = remaining;
        while !line.is_char_boundary(end) { end -= 1; }
        message.push_str(&line[..end]);
        break;
    }
}
```
This caps memory at `MAX_EVENT_SIZE` and avoids the double allocation.

src/scanner/reader.rs
```diff
@@ -3,15 +3,13 @@
 
 //! File reader with offset tracking and content hashing
 
+use super::MAX_EVENT_SIZE;
```
alter-mage, line 6: **Timestamp parsing gaps vs Java implementation**

A couple of differences in `extract_timestamp`:

1. **Timezone suffixes**: Parses exactly 19 characters of ISO-8601 (`YYYY-MM-DDTHH:MM:SS`) and ignores any timezone suffix. Lines like `2024-01-15T10:30:45Z` or `2024-01-15T10:30:45+05:30` would be treated as UTC. Java handles these via `ISO_INSTANT` and `ISO_OFFSET_DATE_TIME` formatters.

2. **JSON structured logs**: Java has a `tryGetStructuredLogMessage` path that deserializes JSON logs and extracts the timestamp from the JSON field. Nucleus supports JSON log format (opt-in), so customers using it would lose accurate timestamp extraction.

The timezone issue affects the default text format and is probably higher priority.

tests/multiline_integration_test.rs
```diff
@@ -0,0 +1,60 @@
...
+#[test]
+fn test_multiline_with_scanner() {
+    let mut file = NamedTempFile::new().unwrap();
+    writeln!(file, "2024-01-15T10:00:00 ERROR NullPointerException").unwrap();
+    writeln!(file, "    at com.example.App.main(App.java:42)").unwrap();
+    writeln!(file, "    at java.lang.Thread.run(Thread.java:750)").unwrap();
+    writeln!(file, "2024-01-15T10:00:01 INFO Recovery complete").unwrap();
+
+    let (lines, _) = read_file_from_offset(file.path(), 0, 1000).unwrap();
+    let line_strs: Vec<String> = lines.into_iter().map(|e| e.message).collect();
```
alter-mage: **`assemble_multiline` takes `Vec<String>` but the upstream produces `Vec<LogEvent>`**

`read_file_from_offset` already parses timestamps into `LogEvent`, but callers have to discard them (`.map(|e| e.message).collect()`) to feed `assemble_multiline`, which then re-parses timestamps from the raw strings. This means timestamps are parsed twice, and the two code paths could theoretically disagree.

Consider accepting `Vec<LogEvent>` directly — merge messages and keep the timestamp from the first event in each group. This would eliminate the redundant parse and make the pipeline type-safe. Not necessarily a blocker for this PR, but worth considering before the API solidifies.

alter-mage:
**Is log level filtering handled elsewhere?**

Java's `processLogLine` drops structured log events below `desiredLogLevel` before they become CloudWatch events. `assemble_multiline` doesn't do any log level filtering. Is this handled in a different layer of the Rust pipeline? If not, customers who configured `minimumLogLevel` would see all log levels uploaded, increasing CloudWatch costs.

alter-mage:
**Are CloudWatch batch constraints handled elsewhere?**

Java enforces several CW Logs API limits during assembly: max batch size (1MB), max 10,000 events per batch, 14-day age rejection, and 23-hour time span within a batch. `assemble_multiline` doesn't enforce any of these. Is there a separate batching layer in the Rust pipeline that handles this? If not, unbounded output from `assemble_multiline` could lead to CW API rejections.

src/scanner/multiline.rs
paabbs: Valid concern. Will fix — tracking as a follow-up to this PR, and change the LLD accordingly to this.

src/scanner/multiline.rs
paabbs: Good catch. Will send a revision

src/scanner/multiline.rs
paabbs: This was intentional. Preserves readability in CloudWatch vs Java's concatenation without separator. Although I can revise this to follow the Java's convention, but I feed like this should be one of the fixes in the Java version.

src/scanner/multiline.rs
paabbs: Good catch, will follow this

src/scanner/multiline.rs
paabbs: Acknowledged that Java chunks instead — tracking as a future improvement, not in scope for this PR.

src/scanner/multiline.rs
paabbs: Will add first line snippet in the revision

src/scanner/multiline.rs
paabbs: Ack. Will be addressed when chunking is implemented. Not in scope for this PR.

src/scanner/reader.rs
paabbs: Valid. Timezone suffixes and JSON structured log timestamps are not handled in reader.rs — will fix in a separate PR targeting the scanner layer.

tests/multiline_integration_test.rs
paabbs: Valid. Will refactor assemble_multiline to accept Vec<LogEvent> in the revision.

paabbs:
> **Is log level filtering handled elsewhere?**
> 
> Java's `processLogLine` drops structured log events below `desiredLogLevel` before they become CloudWatch events. `assemble_multiline` doesn't do any log level filtering. Is this handled in a different layer of the Rust pipeline? If not, customers who configured `minimumLogLevel` would see all log levels uploaded, increasing CloudWatch costs.

Will cover this as a part of batcher/uploader PR

paabbs:
> Java enforces several CW Logs API limits during assembly: max batch size (1MB), max 10,000 events per batch, 14-day age rejection, and 23-hour time span within a batch. `assemble_multiline` doesn't enforce any of these. Is there a separate batching layer in the Rust pipeline that handles this? If not, unbounded output from `assemble_multiline` could lead to CW API rejections.

Same here, will cover them as a part of the batcher PR

alter-mage (approved):

PR #242, opened 2026-04-22
feat(scanner): checkpoint persistence and restart recovery

paabbs:
### Changes
- `FileCheckpoint` with backward-compatible serde field names (V2 format)
- `FileCheckpoint::new()` auto-sets `last_accessed` to now
- Atomic write (write-tmp + fsync file + rename + fsync parent directory)
- Tmp file cleanup on write error
- Load with corrupt/missing file fallback
- Startup stale trim: skip entries where `lastModifiedTime <= lastProcessedTimestamp`
- `recover_offsets`: resume from persisted byte offset per content hash
- Touch-on-read: update `last_accessed` on checkpoint access
- `remove_component`: cleanup when component removed from config
- Config dedup: duplicate component names deduplicated (last-wins)
- I/O errors include file path context

### Deferred to upload path
- TTL eviction (24h) — Java runs on every `put()` after upload
- Move scan-time eviction to upload path (documented DIVERGENCE)
- Update offset after upload
- Completed file detection
- Throttled persistence
- Shutdown flush
- `put` vs `putIfAbsent` semantics (documented in `CheckpointStore` doc comment)

### Tests
15 unit tests + 2 integration tests, 0 failures

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -28,7 +28,24 @@ pub fn load_config(config_arg: &str) -> Result<LogManagerConfig, ConfigError> {
 
 #[must_use = "config result must be handled"]
 pub fn parse_config(json: &str) -> Result<LogManagerConfig, ConfigError> {
-    serde_json::from_str(json).map_err(|e| ConfigError::Parse(format!("{e}")))
+    let mut config: LogManagerConfig =
+        serde_json::from_str(json).map_err(|e| ConfigError::Parse(format!("{e}")))?;
+    // Deduplicate by component name, last-wins.
+    let mut seen = std::collections::HashMap::new();
+    for (i, c) in config.component_logs_configuration.iter().enumerate().rev() {
+        seen.entry(c.component_name.clone()).or_insert(i);
+    }
+    if seen.len() < config.component_logs_configuration.len() {
+        let indices: std::collections::HashSet<usize> = seen.into_values().collect();
+        config.component_logs_configuration = config
+            .component_logs_configuration
+            .into_iter()
+            .enumerate()
+            .filter(|(i, _)| indices.contains(i))
+            .map(|(_, c)| c)
+            .collect();
+    }
+    Ok(config)
 }
 
 /// Validate a single log source's directory, regex, and disk limit.
@@ -248,5 +265,32 @@ mod tests {
         let result = load_config("/nonexistent/config.json");
         assert!(result.is_err());
         assert!(result.unwrap_err().to_string().contains("IO error"));
+    }
+
+    #[test]
+    fn test_duplicate_component_name_last_wins() {
+        let json = r#"{
+            "componentLogsConfiguration": [
+                {
+                    "componentName": "MyApp",
+                    "logFileDirectoryPath": "/first/dir",
+                    "logFileRegex": ".*\\.log"
+                },
+                {
+                    "componentName": "MyApp",
+                    "logFileDirectoryPath": "/second/dir",
+                    "logFileRegex": ".*\\.log"
+                }
+            ]
+        }"#;
+        let config = parse_config(json).unwrap();
+        // Java's Map.put deduplicates by name, last wins
+        assert_eq!(config.component_logs_configuration.len(), 1);
+        assert_eq!(
+            config.component_logs_configuration[0]
+                .source
+                .log_file_directory_path,
+            "/second/dir"
+        );
     }
 }
--- a/src/scanner/checkpoint.rs
+++ b/src/scanner/checkpoint.rs
@@ -2,3 +2,604 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! Checkpoint persistence with atomic writes and restart recovery.
+
+use serde::{Deserialize, Serialize};
+use std::collections::{HashMap, HashSet};
+use std::fs;
+use std::io;
+use std::path::{Path, PathBuf};
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use super::ScannedFile;
+
+/// Per-file checkpoint state. Compatible with Java LogManager v2.3.1+ config.tlog.
+/// Java's deprecated `currentProcessingFileName` field is ignored on read, not written.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct FileCheckpoint {
+    #[serde(rename = "currentProcessingFileHash")]
+    pub file_hash: String,
+    #[serde(rename = "currentProcessingFileStartPosition")]
+    pub start_position: u64,
+    #[serde(rename = "currentProcessingFileLastModified")]
+    pub last_modified_time: u64,
+    /// Epoch millis when this entry was last accessed. Defaults to now on construction.
+    #[serde(rename = "lastAccessed")]
+    pub last_accessed: u64,
+}
+
+impl FileCheckpoint {
+    /// Create a new checkpoint entry. `last_accessed` is set to now automatically.
+    pub fn new(file_hash: String, start_position: u64, last_modified_time: u64) -> Self {
+        Self {
+            file_hash,
+            start_position,
+            last_modified_time,
+            last_accessed: now_ms(),
+        }
+    }
+}
+
+/// Timestamp of the last fully processed file for a component.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct LastFileProcessedTimestamp {
+    #[serde(rename = "lastFileProcessedTimeStamp")]
+    pub last_file_processed_time_stamp: u64,
+}
+
+/// Checkpoint store for all components.
+///
+/// # Write semantics (for batcher implementer)
+/// - Load path: use `putIfAbsent` — don't overwrite existing entries
+/// - Upload path: use `put` — overwrite with new startPosition + trigger TTL eviction
+#[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq)]
+pub struct CheckpointStore {
+    #[serde(rename = "currentComponentFileProcessingInformationV2", default)]
+    pub file_processing_info: HashMap<String, HashMap<String, FileCheckpoint>>,
+    #[serde(rename = "componentLastFileProcessedTimeStamp", default)]
+    pub last_processed_timestamps: HashMap<String, LastFileProcessedTimestamp>,
+}
+
+/// Save checkpoint atomically: write to temp file, then rename.
+///
+/// # Errors
+/// Returns `io::Error` if the file cannot be created, written, synced, or renamed.
+pub fn save_checkpoint(path: &Path, store: &CheckpointStore) -> io::Result<()> {
+    let tmp_path = path.with_extension("tmp");
+    let result = (|| -> io::Result<()> {
+        let file = fs::File::create(&tmp_path)
+            .map_err(|e| io::Error::new(e.kind(), format!("create {}: {e}", tmp_path.display())))?;
+        serde_json::to_writer_pretty(&file, store)
+            .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
+        file.sync_all()
+            .map_err(|e| io::Error::new(e.kind(), format!("sync {}: {e}", tmp_path.display())))?;
+        fs::rename(&tmp_path, path)
+            .map_err(|e| io::Error::new(e.kind(), format!("rename to {}: {e}", path.display())))?;
+        // Fsync parent directory to ensure the rename is durable on power loss
+        if let Some(parent) = path.parent() {
+            fs::File::open(parent)?.sync_all()?;
+        }
+        Ok(())
+    })();
+    if result.is_err() {
+        let _ = fs::remove_file(&tmp_path);
+    }
+    result?;
+    tracing::debug!(path = %path.display(), "Checkpoint saved");
+    Ok(())
+}
+
+/// Load checkpoint from disk. Returns empty store if file missing or corrupt.
+///
+/// # Errors
+/// Returns `io::Error` on read failures other than `NotFound`.
+pub fn load_checkpoint(path: &Path) -> io::Result<CheckpointStore> {
+    match fs::read_to_string(path) {
+        Ok(content) => match serde_json::from_str::<CheckpointStore>(&content) {
+            Ok(store) => {
+                let entry_count: usize = store.file_processing_info.values().map(|m| m.len()).sum();
+                tracing::debug!(path = %path.display(), entry_count = entry_count, "Checkpoint loaded");
+                Ok(store)
+            }
+            Err(e) => {
+                tracing::warn!("Corrupt checkpoint file, starting fresh: {}", e);
+                Ok(CheckpointStore::default())
+            }
+        },
+        Err(e) if e.kind() == io::ErrorKind::NotFound => Ok(CheckpointStore::default()),
+        Err(e) => Err(io::Error::new(
+            e.kind(),
+            format!("read {}: {e}", path.display()),
+        )),
+    }
+}
+
+fn now_ms() -> u64 {
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map(|d| d.as_millis() as u64)
+        .unwrap_or(0)
+}
+
+/// Trim stale entries where lastModifiedTime is at or before the component's
+/// lastFileProcessedTimeStamp. Prevents loading already-completed entries from older LM versions.
+pub fn trim_stale_on_load(store: &mut CheckpointStore) {
+    let ts_snapshot: HashMap<&str, u64> = store
+        .last_processed_timestamps
+        .iter()
+        .map(|(k, v)| (k.as_str(), v.last_file_processed_time_stamp))
+        .collect();
+
+    for (component, files) in &mut store.file_processing_info {
+        if let Some(&cutoff) = ts_snapshot.get(component.as_str()) {
+            let before = files.len();
+            files.retain(|_, cp| cp.last_modified_time > cutoff);
+            let removed = before - files.len();
+            if removed > 0 {
+                tracing::info!(
+                    component = component.as_str(),
+                    removed,
+                    "Trimmed stale entries on load"
+                );
+            }
+        }
+    }
+}
+
+/// Recover file offsets from checkpoint for restart recovery.
+/// Returns (file_path, resume_offset) pairs for each scanned file.
+/// Updates last_accessed on read (touch-on-read).
+/// Evicts entries for files no longer on disk (see DIVERGENCE comment in body).
+///
+/// The caller must pass ALL files for the component in a single call.
+/// One directory per component — the config schema enforces this.
+#[must_use = "recovered offsets must be used for file reading"]
+pub fn recover_offsets(
+    checkpoint: &mut CheckpointStore,
+    log_group_key: &str,
+    scanned_files: &[ScannedFile],
+) -> Vec<(PathBuf, u64)> {
+    let current_hashes: HashSet<&str> = scanned_files
+        .iter()
+        .map(|f| f.content_hash.as_str())
+        .collect();
+    let now = now_ms();
+
+    let result: Vec<(PathBuf, u64)> = scanned_files
+        .iter()
+        .map(|file| {
+            let offset = checkpoint
+                .file_processing_info
+                .get_mut(log_group_key)
+                .and_then(|m| m.get_mut(&file.content_hash))
+                .map(|cp| {
+                    cp.last_accessed = now; // touch-on-read
+                    tracing::debug!(
+                        "Resuming file {} from offset {}",
+                        file.path.display(),
+                        cp.start_position
+                    );
+                    cp.start_position
+                })
+                .unwrap_or_else(|| {
+                    tracing::debug!("New file {}, reading from start", file.path.display());
+                    0
+                });
+            (file.path.clone(), offset)
+        })
+        .collect();
+
+    // DIVERGENCE: Java evicts via TTL on every put() call (triggered by upload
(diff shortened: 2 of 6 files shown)


src/config/mod.rs
```diff
@@ -28,7 +28,24 @@ pub fn load_config(config_arg: &str) -> Result<LogManagerConfig, ConfigError> {
 
 #[must_use = "config result must be handled"]
 pub fn parse_config(json: &str) -> Result<LogManagerConfig, ConfigError> {
-    serde_json::from_str(json).map_err(|e| ConfigError::Parse(format!("{e}")))
+    let mut config: LogManagerConfig =
+        serde_json::from_str(json).map_err(|e| ConfigError::Parse(format!("{e}")))?;
+    // Deduplicate by component name, last-wins (matching Java's Map.put behavior).
```
utsavvyas: this comment doesnt add any value

src/scanner/checkpoint.rs
```diff
@@ -2,3 +2,557 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! Checkpoint persistence with atomic writes and restart recovery.
+//!
+//! Field names use Java LogManager's naming convention for backward compatibility
```
utsavvyas: I think this comment will not add any value later.

src/scanner/checkpoint.rs
```diff
@@ -2,3 +2,557 @@
...
+pub struct CheckpointStore {
+    #[serde(rename = "currentComponentFileProcessingInformationV2", default)]
+    pub file_processing_info: HashMap<String, HashMap<String, FileCheckpoint>>,
+    #[serde(rename = "componentLastFileProcessedTimeStamp", default)]
+    pub last_processed_timestamps: HashMap<String, LastFileProcessedTimestamp>,
+}
+
+/// Save checkpoint atomically: write to temp file, then rename.
+pub fn save_checkpoint(path: &Path, store: &CheckpointStore) -> io::Result<()> {
+    let tmp_path = path.with_extension("tmp");
```
utsavvyas, line 67: Do we need to use some gg directory ? Wouldnt it might get deleted ?

src/scanner/checkpoint.rs
```diff
@@ -2,3 +2,557 @@
...
+/// the component's lastFileProcessedTimeStamp. Matches Java's loadProcessingFilesConfig
+/// stale trim that prevents loading already-completed entries from old LM versions.
+pub fn trim_stale_on_load(store: &mut CheckpointStore) {
+    for (component, files) in &mut store.file_processing_info {
+        if let Some(ts) = store.last_processed_timestamps.get(component.as_str()) {
+            let before = files.len();
+            files.retain(|_, cp| cp.last_modified_time > ts.last_file_processed_time_stamp);
+            let removed = before - files.len();
+            if removed > 0 {
+                tracing::debug!(component, removed, "Trimmed stale entries on load");
```
utsavvyas: maybe info, it will only loaded when you load it for first time.

src/scanner/checkpoint.rs
```diff
@@ -2,3 +2,557 @@
...
+            },
+        );
+
+        save_checkpoint(&path, &store).unwrap();
+        let loaded = load_checkpoint(&path).unwrap();
+        assert_eq!(store, loaded);
+    }
+
+    #[test]
+    fn test_json_field_names_match_java() {
```
utsavvyas, line 255: java naming doesnt make sense. It makes readablility bad

src/scanner/checkpoint.rs
```diff
@@ -2,3 +2,557 @@
...
+            file_hash,
+            start_position,
+            last_modified_time,
+            last_accessed: now_ms(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct LastFileProcessedTimestamp {
```
utsavvyas, line 44: Do we really need this struct ?

utsavvyas, line 44: Do we really need this struct ?

src/scanner/checkpoint.rs
```diff
@@ -2,3 +2,557 @@
...
+    fn test_missing_file_returns_empty() {
+        let dir = tempdir().unwrap();
+        let path = dir.path().join("nonexistent.json");
+        let store = load_checkpoint(&path).unwrap();
+        assert!(store.file_processing_info.is_empty());
+        assert!(store.last_processed_timestamps.is_empty());
+    }
+
+    #[test]
+    fn test_corrupt_json_returns_empty() {
```
utsavvyas, line 341: Should it validate some metric or log ?

utsavvyas:
In general integ tests are missing.

src/config/mod.rs
paabbs: I can clean them up

src/scanner/checkpoint.rs
paabbs: I can clean them up

src/scanner/checkpoint.rs
paabbs: The checkpoint path should be provided by the caller imo.  On a real device, the recipe will pass {work:path}/checkpoint.json which resolves to {GG_ROOT}/work/gg-log-manager/ on both Classic and Lite. This directory persists across restarts and is only cleaned on component removal.

paabbs:
> In general integ tests are missing.

Thought of adding them as a part of the upload hookup but I can add the tests for loading from disk and checkpointing.

src/scanner/checkpoint.rs
paabbs: Thanks, will change it!

src/scanner/checkpoint.rs
paabbs: Will clean this up

src/scanner/checkpoint.rs
paabbs: The struct is needed for the serde field rename (lastFileProcessedTimeStamp). Although wthout it, we'd need a custom serializer to produce the correct JSON key name

paabbs: Duplicate of the last one

src/scanner/checkpoint.rs
paabbs: The production path already emits `tracing::warn!("Corrupt checkpoint file, starting fresh: {}", e)` which will surface in device logs. Asserting on log output in a unit test requires pulling in tracing-test as a dev dependency(which I didn't wanna take) but happy to add that if you'd like the test to verify the warn fires. Will also add an assertion on `last_processed_timestamps.is_empty()` which is currently missing.

utsavvyas: I think that is fine but how do we validate that it didnt crap out. we should validate that atleast

src/scanner/checkpoint.rs
utsavvyas: I do believe gg specific folder do exist. It means we can use it for the checkpoint. What does java code uses ?

src/scanner/checkpoint.rs
```diff
@@ -2,3 +2,557 @@
...
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use super::ScannedFile;
+
+/// Per-file checkpoint state. Compatible with Java LogManager v2.3.1+ config.tlog.
+/// Java's deprecated `currentProcessingFileName` field is ignored on read, not written.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct FileCheckpoint {
+    #[serde(rename = "currentProcessingFileHash")]
+    pub file_hash: String,
```
alter-mage, line 20: `file_hash` is stored both as the `HashMap` key in `file_processing_info` and as a field inside `FileCheckpoint`. Since the key and the field always hold the same value, is there a reason to keep both? Could the field be removed and the hash derived from the map key at the call site when needed?

src/scanner/checkpoint.rs
```diff
@@ -2,3 +2,557 @@
...
+                tracing::warn!("Corrupt checkpoint file, starting fresh: {}", e);
+                Ok(CheckpointStore::default())
+            }
+        },
+        Err(e) if e.kind() == io::ErrorKind::NotFound => Ok(CheckpointStore::default()),
+        Err(e) => Err(e),
+    }
+}
+
+fn now_ms() -> u64 {
```
alter-mage, line 116: `now_ms()` is called directly inside `FileCheckpoint::new()` and `recover_offsets()`, making time non-injectable. Tests work around this with `before/after` bracketing which is fragile. Would it make sense to accept a timestamp parameter (or a clock abstraction) to make this deterministic in tests?

src/scanner/checkpoint.rs
```diff
@@ -2,3 +2,557 @@
...
+        .map(|d| d.as_millis() as u64)
+        .unwrap_or(0)
+}
+
+/// Trim entries loaded from a Java checkpoint where lastModifiedTime is at or before
+/// the component's lastFileProcessedTimeStamp. Matches Java's loadProcessingFilesConfig
+/// stale trim that prevents loading already-completed entries from old LM versions.
+pub fn trim_stale_on_load(store: &mut CheckpointStore) {
+    for (component, files) in &mut store.file_processing_info {
+        if let Some(ts) = store.last_processed_timestamps.get(component.as_str()) {
```
alter-mage: This compiles today because `file_processing_info` and `last_processed_timestamps` are disjoint fields, but it is fragile under refactoring. Would it be cleaner to collect the timestamps into a local map first, then mutate `file_processing_info` separately?

src/scanner/checkpoint.rs
```diff
@@ -2,3 +2,557 @@
...
+    #[serde(rename = "componentLastFileProcessedTimeStamp", default)]
+    pub last_processed_timestamps: HashMap<String, LastFileProcessedTimestamp>,
+}
+
+/// Save checkpoint atomically: write to temp file, then rename.
+pub fn save_checkpoint(path: &Path, store: &CheckpointStore) -> io::Result<()> {
+    let tmp_path = path.with_extension("tmp");
+    let json = serde_json::to_string_pretty(store)
+        .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
+    let mut file = fs::File::create(&tmp_path)?;
```
alter-mage: If `write_all` or `sync_all` returns an error, the `?` exits early and `tmp_path` is left on disk. Would it be worth adding cleanup on error (e.g. `let _ = fs::remove_file(&tmp_path)`) to avoid leaving partial files around?

src/scanner/checkpoint.rs
```diff
@@ -2,3 +2,557 @@
...
+}
+
+/// Recover file offsets from checkpoint for restart recovery.
+/// Returns (file_path, resume_offset) pairs for each scanned file.
+/// Updates last_accessed on read. Evicts entries for files no longer on disk.
+///
+/// The caller must pass ALL files for the component in a single call.
+/// Java enforces one directory per component via Map<String, ComponentLogConfiguration>.
+#[must_use = "recovered offsets must be used for file reading"]
+pub fn recover_offsets(
```
alter-mage, line 156: `recover_offsets` both returns file offsets and mutates the store by evicting entries for files no longer on disk. Should these be separate functions so a caller can recover offsets without triggering eviction as a side effect?

src/scanner/checkpoint.rs
```diff
@@ -2,3 +2,557 @@
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct LastFileProcessedTimestamp {
+    #[serde(rename = "lastFileProcessedTimeStamp")]
+    pub last_file_processed_time_stamp: u64,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq)]
+pub struct CheckpointStore {
```
alter-mage, line 55: All fields are `pub`, allowing callers to mutate internal state directly and bypass any invariants. Would it make sense to make these fields private and expose the operations (`recover_offsets`, `trim_stale_on_load`, etc.) as methods on `CheckpointStore` instead?

src/config/mod.rs
```diff
@@ -28,7 +28,24 @@ pub fn load_config(config_arg: &str) -> Result<LogManagerConfig, ConfigError> {
...
+    let mut seen = std::collections::HashMap::new();
+    for (i, c) in config.component_logs_configuration.iter().enumerate() {
+        seen.insert(c.component_name.clone(), i);
```
alter-mage: The last-wins dedup relies on `HashMap` overwriting earlier indices with later ones during iteration, then filtering by the surviving indices. The correctness is not immediately obvious to a reader. Would a more explicit approach — e.g. iterating in reverse and using `entry().or_insert()` — make the intent clearer?

src/scanner/checkpoint.rs
```diff
@@ -2,3 +2,557 @@
...
+                });
+            (file.path.clone(), offset)
+        })
+        .collect();
+
+    // Remove entries for files no longer on disk.
+    // TODO: Add TTL eviction (24h) in the upload path — Java runs evictStaleEntries()
+    // on put() after upload, not during scan. See Asana task under 1.6 Batcher.
+    if let Some(file_map) = checkpoint.file_processing_info.get_mut(log_group_key) {
+        file_map.retain(|hash, _| {
```
alter-mage, line 196: In Java, `ProcessingFiles.remove(deletedHashes)` is called explicitly from `deleteFileFromGroup` after a file has been successfully uploaded — meaning a checkpoint is only discarded once upload is confirmed. Here, entries for files not currently on disk are evicted on every scan cycle.

If a file disappears mid-rotation before the uploader processes it, its checkpoint gets evicted at the next scan. When the uploader runs, there is no offset to resume from — it would either re-upload from the start or skip the file entirely.

Is this divergence from Java intentional, or should eviction be deferred to the upload path?

src/scanner/checkpoint.rs
alter-mage, line 55: In Java, `ProcessingFiles` distinguishes between two write operations:
- `putIfAbsent` — used on load, does not overwrite existing entries
- `put` — used after upload, overwrites and triggers `evictStaleEntries()`

When the upload path is implemented in Rust, would it make sense to add a note or a stub here to ensure `put` semantics (overwrite + TTL eviction) are used on upload, rather than `putIfAbsent` semantics? Easy to get wrong when the batcher lands.

src/scanner/checkpoint.rs
paabbs: So the java version of log manager is a plugin of Nucleus component that uses ~~{GG_ROOT}/work/aws.greengrass.LogManager/checkpoint.json~~ Nucleus Topics API (`getRuntimeConfig().lookupTopics(...)`) which serializes to `{GG_ROOT}/config/config.tlog`, a shared transaction log for ALL components. But lite did not have any checkpointing behavior from what I have checked.

src/scanner/checkpoint.rs
paabbs: For that a couple of options at hand are-
- Return a flag:  load_checkpoint could return (CheckpointStore, bool) where the bool indicates whether it loaded successfully or fell back to empty. The caller can then decide to log a metric or alert.

- Metric addition (sth like `checkpoint_load_failures`):  `tracing::warn` for now and a dedicated metric can be added when the metrics infrastructure is wired up..

src/scanner/checkpoint.rs
utsavvyas: Do we want to update the checkpoint or you create new checkpoint everytime ? It is single thread so I dont expect to have race condition on overwrites

src/scanner/checkpoint.rs
utsavvyas: Yes, but we have to built it in lite as well

src/scanner/checkpoint.rs
utsavvyas: Dont we need the checkpoint persisted so that you know from where to start ?

src/scanner/checkpoint.rs
```diff
@@ -2,3 +2,557 @@
...
+
+/// Save checkpoint atomically: write to temp file, then rename.
+pub fn save_checkpoint(path: &Path, store: &CheckpointStore) -> io::Result<()> {
+    let tmp_path = path.with_extension("tmp");
+    let json = serde_json::to_string_pretty(store)
+        .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
+    let mut file = fs::File::create(&tmp_path)?;
+    file.write_all(json.as_bytes())?;
+    file.sync_all()?;
+    fs::rename(&tmp_path, path)?;
```
yitingb: At here, `sync_all()` flushes the file data and inode to disk, but `rename()` modifies the directory, which is a separate structure. The directory entry change sits in the kernel page cache until the OS flushes it — up to 5 seconds on ext4 (default commit interval). A power loss in that window reverts the directory to its previous state: `checkpoint.json` points to the old inode, and the new data is orphaned on disk.

On reboot, `load_checkpoint` reads the old offsets, and the scanner re-reads and re-uploads already-processed log data to CloudWatch.

Question here: are we allowing uploading duplicate logs here, if so, then there is no concern.

src/scanner/checkpoint.rs
paabbs: The `file_hash` field is redundant with the map key, but it's required for Java checkpoint wire format compatibility. Java's `CurrentProcessingFileInformation` serializes `currentProcessingFileHash` as a field in the JSON value object. Removing it would break cross-version checkpoint reads during Java→Rust upgrades or rollbacks

src/scanner/checkpoint.rs
paabbs: This would be covered as an integration test in the followup PR

src/scanner/checkpoint.rs
paabbs: Thanks, will implement this in the revision

src/scanner/checkpoint.rs
paabbs: Since the Rust LM is a generic component (not a Nucleus plugin), we don't have access to the Topics API, so we write our own `checkpoint.json` instead. The path is caller-controlled via `save_checkpoint(path, store)`.

For the recipe wiring, we need to pass `{work:path}` to the binary. The recipe `run` lifecycle will become something like:
```
 run: "{artifacts:path}/gg-log-manager --config '{configuration:/}' --work-dir '{work:path}'"
```
 This resolves to `{GG_ROOT}/work/gg-log-manager/` on both Classic and Lite. The checkpoint file lands at `{work:path}/checkpoint.json`. This directory is created by the Nucleus on deployment and persists across restarts — only cleaned on component removal.

The recipe + `--work-dir` arg parsing will be wired up in the main.rs integration PR

src/scanner/checkpoint.rs
paabbs: The cleanup is for the `.tmp` file only — not the actual `checkpoint.json`. If the write fails mid-way, `checkpoint.json.tmp` is left as a partial/corrupt file on disk. The real `checkpoint.json` (from the previous successful save) is untouched because the `rename` never executed. We clean up the orphaned `.tmp` to avoid leaving garbage on disk

src/scanner/checkpoint.rs
paabbs: Yes, they should be separate. Java's scan path only does offset lookup — no eviction. Currently we evict on scan as a temporary measure (marked with DIVERGENCE comment). Will move to upload path when batcher land.

src/scanner/checkpoint.rs
paabbs: Agree. Java's `ProcessingFiles` is fully encapsulated — `private final Map<String, Node> cache` with all access through `put/putIfAbsent/get/remove/toMap/getMostRecentlyUsed`. The Rust `CheckpointStore` currently exposes `pub` fields, which lets callers bypass invariants. I'll encapsulate the fields and add proper methods along with the batcher , since the methods needed (`update_offset`, `put`, `putIfAbsent`) don't exist yet. Adding them now would be dead API surface.

src/config/mod.rs
paabbs: Sure, I can switch to reverse iteration with `entry().or_insert()` to make the last-wins intent  moer explicit.

src/scanner/checkpoint.rs
paabbs: Thanks for the catch. I iwll be adding a doc comment on CheckpointStore noting the distinction for the batcher implementer

src/scanner/checkpoint.rs
paabbs: Yes, this was intentional and was deferred to upload path

src/scanner/checkpoint.rs
paabbs: Yes, the Java LogManager has at-least-once delivery semantics. Any crash between a successful CloudWatch upload and the next checkpoint persist can cause re-uploads on restart. This is true for both Java (tlog append lag + truncation rename gap) and our Rust implementation (write-tmp-rename gap).

That said, we should minimize the window. Java's normal path (append with `O_SYNC`) has no rename gap — only the periodic tlog truncation does. Our Rust code has the gap on every save. I can add a directory fsync after the rename to close it.

src/scanner/checkpoint.rs
alter-mage: Thanks for agreeing on this. Is the encapsulation fix planned for this PR or a follow-up? Just want to make sure it's tracked somewhere before this merges.

alter-mage (approved):

src/scanner/checkpoint.rs
paabbs: Yeah that would addressed in a followup PR for hooking the upload

yitingb (approved):

utsavvyas:
Did we confirm that the file location doesnt wipe out the data with new deployment ?

PR #245, opened 2026-04-24
fix: match Java diskSpaceLimit defaults (#243)

### LINKED ISSUE #243: Bug: diskSpaceLimitUnit defaults to MB in Rust, but KB in Java
## Summary

The Rust v2 LogManager defaults `diskSpaceLimitUnit` to `MB` when not specified, while the Java LogManager defaults to `KB`. This is a 1024x difference in disk space limit interpretation for configs that omit the unit.

## Java behavior

When `diskSpaceLimitUnit` is empty, Java defaults to `KB`:

https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/main/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java#L268-L270

```java
if (StringUtils.isEmpty(diskSpaceLimitUnit)) {
    diskSpaceLimitUnit = "KB";
}
```

## Rust behavior

The Rust version defaults to `MB`:

https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/rust/v2/src/config/schema.rs#L63-L67

```rust
impl Default for DiskSpaceLimitUnit {
    fn default() -> Self {
        Self::MB
    }
}
```

## Impact

A config like `{"diskSpaceLimit": "25"}` with no unit specified would be interpreted as:
- **Java**: 25 KB (25,600 bytes)
- **Rust**: 25 MB (26,214,400 bytes)

Customers migrating from the Java LogManager to the Rust version without changing their config would silently get a much larger disk space allowance.

## Suggested fix

Change the Rust default to `KB` to match Java:

```rust
impl Default for DiskSpaceLimitUnit {
    fn default() -> Self {
        Self::KB
    }
}
```

Introduced in PR #239.

paabbs:
## Fix

Matches Java LogManager disk space limit behavior for backward compatibility during migration.

### Changes

1. **diskSpaceLimitUnit default**: `MB` → `KB` (matching Java's `if (StringUtils.isEmpty(diskSpaceLimitUnit)) diskSpaceLimitUnit = "KB"`)
2. **diskSpaceLimit type**: `String` with default `"25"` → `Option<String>` with default `None`. When omitted, disk management is disabled — matching Java where `null` diskSpaceLimit causes `DiskSpaceManagementService.freeDiskSpace()` to return early.
3. **parse_disk_space_limit**: Updated to accept `Option<&str>`, returns `Result<Option<u64>>`

### Java behavior verified

| Config | Java | Rust (before) | Rust (after) |
|---|---|---|---|
| `{"diskSpaceLimit": "25"}` (no unit) | 25 KB | 25 MB | 25 KB ✅ |
| No `diskSpaceLimit` field | Disk mgmt disabled | 25 MB limit | Disabled ✅ |
| `{"diskSpaceLimit": "50", "diskSpaceLimitUnit": "MB"}` | 50 MB | 50 MB | 50 MB ✅ |

### Tests
111 tests, 0 failures

Closes #243

### DIFF
--- a/src/config/schema.rs
+++ b/src/config/schema.rs
@@ -7,7 +7,6 @@ use serde::{Deserialize, Serialize};
 use std::fmt;
 
 pub const DEFAULT_UPLOAD_INTERVAL_SEC: u64 = 300;
-pub const DEFAULT_DISK_SPACE_LIMIT: u64 = 25;
 
 /// Custom error type for configuration operations.
 #[derive(Debug)]
@@ -54,8 +53,8 @@ pub enum LogLevel {
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Default)]
 pub enum DiskSpaceLimitUnit {
-    KB,
     #[default]
+    KB,
     MB,
     GB,
 }
@@ -78,18 +77,14 @@ pub struct LogSourceConfig {
     pub log_file_regex: String,
     #[serde(default)]
     pub minimum_log_level: LogLevel,
-    #[serde(default = "default_disk_limit")]
-    pub disk_space_limit: String,
+    #[serde(default)]
+    pub disk_space_limit: Option<String>,
     #[serde(default)]
     pub disk_space_limit_unit: DiskSpaceLimitUnit,
     #[serde(default)]
     pub delete_log_file_after_cloud_upload: bool,
     pub multi_line_start_pattern: Option<String>,
     pub upload_interval_sec: Option<u64>,
-}
-
-fn default_disk_limit() -> String {
-    DEFAULT_DISK_SPACE_LIMIT.to_string()
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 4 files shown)


src/config/schema.rs
```diff
@@ -143,8 +138,8 @@ mod tests {
     fn test_component_defaults() {
         let json = r#"{"componentName":"test","logFileDirectoryPath":"/tmp","logFileRegex":".*"}"#;
         let comp: ComponentLogConfig = serde_json::from_str(json).unwrap();
-        assert_eq!(comp.source.disk_space_limit, "25");
-        assert_eq!(comp.source.disk_space_limit_unit, DiskSpaceLimitUnit::MB);
+        assert_eq!(comp.source.disk_space_limit, None);
```
limingyaoo, line 141: nit: consider adding a test for "disk_space_limit": ""

limingyaoo (approved):

src/config/schema.rs
paabbs: Will add that in the follow up fix

paabbs: Addressed in the latest push for https://github.com/aws-greengrass/aws-greengrass-log-manager/pull/246

alter-mage (approved):

PR #246, opened 2026-04-24
fix: warn instead of reject when log directory missing at config time (#244)

### LINKED ISSUE #243: Bug: diskSpaceLimitUnit defaults to MB in Rust, but KB in Java

### LINKED ISSUE #244: Bug: validate_directory rejects configs when log directory doesn't exist yet
## Summary

The Rust v2 LogManager validates that log directories exist at config load time. This can cause valid configs to be rejected if LogManager starts before the component whose logs it is configured to tail.

## Rust behavior

`validate_directory` checks `is_dir()` at startup and returns an error if the directory doesn't exist:

https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/rust/v2/src/config/mod.rs#L55-L62

```rust
fn validate_directory(path: impl AsRef<Path>) -> Result<(), ConfigError> {
    let p = path.as_ref();
    if !p.is_dir() {
        return Err(ConfigError::Validation(format!(
            "Directory does not exist: {}",
            p.display()
        )));
    }
    Ok(())
}
```

This is called during `validate_config()` for every component and system log source.

## Java behavior

The Java LogManager does not validate directory existence at config load time. Instead, it handles missing directories gracefully at scan time — `LogFileGroup.create()` catches `InvalidLogGroupException` and logs a debug message:

https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/main/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java#L350-L355

```java
} catch (InvalidLogGroupException e) {
    logger.atDebug().cause(e).log("Unable to read the directory");
}
```

## Impact

On a Greengrass device, components can start in any order. If LogManager starts before a component that creates its own log directory, the Rust version would reject the config and fail to monitor that component's logs — even though the directory will exist shortly.

## Suggested fix

Move directory existence checks from config validation to scan time, matching Java's behavior. Log a warning when a directory is missing and retry on the next scan cycle.

Introduced in PR #239.

paabbs:
## Fix

Log directories may not exist at config load time if LogManager starts before the component that creates its log directory. GG components can start in any order.

### Change

`validate_directory` now logs a warning and continues instead of returning an error. Directory existence is checked at scan time.

### Stacked on
- #245 (fix #243 — diskSpaceLimit defaults)

### Tests
111 tests, 0 failures

Closes #244

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -58,13 +58,13 @@ pub fn validate_config(config: &LogManagerConfig) -> Result<(), ConfigError> {
     Ok(())
 }
 
+// Returns Result for forward compatibility — may add permission checks later.
 fn validate_directory(path: impl AsRef<Path>) -> Result<(), ConfigError> {
     let p = path.as_ref();
     if !p.is_dir() {
-        return Err(ConfigError::Validation(format!(
-            "Directory does not exist: {}",
-            p.display()
-        )));
+        // Warn but don't reject — directory may not exist yet if LogManager starts
+        // before the component that creates it.
+        tracing::warn!(directory = %p.display(), "Configured log directory does not exist yet");
     }
     Ok(())
 }
@@ -90,15 +90,21 @@ fn validate_disk_limit(limit: &str) -> Result<(), ConfigError> {
 }
 
 /// Parse diskSpaceLimit string to u64. Returns None if limit is not configured.
+/// Rejects 0 (same as validate_disk_limit).
 #[must_use = "parse result must be handled"]
 pub fn parse_disk_space_limit(limit: Option<&str>) -> Result<Option<u64>, ConfigError> {
-    match limit {
-        Some(l) => l
-            .parse()
-            .map(Some)
-            .map_err(|_| ConfigError::Parse(format!("Invalid diskSpaceLimit: {l}"))),
-        None => Ok(None),
-    }
+    limit
+        .filter(|l| !l.is_empty())
+        .map(|l| {
+            let parsed: u64 = l
+                .parse()
+                .map_err(|_| ConfigError::Parse(format!("Invalid diskSpaceLimit: {l}")))?;
+            if parsed == 0 {
+                return Err(ConfigError::Parse("diskSpaceLimit must be positive".into()));
+            }
+            Ok(parsed)
+        })
+        .transpose()
 }
 
 /// Derive log group name: /aws/greengrass/{componentType}/{region}/{componentName}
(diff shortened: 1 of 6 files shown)


src/config/mod.rs
```diff
@@ -61,10 +61,9 @@ pub fn validate_config(config: &LogManagerConfig) -> Result<(), ConfigError> {
 fn validate_directory(path: impl AsRef<Path>) -> Result<(), ConfigError> {
     let p = path.as_ref();
     if !p.is_dir() {
-        return Err(ConfigError::Validation(format!(
-            "Directory does not exist: {}",
-            p.display()
-        )));
+        // Warn but don't reject — directory may not exist yet if LogManager starts
+        // before the component that creates it.
+        tracing::warn!(directory = %p.display(), "Configured directory does not exist yet, will retry at scan time");
```
limingyaoo: We're saying "will retry at scan time" but that logic seems to be missing in the current change.

alter-mage (approved):

limingyaoo (approved):

PR #247, opened 2026-04-29
feat(uploader): CloudWatch Logs client with retry

utsavvyas:
**Issue #, if available:**

N/A — CloudWatch Logs client for Rust v2 LogManager

**Description of changes:**

CloudWatch Logs API client with typed error handling, SDK-managed retry, and log group/stream lifecycle.

- `uploader/cw_client.rs` — `CwLogsClient` wrapping `aws-sdk-cloudwatchlogs::Client`. Typed error matching on all SDK exceptions (Java parity): `DataAlreadyAcceptedException` → success, `ResourceNotFoundException` → create group/stream, `UnrecognizedClientException` → auth error, `InvalidParameterException` → non-retriable. Auto-creates log groups and streams with idempotent `ResourceAlreadyExistsException` handling. Error type uses `thiserror` for `Display`/`Error` impls.
- `uploader/mod.rs` — `SealedBatch` type shared with future batcher PR. `format_log_stream_name` and `is_different_date` utilities using `time` crate for readable date handling.
- `Cargo.toml` — `aws-sdk-cloudwatchlogs`, `aws-config`, `time`, `thiserror`. Feature-gated behind `aws-sdk`. Single-thread tokio runtime (`current_thread`).
- `tests/sdk_retry_test.rs` — 4 mock HTTP integration tests: SDK retry on 429, non-retriable 400 not retried, DataAlreadyAccepted, ResourceNotFound.

**Key design decisions:**

- **AWS SDK over raw HTTP** — We chose `aws-sdk-cloudwatchlogs` over the raw HTTP + SigV4 approach used by SystemLogForwarder. The SDK provides typed request builders, typed error variants, automatic credential refresh (TES), and built-in retry — reducing custom code. The trade-off is ~100KB additional binary size (1.5MB vs ~500KB). The CW client is feature-gated behind `aws-sdk` in Cargo.toml, so it can be swapped to a raw HTTP implementation by creating a shared CloudWatch HTTP client crate between this component and SystemLogForwarder, without changing any other code. Estimated swap effort: ~3 days, single PR touching only `cw_client.rs` + `Cargo.toml`.
- **No custom retry** — SDK's built-in retry (`max_attempts=5`, exponential backoff) handles throttling, transient errors, and connection failures. Matches Java's approach of no per-call retry.
- **No per-call rate limiting** — Periodic upload interval (default 300s) is the natural throttle, matching Java behavior.
- **No sequence tokens** — Deprecated since late 2023. We create new streams daily, so tokens are never needed.
- **Typed SDK exceptions** — All error handling uses SDK typed variants, not string matching. Justifies the SDK dependency.
- **Single-thread runtime** — `current_thread` tokio, matching Java's single-threaded upload model.
- **`thiserror` for error types** — `CwUploadError` derives `thiserror::Error` for `Display` and `std::error::Error` impls. This enables `?` propagation and `{}` formatting in logs. Not strictly needed today (callers match on variants directly), but prepares for the upload orchestrator PR where errors will flow through `Result<_, Box<dyn Error>>` chains.

The CW client is the upload layer — it sends batched log events to CloudWatch Logs. Every subsequent PR (batcher, orchestrator, disk management) depends on this. The typed error handling and SDK retry provide robust upload behavior without custom retry code.

- 132 tests total (78 unit + 54 integration)
- 4 mock HTTP integration tests verifying SDK retry behavior ✅
- `cargo clippy` ✅ (clean)
- `cargo zigbuild --target aarch64-unknown-linux-musl` ✅

This PR depends on the scanner module (merged in PR #240) for the `LogEvent` type. The `SealedBatch` struct in `mod.rs` will be consumed by the batcher PR (in progress by @paabbs). The upload orchestrator (wiring scanner → batcher → CW client) is deferred until checkpoint PR #242 merges.

**Checklist:**

### DIFF
--- a/src/uploader/cw_client.rs
+++ b/src/uploader/cw_client.rs
@@ -2,3 +2,431 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! CloudWatch Logs API client
+
+use super::SealedBatch;
+use aws_sdk_cloudwatchlogs::{
+    error::SdkError, operation::put_log_events::PutLogEventsError, types::InputLogEvent, Client,
+};
+use std::collections::HashSet;
+
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum CwUploadError {
+    #[error("authentication error")]
+    AuthError,
+    #[error("retriable: {0}")]
+    Retriable(String),
+    #[error("{0}")]
+    Other(String),
+}
+
+pub(crate) struct CwLogsClient {
+    client: Client,
+    config: aws_sdk_cloudwatchlogs::Config,
+    created_groups: HashSet<String>,
+    created_streams: HashSet<String>,
+}
+
+impl CwLogsClient {
+    pub async fn new() -> Self {
+        let retry_config = aws_config::retry::RetryConfig::standard()
+            .with_max_attempts(5);
+        let sdk_config = aws_config::defaults(aws_config::BehaviorVersion::latest())
+            .retry_config(retry_config)
+            .load()
+            .await;
+        let config = aws_sdk_cloudwatchlogs::Config::new(&sdk_config);
+        let client = Client::from_conf(config.clone());
+        Self {
+            client,
+            config,
+            created_groups: HashSet::new(),
+            created_streams: HashSet::new(),
+        }
+    }
+
+    #[cfg(test)]
+    fn new_with_client(client: Client, config: aws_sdk_cloudwatchlogs::Config) -> Self {
+        Self {
+            client,
+            config,
+            created_groups: HashSet::new(),
+            created_streams: HashSet::new(),
+        }
+    }
+
+    /// Recreate the SDK client (e.g., after persistent network errors).
+    /// Caches are preserved — if a resource was deleted externally, the next
+    /// put_log_events will get ResourceNotFoundException which clears the cache.
+    pub(crate) fn recreate_client(&mut self) {
+        self.client = Client::from_conf(self.config.clone());
+    }
+
+    pub(crate) async fn upload_batch(&mut self, batch: &SealedBatch) -> Result<(), CwUploadError> {
+        if batch.events.is_empty() {
+            return Ok(());
+        }
+        self.ensure_log_group(&batch.log_group).await?;
+        self.ensure_log_stream(&batch.log_group, &batch.log_stream)
+            .await?;
+        self.put_log_events(batch).await
+    }
+
+    async fn ensure_log_group(&mut self, log_group: &str) -> Result<(), CwUploadError> {
+        if self.created_groups.contains(log_group) {
+            return Ok(());
+        }
+        match self
+            .client
+            .create_log_group()
+            .log_group_name(log_group)
+            .send()
+            .await
+        {
+            Ok(_) => {
+                tracing::info!(log_group = %log_group, "Created log group");
+                self.created_groups.insert(log_group.to_string());
+                Ok(())
+            }
+            Err(SdkError::ServiceError(e)) if is_group_exists(e.err()) => {
+                self.created_groups.insert(log_group.to_string());
+                Ok(())
+            }
+            Err(SdkError::ServiceError(e)) if is_limit_exceeded_group(e.err()) => {
+                Err(CwUploadError::Retriable("LimitExceededException on create_log_group".to_string()))
+            }
+            Err(SdkError::ServiceError(e)) if is_auth_error(e.err()) => {
+                Err(CwUploadError::AuthError)
+            }
+            Err(e) => Err(CwUploadError::Retriable(e.to_string())),
+        }
+    }
+
+    async fn ensure_log_stream(
+        &mut self,
+        log_group: &str,
+        log_stream: &str,
+    ) -> Result<(), CwUploadError> {
+        let key = format!("{}:{}", log_group, log_stream);
+        if self.created_streams.contains(&key) {
+            return Ok(());
+        }
+        match self
+            .client
+            .create_log_stream()
+            .log_group_name(log_group)
+            .log_stream_name(log_stream)
+            .send()
+            .await
+        {
+            Ok(_) => {
+                tracing::info!(log_group = %log_group, log_stream = %log_stream, "Created log stream");
+                self.created_streams.insert(key);
+                Ok(())
+            }
+            Err(SdkError::ServiceError(e)) if is_stream_exists(e.err()) => {
+                self.created_streams.insert(key);
+                Ok(())
+            }
+            Err(SdkError::ServiceError(e)) if is_limit_exceeded_stream(e.err()) => {
+                Err(CwUploadError::Retriable("LimitExceededException on create_log_stream".to_string()))
+            }
+            Err(SdkError::ServiceError(e)) if is_auth_error(e.err()) => {
+                Err(CwUploadError::AuthError)
+            }
+            Err(e) => Err(CwUploadError::Retriable(e.to_string())),
+        }
+    }
+
+    async fn put_log_events(&mut self, batch: &SealedBatch) -> Result<(), CwUploadError> {
+        let events: Result<Vec<InputLogEvent>, _> = batch
+            .events
+            .iter()
+            .map(|e| {
+                InputLogEvent::builder()
+                    .timestamp(e.timestamp)
+                    .message(&e.message)
+                    .build()
+            })
+            .collect();
+        let events = events.map_err(|e| CwUploadError::Other(format!("Failed to build log event: {e}")))?;
+
+        // EMF metrics are auto-extracted by CloudWatch when log events contain the _aws key.
+        // No sequence tokens needed — deprecated since late 2023, and we create new streams daily.
+        let req = self
+            .client
+            .put_log_events()
+            .log_group_name(&batch.log_group)
+            .log_stream_name(&batch.log_stream)
+            .set_log_events(Some(events));
+
+        match req.send().await {
+            Ok(output) => {
+                if let Some(rejected) = output.rejected_log_events_info() {
+                    tracing::error!(log_group = %batch.log_group, "Log events rejected by CloudWatch (data loss): {:?}", rejected);
+                }
+                tracing::debug!(log_group = %batch.log_group, log_stream = %batch.log_stream, event_count = batch.events.len(), "Upload successful");
+                Ok(())
+            }
+            Err(SdkError::ServiceError(e)) => match e.into_err() {
+                PutLogEventsError::DataAlreadyAcceptedException(_) => {
+                    tracing::debug!(log_group = %batch.log_group, "Data already accepted");
+                    Ok(())
+                }
+                PutLogEventsError::UnrecognizedClientException(_) => {
+                    Err(CwUploadError::AuthError)
+                }
+                PutLogEventsError::ServiceUnavailableException(_) => {
+                    Err(CwUploadError::Other("ServiceUnavailable after SDK retries exhausted".to_string()))
+                }
+                PutLogEventsError::ResourceNotFoundException(_) => {
+                    // Clear cache so next retry re-creates the group/stream
+                    self.created_groups.remove(&batch.log_group);
+                    let key = format!("{}:{}", batch.log_group, batch.log_stream);
+                    self.created_streams.remove(&key);
+                    Err(CwUploadError::Retriable("ResourceNotFound".to_string()))
+                }
+                PutLogEventsError::InvalidParameterException(ex) => {
+                    Err(CwUploadError::Other(ex.to_string()))
+                }
+                other => {
+                    use aws_sdk_cloudwatchlogs::error::ProvideErrorMetadata;
+                    if other.code() == Some("ThrottlingException") {
+                        Err(CwUploadError::Other("Throttled after SDK retries exhausted".to_string()))
+                    } else {
+                        Err(CwUploadError::Other(other.to_string()))
+                    }
+                }
+            },
+            Err(e) => Err(CwUploadError::Retriable(e.to_string())),
+        }
+    }
+
+    #[cfg(test)]
+    pub(crate) fn mark_group_created(&mut self, group: &str) {
+        self.created_groups.insert(group.to_string());
+    }
+
+    #[cfg(test)]
+    pub(crate) fn mark_stream_created(&mut self, group: &str, stream: &str) {
+        self.created_streams.insert(format!("{}:{}", group, stream));
+    }
+
+    #[cfg(test)]
+    pub(crate) fn created_groups(&self) -> &HashSet<String> {
+        &self.created_groups
+    }
+
+    #[cfg(test)]
+    pub(crate) fn created_streams(&self) -> &HashSet<String> {
+        &self.created_streams
+    }
+}
+
+fn is_group_exists(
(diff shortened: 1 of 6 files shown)


src/uploader/cw_client.rs
```diff
@@ -2,3 +2,333 @@
...
+                    .build()
+            })
+            .collect();
+        let events = events.map_err(|e| CwUploadError::Other(format!("Failed to build log event: {e}")))?;
+
+        // EMF metrics are auto-extracted by CloudWatch when log events contain the _aws key.
+        // No sequence tokens needed — deprecated since late 2023, and we create new streams daily.
+        let req = self
+            .client
+            .put_log_events()
```
paabbs, line 157: Do we want to add the EMF specific headers in the orchestrator hookup because the current flow does not have those, it should be sth like-

**Fix:** Add the header via `put_log_events().customize()`:
```rust
let req = self.client.put_log_events()
    .log_group_name(&batch.log_group)
    .log_stream_name(&batch.log_stream)
    .set_log_events(Some(events))
    .customize()
    .mutate_request(|req| {
        req.headers_mut().insert("x-amzn-logs-format", "json/emf".parse().unwrap());
    });
```

src/uploader/cw_client.rs
```diff
@@ -2,3 +2,333 @@
...
+                    tracing::debug!(log_group = %batch.log_group, "Data already accepted");
+                    Ok(())
+                }
+                PutLogEventsError::UnrecognizedClientException(_) => {
+                    Err(CwUploadError::AuthError)
+                }
+                PutLogEventsError::ServiceUnavailableException(_) => {
+                    Err(CwUploadError::Other("ServiceUnavailable after SDK retries exhausted".to_string()))
+                }
+                PutLogEventsError::ResourceNotFoundException(_) => {
```
paabbs, line 181: When `put_log_events` returns `ResourceNotFoundException`, the Rust code returns `CwUploadError::Retriable("ResourceNotFound")`. But the `created_groups` and `created_streams` caches still contain the deleted group/stream. When the caller retries by calling `upload_batch` again, `ensure_log_group`/`ensure_log_stream` will see the cached entries and **skip creation**, causing the retry to fail with the same `ResourceNotFoundException` in an infinite loop.

Java handles this differently — it creates the group/stream inline and retries immediately within the same method.

**Fix:** Clear the relevant cache entries before returning `Retriable`:
```rust
PutLogEventsError::ResourceNotFoundException(_) => {
    self.created_groups.remove(&batch.log_group);
    let key = format!("{}:{}", batch.log_group, batch.log_stream);
    self.created_streams.remove(&key);
    Err(CwUploadError::Retriable("ResourceNotFound".to_string()))
}
```

src/uploader/cw_client.rs
utsavvyas: They do exist by default within SDK calls.

paabbs: Ah, I see

src/uploader/cw_client.rs
utsavvyas: Ideally this scenario can happen except if someone actually delete the log group manully between our check of log group exists and when we upload but adding that update in the codebase.

paabbs (approved):

tests/sdk_retry_test.rs
```diff
@@ -0,0 +1,183 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Integration tests verifying AWS SDK retry behavior and our CwLogsClient error handling.
```
aws-kevinrickard, line 6: This file breaks `cargo check --no-default-features --all-targets`. It uses `aws_sdk_cloudwatchlogs` and `aws_smithy_http_client` unconditionally, so when the `aws-sdk` feature is off, compilation fails with 10 errors.

Add `#![cfg(feature = "aws-sdk")]` at the top of the file, and consider making the aws-smithy `[dev-dependencies]` optional too. Without this, the PR's stated "swap to raw HTTP impl by touching only `cw_client.rs` + `Cargo.toml`" design doesn't hold.

src/uploader/cw_client.rs
```diff
@@ -2,3 +2,416 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum CwUploadError {
+    #[error("authentication error")]
+    AuthError,
+    #[error("retriable: {0}")]
+    Retriable(String),
+    #[error("{0}")]
+    Other(String),
+}
```
aws-kevinrickard, line 20: Might be good to document the caller contract on `CwUploadError` — the Rust-default reading of `Retriable` is "retry a bounded number of times," but this is a remote IoT device that needs to self-heal without operator intervention, so the actual contract is different. Suggested:

```rust
/// Errors from the CloudWatch upload path.
///
/// Callers on a remote IoT device should treat these as follows:
/// - `AuthError`: refresh credentials via TES, then retry. The retry is
///   what actually recovers; just retrying without refresh will always fail.
/// - `Retriable`: retry indefinitely with exponential backoff (cap the delay
///   at several minutes). Do not give up — the device cannot be manually
///   touched per-device when the underlying cloud condition clears.
/// - `Other`: request-level failure that will not self-heal by retrying the
///   same request (e.g. malformed payload). Log, drop this batch, continue
///   with the next batch.
```

src/uploader/cw_client.rs
```diff
@@ -2,3 +2,416 @@
...
+                Ok(())
+            }
+            Err(SdkError::ServiceError(e)) if is_group_exists(e.err()) => {
+                self.created_groups.insert(log_group.to_string());
+                Ok(())
+            }
+            Err(SdkError::ServiceError(e)) if is_limit_exceeded_group(e.err()) => {
+                Err(CwUploadError::Retriable("LimitExceededException on create_log_group".to_string()))
+            }
+            Err(e) => Err(CwUploadError::Retriable(e.to_string())),
```
aws-kevinrickard, line 100: This catch-all classifies everything — including auth failures — as `Retriable`, which is inconsistent with `put_log_events` which correctly breaks `UnrecognizedClientException` out as `AuthError`. The create-group path should match:

```rust
Err(SdkError::ServiceError(e)) => {
   use aws_sdk_cloudwatchlogs::error::ProvideErrorMetadata;
   match e.err().code() {
       Some("UnrecognizedClientException")
       | Some("ExpiredTokenException")
       | Some("AccessDeniedException") => Err(CwUploadError::AuthError),
  => Err(CwUploadError::Retriable(e.err().tostring())),
   }
}
Err(e) => Err(CwUploadError::Retriable(e.to_string())),
```
TES rotates credentials automatically and the AWS SDK re-polls its credential provider on each request, so in the happy path the component never sees an auth error. When it does, it's a signal of something unusual — clock skew, IAM drift, a TES hiccup, or a stale credential cache — and it's worth distinguishing from generic `Retriable` so the orchestrator can:

1. Emit a dedicated fleet metric for auth failures (different operational signal from "CW throttled me")
2. Call `recreate_client()` to force the SDK to drop any cached credential before the next attempt (this PR already sets that up)
3. Apply a longer backoff — auth errors typically take minutes-to-hours to clear, not seconds

Without the separation all three behaviors collapse into generic retry, and fleet dashboards lose the signal.

src/uploader/cw_client.rs
```diff
@@ -2,3 +2,416 @@
...
+                Ok(())
+            }
+            Err(SdkError::ServiceError(e)) if is_stream_exists(e.err()) => {
+                self.created_streams.insert(key);
+                Ok(())
+            }
+            Err(SdkError::ServiceError(e)) if is_limit_exceeded_stream(e.err()) => {
+                Err(CwUploadError::Retriable("LimitExceededException on create_log_stream".to_string()))
+            }
+            Err(e) => Err(CwUploadError::Retriable(e.to_string())),
```
aws-kevinrickard, line 136: Same as the ensure_log_group catch-all — CreateLogStream auth failures get swept into Retriable rather than surfacing as AuthError like put_log_events does. Apply the same fix: match on UnrecognizedClientException / ExpiredTokenException / AccessDeniedException from the SDK error code and return AuthError; everything else stays Retriable.

TES + the SDK's credential provider handle rotation automatically in the happy path, so AuthError should be rare. When it does fire, separating it lets the orchestrator drop the stale credential cache via recreate_client(), emit a distinct fleet metric, and/or back off longer than it would for a transient throttle.

src/uploader/cw_client.rs
```diff
@@ -2,3 +2,416 @@
...
+            client,
+            config,
+            created_groups: HashSet::new(),
+            created_streams: HashSet::new(),
+        }
+    }
+
+    /// Recreate the AWS SDK client (e.g., after network errors)
+    /// Recreate the SDK client (e.g., after persistent network errors).
+    /// Caches are preserved — if a resource was deleted externally, the next
```
aws-kevinrickard, line 58: Nit: two stacked doc comments (`/// Recreate the AWS SDK client (e.g., after network errors)` followed immediately by `/// Recreate the SDK client (e.g., after persistent network errors). ...`). Looks like a copy-paste leftover — keep the more informative second one.

src/uploader/mod.rs
```diff
@@ -1,8 +1,132 @@
...
+/// Check if a timestamp (epoch millis) falls on a different UTC date.
+/// Used to detect when a new log stream should be created at midnight.
+#[allow(dead_code, reason = "used by upload orchestrator in follow-up PR")]
+pub(crate) fn is_different_date(
+    timestamp_ms: i64,
+    stream_year: i32,
+    stream_month: u32,
+    stream_day: u32,
+) -> bool {
+    if timestamp_ms < 0 {
```
aws-kevinrickard: Returning `true` unconditionally for `timestamp_ms < 0` is a weird special case with no contract behind it, and I think it should just go away.

Walking the proof chain:

1. **Nothing upstream produces a negative timestamp.** `scanner::reader::extract_timestamp` constrains parsed values to `1_000_000_000_000..=4_102_444_800_000` (≈Sept 2001 – Jan 2100); anything else returns `None` and the caller substitutes `default_timestamp`. `scanner::multiline` uses `NO_TIMESTAMP: i64 = 0` — zero, not negative — as its sentinel for "unparsed," and falls back to `SystemTime::now()` otherwise.
2. **Java log manager has no such guard.** `CloudWatchAttemptLogsProcessor` feeds parsed timestamps straight into `DATE_FORMATTER.format(new Date(...))` and `Instant.ofEpochMilli(...)`. The only temporal filter is the CloudWatch 14-day cutoff, which a pre-1970 timestamp would fail — so it'd be dropped as "too old" long before the stream-rotation path, not rotated into a new stream.
3. **CloudWatch rejects pre-1970 at the API.** `PutLogEvents` rejects anything older than 14 days; a negative epoch is >55 years old and can never land regardless of what we decide here.
4. `OffsetDateTime::from_unix_timestamp` accepts negative values fine, and the existing `Err(_) => true` fallback already handles genuinely malformed inputs uniformly — no pre-check needed.

So the branch is unreachable today and becomes a latent behavioral divergence from Java (and from the normal path) if someone later hands this function a sentinel or a garbage value. Recommend dropping the guard and letting the normal path handle it; if you want an explicit "unknown timestamp ⇒ force new stream" contract, the more honest thing is to introduce a named sentinel (`NO_TIMESTAMP` already exists) and check for *that*, not `< 0`.

One question worth pinning down before this lands: when the follow-up upload orchestrator wires this up, what's the source of `timestamp_ms`? If it's `LogEvent::timestamp` straight from the scanner we're fine as discussed. If it's something that could deserialize an untrusted JSON field into an `i64`, we should decide explicitly whether to mirror Java (let the 14-day filter drop it) or diverge — but the current guard isn't the right answer either way.

aws-kevinrickard (commented):
Still reviewing, here's some comments so far

tests/sdk_retry_test.rs
utsavvyas: Down side of not adding CI actions. I will fix it. I do have launch manager in progress

src/uploader/cw_client.rs
utsavvyas: The AWS SDK for Rust uses a LazyCache identity provider that auto-refreshes credentials from TES before expiry. Each API call checks the cache and re-fetches if expired — no manual refresh needed from the caller.

Auth failures that reach our code mean the SDK already tried with its best available credentials and failed. The underlying causes (TES down, IAM policy change, clock skew) are external conditions that won't resolve by retrying the same request immediately. Recovery happens naturally on the next periodic upload cycle (default 300s) when the external condition clears.

We've separated AuthError from Retriable so the caller can skip the batch immediately (no wasted retries) and log it distinctly. But the recovery action is the same as Java — wait for next cycle. Adding IoT-specific retry semantics to the doc would imply the caller needs to do something different per variant beyond "skip vs retry," which it doesn't today. Happy to add the doc comment when the orchestrator PR introduces metric emission and the distinction becomes actionable.

src/uploader/cw_client.rs
utsavvyas: Fixed — added is_auth_error() helper that matches UnrecognizedClientException, ExpiredTokenException, and AccessDeniedException via ProvideErrorMetadata::code(). Returns AuthError so the caller doesn't waste retries on a request that will always fail with the same credentials.

src/uploader/mod.rs
utsavvyas: Removed. The scanner constrains timestamps to valid ranges (1_000_000_000_000..=4_102_444_800_000) and falls back to SystemTime::now() for unparsed lines. Negative timestamps are unreachable from the current scanner. The Err(_) => true fallback in OffsetDateTime::from_unix_timestamp already handles any truly invalid input uniformly. Existing test test_is_different_date_negative_timestamp confirms small negatives resolve correctly via integer division (e.g., -1/1000 = 0 → epoch → different from any modern date).

src/uploader/cw_client.rs
utsavvyas: Fixed — removed the first line, kept the more informative second one.

src/uploader/cw_client.rs
utsavvyas: I would refer to write what in the code. It can be infer from code regardless.

src/uploader/cw_client.rs
```diff
@@ -2,3 +2,431 @@
...
+            .client
+            .put_log_events()
+            .log_group_name(&batch.log_group)
+            .log_stream_name(&batch.log_stream)
+            .set_log_events(Some(events));
+
+        match req.send().await {
+            Ok(output) => {
+                if let Some(rejected) = output.rejected_log_events_info() {
+                    tracing::warn!(log_group = %batch.log_group, "Some log events were rejected: {:?}", rejected);
```
aws-kevinrickard: This means data loss happened, we should at a minimum report at error level (as java does).

Even better would be having some feedback such as emitting a metric or returning failure information to the caller to decide how to proceed when this happens, but I'm not sure what that might look like. That would be considered an improvement over java.

utsavvyas: This failure can happen due to client side errors which means it is not retryable. I did make it error but nothing which client can do with this failure.

aws-kevinrickard: The concern isn't about the client doing something, it's about the customer having visibility of what could be a problem for them

utsavvyas: I do get it and that is why bumped up to the error.

aws-kevinrickard (approved):

PR #248, opened 2026-04-30
feat(disk): disk space management — delete oldest processed files whe…

utsavvyas:
**Issue #, if available:**

N/A — Disk space management for Rust v2 LogManager

**Description of changes:**

Disk space management module — deletes oldest already-processed log files when directory size exceeds the configured `diskSpaceLimit`.

- `disk/mod.rs` — `free_disk_space(dir, limit_bytes, processed_files)` function. Calculates directory size, sorts processed files by modified time (oldest first), deletes until under limit. Returns list of deleted paths.
- `dir_size()` helper — calculates total size of regular files in a directory. Uses `file_type()` (lstat) to skip symlinks. Logs warning on read errors instead of silently failing.
- 8 unit tests covering: no-op under limit, oldest-first deletion, only processed files touched, missing files, empty list, all files over limit, zero limit, permission-denied skip.

**Key design decisions:**

- **"Dumb deleter" pattern** — This function trusts the `processed_files` list it receives. Safety logic (active file exclusion, upload verification, rotation awareness) is the orchestrator's responsibility, matching Java's architecture where `DiskSpaceManagementService.freeDiskSpace()` operates on a pre-filtered `LogFileGroup`.
- **Directory-level size check** — `dir_size()` counts all regular files in the directory (top-level only). Java scopes to pattern-matched files via `LogFileGroup.totalSizeInBytes()`. For GG components with dedicated log directories, these are equivalent. If shared directories become a concern, the orchestrator can pass a scoped file list for size calculation.
- **Graceful error handling** — Permission-denied on delete → log warning, skip, continue. Missing files → filtered out during metadata collection. Unreadable directory → log warning, return empty.
- **No hardlinks** — Java uses hardlinks to prevent TOCTOU races during log rotation. We skip this complexity since GG components use numbered rotation suffixes (new names), not in-place overwrites.

Disk management prevents log accumulation from filling device storage during WAN outages or high-volume logging. Without it, a device offline for days would run out of disk space. This matches the Java LogManager's `diskSpaceLimit` config behavior.

- 8 unit tests (inline in disk/mod.rs) including Unix-specific permission test
- `cargo clippy` ✅ (clean)
- `cargo test` ✅ (all pass)

This module is not wired into the main loop yet (`#[allow(dead_code)]`). The upload orchestrator (future PR) will call `free_disk_space` after each successful upload cycle, passing only files confirmed uploaded via the checkpoint system (PR #242, merged).

**Checklist:**

### DIFF
--- a/src/disk/mod.rs
+++ b/src/disk/mod.rs
@@ -1,4 +1,263 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Disk space management - per-log-group limits, file cleanup
+//! Disk space management — per-component log directory size limits.
+//! Deletes oldest already-processed (uploaded) files when total exceeds diskSpaceLimit.
+
+use regex::Regex;
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::time::SystemTime;
+
+/// Calculate total size of regular files matching `pattern` in `dir` (top-level only).
+/// Skips symlinks and subdirectories.
+fn dir_size(dir: &Path, pattern: &Regex) -> u64 {
+    match fs::read_dir(dir) {
+        Ok(entries) => entries
+            .filter_map(|e| e.ok())
+            .filter(|e| e.file_type().is_ok_and(|ft| ft.is_file()))
+            .filter(|e| pattern.is_match(&e.file_name().to_string_lossy()))
+            .filter_map(|e| e.metadata().ok())
+            .map(|m| m.len())
+            .sum(),
+        Err(e) => {
+            tracing::warn!(dir = %dir.display(), error = %e, "Cannot read directory for disk management");
+            0
+        }
+    }
+}
+
+/// Free disk space by deleting oldest processed files until directory size ≤ limit.
+///
+/// Only deletes files in `processed_files` — caller must ensure these are safe to delete
+/// (fully uploaded, not actively written to). This function trusts the list it receives.
+///
+/// Returns paths of successfully deleted files.
+pub(crate) fn free_disk_space(
+    dir: &Path,
+    pattern: &Regex,
+    limit_bytes: u64,
+    processed_files: &[PathBuf],
+) -> Vec<PathBuf> {
+    if processed_files.is_empty() {
+        return Vec::new();
+    }
+
+    let total_size = dir_size(dir, pattern);
+
+    if total_size <= limit_bytes {
+        return Vec::new();
+    }
+
+    // Cache metadata upfront to avoid double reads (once for sort, once for size).
+    let mut file_info: Vec<_> = processed_files
+        .iter()
+        .filter_map(|p| {
+            fs::metadata(p)
+                .ok()
+                .map(|m| (p.clone(), m.len(), m.modified().unwrap_or(SystemTime::UNIX_EPOCH)))
+        })
+        .collect();
+    file_info.sort_by_key(|(_, _, mtime)| *mtime); // oldest first
+
+    let mut bytes_to_free = total_size.saturating_sub(limit_bytes);
+    let mut deleted = Vec::with_capacity(file_info.len());
+    let mut actually_freed: u64 = 0;
+
+    for (path, size, _) in file_info {
+        if bytes_to_free == 0 {
+            break;
+        }
+        if let Err(e) = fs::remove_file(&path) {
+            tracing::warn!(path = %path.display(), error = %e, "Failed to delete processed file, skipping");
+            continue;
+        }
+        tracing::info!(path = %path.display(), freed_bytes = size, "Deleted processed log file");
+        deleted.push(path);
+        actually_freed += size;
+        bytes_to_free = bytes_to_free.saturating_sub(size);
+    }
+
+    if !deleted.is_empty() {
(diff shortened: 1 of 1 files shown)


src/disk/mod.rs
```diff
@@ -1,4 +1,236 @@
...
+//! Disk space management — per-component log directory size limits.
+//! Deletes oldest already-processed (uploaded) files when total exceeds diskSpaceLimit.
+
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::time::SystemTime;
+
+/// Calculate total size of regular files in `dir` (top-level only, not recursive).
+/// Skips symlinks and subdirectories. Matches Java LogManager's flat directory scanning.
+fn dir_size(dir: &Path) -> u64 {
```
alter-mage: Java's `totalSizeInBytes()` sums only the files it's tracking (the `logFiles` list), whereas `dir_size()` here sums all files in the directory. If there are untracked files in the log directory, this implementation will see a larger total and delete more processed files than the Java version would for the same limit. Is this divergence intentional?

src/disk/mod.rs
```diff
@@ -1,4 +1,236 @@
...
+    }
+
+    if !deleted.is_empty() {
+        let remaining = dir_size(dir);
+        tracing::info!(
+            total_freed = total_size.saturating_sub(remaining),
+            files_deleted = deleted.len(),
+            "Disk space freed"
+        );
+    }
```
alter-mage, line 83: The post-deletion `dir_size()` call is only used for the log message — could we avoid the extra directory scan by tracking freed bytes directly from the deleted files' sizes instead?

```rust
let mut actually_freed: u64 = 0;
// in the loop, on successful delete:
actually_freed += size;

// then in the log:
tracing::info!(total_freed = actually_freed, files_deleted = deleted.len(), "Disk space freed");
```

src/disk/mod.rs
```diff
@@ -1,4 +1,236 @@
...
+        assert!(!f1.exists());
+        assert!(!f2.exists());
+    }
+
+    #[cfg(unix)]
+    #[test]
+    fn skips_undeletable_files() {
+        use std::os::unix::fs::PermissionsExt;
+
+        let tmp = TempDir::new().unwrap();
```
alter-mage: This test may fail in CI if the process runs as root (e.g. CodeBuild) — root ignores directory permission bits, so `chmod 555` won't prevent deletion and `assert!(deleted.is_empty())` will fail. Consider adding a runtime root check to skip the test:
```rust
if std::process::Command::new("id").arg("-u").output()
    .map(|o| o.stdout.starts_with(b"0"))
    .unwrap_or(false)
{
    return;
}
```

src/disk/mod.rs
```diff
@@ -1,4 +1,236 @@
...
+/// (fully uploaded, not actively written to). This function trusts the list it receives.
+///
+/// Returns paths of successfully deleted files.
+#[allow(dead_code)] // Wired in by the upload orchestrator (future PR)
+pub(crate) fn free_disk_space(
+    dir: &Path,
+    limit_bytes: u64,
+    processed_files: &[PathBuf],
+) -> Vec<PathBuf> {
+    if processed_files.is_empty() {
```
alter-mage, line 42: `#[allow(dead_code)]` has no effect on `pub(crate)` items — the compiler never emits `dead_code` warnings for them. Safe to remove.

src/disk/mod.rs
```diff
@@ -1,4 +1,236 @@
...
+        )
+        .unwrap();
+        let t = filetime::FileTime::from_unix_time(now - secs_ago, 0);
+        filetime::set_file_mtime(path, t).unwrap();
+    }
+
+    #[test]
+    fn no_op_when_under_limit() {
+        let tmp = TempDir::new().unwrap();
+        let f1 = create_file(tmp.path(), "a.log", 100);
```
alter-mage, line 120: nit: prefer `.expect("set_mtime failed")` over `.unwrap()` in test helpers — makes failures easier to diagnose when the test panics.

src/disk/mod.rs
utsavvyas: Good catch. Will fix it.

src/disk/mod.rs
utsavvyas: Good catch. I will add that regex api here which orchestrator can pass.

utsavvyas: I dont understand why we have regex based filter but making it backward compatible regardless.

src/disk/mod.rs
utsavvyas: Fixed it.

src/disk/mod.rs
```diff
@@ -1,4 +1,263 @@
...
+    }
+
+    fn set_mtime(path: &Path, secs_ago: i64) {
+        let now = i64::try_from(
+            std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .unwrap()
+                .as_secs(),
+        )
+        .unwrap();
```
alter-mage: nit: `create_file` and `set_file_mtime` were updated to `.expect()` but these two `.unwrap()` calls in `set_mtime` were missed — inconsistent within the same helper.

alter-mage (approved):

paabbs (approved):

PR #249, opened 2026-04-30
feat(uploader): log event batcher with CW API limits and Java parity

paabbs:
**Issue #, if available:**

N/A — Log event batcher for Rust v2 LogManager

**Description of changes:**

`seal_batches()` in `uploader/batcher.rs` — batches LogEvents into SealedBatches respecting CW Logs API limits:
- 1 MB max batch payload (with 26+8 byte per-event overhead)
- 10,000 max events per batch
- 23-hour max time span within a batch (Java uses 23h leeway on CW 24h limit)
- 14-day past / 2-hour future timestamp filtering
- Oversized events (>262,110 bytes) chunked into multiple CW events with same timestamp
- minimumLogLevel filtering via JSON deserialization matching Java `tryGetStructuredLogMessage` + `checkAndAddNewLogEvent`
- Events sorted by timestamp ascending (stable sort)
- Pure function: `now_ms` injected as parameter for testability
- `SealedBatch` type and `MAX_EVENT_BYTES` constant in `uploader/mod.rs`

Scanner changes:
- Removed truncation from `reader.rs` and `multiline.rs` — batcher handles chunking (no data loss)
- Moved `MAX_EVENT_SIZE` from `scanner/mod.rs` to `uploader/batcher.rs` as `MAX_EVENT_BYTES`

Known divergence (documented in module doc):
- Dropped events do not advance checkpoint — orchestrator responsibility

Deferred to upload orchestrator:
- Per-stream date routing (midnight boundary)
- Checkpoint advancement for dropped events
- CheckpointStore encapsulation
- Eviction separation from recover_offsets

The batcher is the core batching layer that sits between the scanner (which reads log events) and the CW client (which uploads them). Without it, events cannot be grouped into valid PutLogEvents requests.

- 31 unit tests covering boundary-exact cases, chunking, filtering, overflow, and timestamp sorting
- 190 total tests passing
- `seal_batches` is a pure function (injected `now_ms`) — deterministic, no timing flakiness

N/A

**Checklist:**

### DIFF
--- a/src/scanner/multiline.rs
+++ b/src/scanner/multiline.rs
@@ -4,7 +4,6 @@
 //! Multi-line log assembly
 
 use super::reader::{extract_timestamp, LogEvent};
-use super::MAX_EVENT_SIZE;
 use regex::Regex;
 use std::sync::LazyLock;
 use std::time::SystemTime;
@@ -75,23 +74,8 @@ fn emit_buffered(buffer: &[String], first_event_timestamp: Option<i64>) -> LogEv
                 .unwrap_or(NO_TIMESTAMP)
         });
     // Join without separator to match Java's StringBuilder.append(partialLogLine) behavior.
-    let joined = buffer.join("");
-    let message = if joined.len() <= MAX_EVENT_SIZE {
-        joined
-    } else {
-        let first_line = &buffer[0][..80.min(buffer[0].len())];
-        tracing::warn!(
-            original_size = joined.len(),
-            max_size = MAX_EVENT_SIZE,
-            first_line,
-            "Truncating oversized multiline event"
-        );
-        let mut end = MAX_EVENT_SIZE;
-        while !joined.is_char_boundary(end) {
-            end -= 1;
-        }
-        joined[..end].to_string()
-    };
+    // No truncation here — the batcher handles chunking oversized events.
+    let message = buffer.join("");
     LogEvent { timestamp, message }
 }
 
@@ -244,7 +228,7 @@ mod tests {
     }
 
     #[test]
-    fn test_oversized_multiline_truncation() {
+    fn test_oversized_multiline_passed_raw() {
         let pattern = Regex::new(r"^START").unwrap();
         let large_line = "x".repeat(200_000);
         let lines = vec![
@@ -254,7 +238,9 @@ mod tests {
         ];
         let events = assemble_multiline(lines, Some(&pattern));
         assert_eq!(events.len(), 1);
-        assert!(events[0].message.len() <= MAX_EVENT_SIZE);
+        // No truncation — batcher handles chunking
+        // assemble_multiline joins without separator: "START event" (11) + 200_000 + 200_000 = 400_011
+        assert_eq!(events[0].message.len(), 400_011);
     }
 
     #[test]
--- a/src/uploader/batcher.rs
+++ b/src/uploader/batcher.rs
@@ -1,4 +1,641 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Log event batching for PutLogEvents
+//! Log event batching for PutLogEvents.
+//!
+//! Accumulates [`LogEvent`]s into [`SealedBatch`]es respecting CloudWatch Logs API limits.
+//! Matches Java `CloudWatchAttemptLogsProcessor.addNewLogEvent` behavior:
+//! - 1 MB max batch payload (including 26+8 byte per-event overhead)
+//! - 10,000 max events per batch
+//! - 23-hour max time span within a batch
+//! - 14-day age filter (events older than 14 days are dropped)
+//! - 2-hour future filter (CW rejects events >2h ahead; Java relies on server-side rejection)
+//! - Oversized events (>262,110 bytes) are chunked, not truncated
+//!
+//! # Known divergences from Java
+//!
+//! **Dropped events don't advance checkpoint.** Java returns consumed byte count for
+//! dropped events (14-day filter) so the caller advances the file offset past them.
+//! This batcher silently drops them. The orchestrator must filter old events before
+//! batching, or accept re-reading them each cycle. Tracked under Asana 1.8.
+//!
+//! **Log level filtering matches Java behavior.** JSON-structured `GreengrassLogMessage`
+//! lines are deserialized to extract the `level` field. Text-format lines pass through
+//! without level filtering (matching Java's `addNewLogEvent` path).
+
+use super::SealedBatch;
+use crate::config::LogLevel;
+use crate::scanner::LogEvent;
+
+/// CW per-event framing overhead (bytes).
+const EVENT_STORAGE_OVERHEAD: usize = 26;
+/// Timestamp field size (bytes).
+const TIMESTAMP_BYTES: usize = 8;
+/// CW PutLogEvents max payload: 1 MB.
+const MAX_BATCH_SIZE: usize = 1_048_576;
+/// CW max message bytes per event: 256KB - 8 - 26 = 262,110.
+pub(crate) const MAX_EVENT_BYTES: usize = 256 * 1024 - TIMESTAMP_BYTES - EVENT_STORAGE_OVERHEAD;
+/// CW max events per PutLogEvents call.
+const MAX_NUM_OF_LOG_EVENTS: usize = 10_000;
+/// Max time span between earliest and latest event in a batch (23 hours in ms).
+const MAX_TIME_SPAN_MS: i64 = 23 * 60 * 60 * 1000;
+/// Events older than 14 days are rejected by CW.
+const FOURTEEN_DAYS_MS: i64 = 14 * 24 * 60 * 60 * 1000;
+/// Events more than 2 hours in the future are rejected by CW.
+const TWO_HOURS_FUTURE_MS: i64 = 2 * 60 * 60 * 1000;
+
+/// Per-event effective size for batch accounting.
+fn event_wire_size(message_bytes: usize) -> usize {
+    message_bytes + TIMESTAMP_BYTES + EVENT_STORAGE_OVERHEAD
+}
+
+/// Seal log events into batches ready for CloudWatch upload.
+///
+/// Events are sorted by timestamp, filtered by age and log level,
+/// oversized events are chunked, and batches are sealed at CW API limits.
+#[must_use]
+pub(crate) fn seal_batches(
+    mut events: Vec<LogEvent>,
+    log_group: &str,
+    log_stream: &str,
+    min_log_level: Option<LogLevel>,
+    now_ms: i64,
+) -> Vec<SealedBatch> {
+    // CW requires events sorted by timestamp. Must use stable sort to preserve
+    // insertion order for events with equal timestamps.
+    events.sort_by_key(|e| e.timestamp);
+
+    let cutoff = now_ms - FOURTEEN_DAYS_MS;
+    let max_future = now_ms + TWO_HOURS_FUTURE_MS;
+
+    let mut batches: Vec<SealedBatch> = Vec::new();
+    let mut current_events: Vec<LogEvent> = Vec::new();
+    let mut current_size: usize = 0;
+    let mut earliest_ts: Option<i64> = None;
+
+    let seal_current = |events: &mut Vec<LogEvent>,
+                        size: &mut usize,
+                        earliest: &mut Option<i64>,
+                        batches: &mut Vec<SealedBatch>| {
+        if !events.is_empty() {
+            batches.push(SealedBatch {
+                log_group: log_group.to_string(),
+                log_stream: log_stream.to_string(),
+                events: std::mem::take(events),
+            });
+            *size = 0;
+            *earliest = None;
+        }
+    };
+
+    for event in events {
+        // Drop events outside CW allowed range
+        if event.timestamp < cutoff || event.timestamp > max_future {
+            tracing::debug!(
+                timestamp = event.timestamp,
+                cutoff = cutoff,
+                max_future = max_future,
+                "Dropping event outside CW allowed range"
+            );
+            continue;
+        }
+
+        // minimumLogLevel filtering for GG structured log format
+        if let Some(min_level) = min_log_level {
+            if should_filter_by_level(&event.message, min_level) {
+                tracing::debug!(
+                    min_level = ?min_level,
+                    "Filtering event below minimum log level"
+                );
+                continue;
+            }
+        }
+
+        // Chunk oversized messages
+        let chunks = chunk_message(&event.message, event.timestamp);
+
+        for chunk in chunks {
+            if chunk.message.is_empty() {
+                continue;
+            }
+
+            let wire_size = event_wire_size(chunk.message.len());
+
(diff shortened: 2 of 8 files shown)


src/uploader/batcher.rs
```diff
@@ -1,4 +1,594 @@
...
+    log_group: &str,
+    log_stream: &str,
+    min_log_level: Option<LogLevel>,
+) -> Vec<SealedBatch> {
+    events.sort_by_key(|e| e.timestamp);
+
+    let now_ms = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| d.as_millis() as i64)
+        .unwrap_or(0);
```
alter-mage: If `duration_since` ever returns `Err` (clock before epoch), `now_ms` becomes 0 and every event gets silently dropped as too old. Since this can't happen on a real device, could we use `expect` here to make the failure loud rather than silent?
```rust
let now_ms = std::time::SystemTime::now()
    .duration_since(std::time::UNIX_EPOCH)
    .expect("system clock is before Unix epoch")
    .as_millis() as i64; // safe: won't exceed i64::MAX until year ~292 million
```

src/uploader/batcher.rs
```diff
@@ -1,4 +1,594 @@
...
+
+/// Check if an event should be filtered based on minimumLogLevel.
+/// Only applies to Greengrass structured log format: `YYYY-MM-DD HH:MM:SS [LEVEL]`.
+/// Non-matching lines (EMF JSON, plain text) always pass through.
+///
+/// DIVERGENCE: Java deserializes each line as `GreengrassLogMessage` JSON to extract
+/// the level. We use text pattern matching instead — cheaper, and sufficient for the
+/// text-based GG log format. JSON-structured GG logs would bypass this filter.
+/// For EMF sources, `minimumLogLevel` is always `DEBUG` so this is a no-op.
+fn should_filter_by_level(message: &str, min_level: LogLevel) -> bool {
```
alter-mage, line 227: nit: the magic indices are hard to follow without context — could you add a comment mapping them to the format?
```rust
// Expected format: "YYYY-MM-DD HH:MM:SS [LEVEL] ..."
//                   0123456789012345678901
//                   [4]='-', [10]=' ', [13]=':', [19..] contains "[LEVEL]"
```

src/uploader/batcher.rs
```diff
@@ -1,4 +1,594 @@
...
+                continue;
+            }
+
+            let wire_size = event_wire_size(chunk.message.len());
+
+            // Check 23-hour span
+            let earliest = earliest_ts.unwrap_or(chunk.timestamp);
+            if chunk.timestamp - earliest > MAX_TIME_SPAN_MS && !current_events.is_empty() {
+                batches.push(SealedBatch {
+                    log_group: log_group.to_string(),
```
alter-mage: `log_group.to_string()` and `log_stream.to_string()` are repeated 4 times each. A local closure would remove the duplication:
```rust
let make_batch = |events| SealedBatch {
    log_group: log_group.to_string(),
    log_stream: log_stream.to_string(),
    events,
};
```
Then each seal becomes `batches.push(make_batch(std::mem::take(&mut current_events)));`

src/uploader/batcher.rs
```diff
@@ -1,4 +1,594 @@
...
+        let events: Vec<LogEvent> = (0..10_000).map(|i| ev(now + i, "x")).collect();
+        let batches = seal_batches(events, "/grp", "/s", None);
+        assert_eq!(batches.len(), 1);
+        assert_eq!(batches[0].events.len(), 10_000);
+    }
+
+    #[test]
+    fn test_batch_overflow_not_dropped() {
+        let now = now_ms();
+        // Each event: 600KB msg + 34 overhead ≈ 600KB wire. Two = ~1.2MB > 1MB limit.
```
alter-mage: nit: the comment "Two = ~1.2MB > 1MB limit" is misleading — each 600KB message gets chunked into ~3 × 256KB chunks before batching, so the splits are driven by chunk sizes not original message sizes. Could you update the comment to reflect what's actually happening?

src/uploader/batcher.rs
```diff
@@ -1,4 +1,594 @@
...
+            tracing::debug!(
+                timestamp = event.timestamp,
+                "Dropping event outside CW allowed range"
+            );
+            continue;
+        }
+
+        // minimumLogLevel filtering for GG structured log format
+        if let Some(min_level) = min_log_level {
+            if should_filter_by_level(&event.message, min_level) {
```
alter-mage, line 105: Java's `checkAndAddNewLogEvent` deserializes each line as `GreengrassLogMessage` JSON and applies the level filter using the embedded `level` field. Rust's text pattern only matches `YYYY-MM-DD HH:MM:SS [LEVEL]` — JSON-structured GG component logs don't match this pattern and always pass through regardless of `minimumLogLevel`. The PR notes this is fine for EMF (where `minimumLogLevel` is always `DEBUG`), but for GG component log upload, the level filter would be silently bypassed. Is that an acceptable divergence for this PR's scope?

src/uploader/batcher.rs
paabbs: Will change that, good catch!

src/uploader/batcher.rs
paabbs: Will add the comment

src/uploader/batcher.rs
paabbs: Ack, will refactor

src/uploader/batcher.rs
paabbs: Sure, I can update the comment

src/uploader/batcher.rs
paabbs: Will add this as a part of this PR

alter-mage (approved):

src/uploader/batcher.rs
```diff
@@ -1,4 +1,627 @@
...
+            "2024-01-15 10:30:45 [DEBUG] text log",
+            LogLevel::Error
+        ));
+        assert!(!should_filter_by_level("short", LogLevel::Error));
+    }
+
+    // --- Boundary tests (exact-at-limit) ---
+
+    #[test]
+    fn test_14_day_boundary_exact_kept() {
```
aws-kevinrickard, line 525: `test_14_day_boundary_exact_kept` and `test_2h_future_boundary_one_ms_over_dropped` are timing-flaky. Both tests capture `SystemTime::now()` to compute the event timestamp, and `seal_batches` independently captures its own `SystemTime::now()` a moment later. When those two calls don't land in the same millisecond, the tests flip.

Let `T` = the test's `now_ms()`, `T + δ` = `seal_batches`'s `now_ms()`, δ ≥ 0.

**`test_14_day_boundary_exact_kept`:**

- event.timestamp = `T - 14d`
- cutoff = `(T + δ) - 14d`
- check: `event.timestamp < cutoff` ⇔ `0 < δ`
- δ = 0 → kept (test passes). δ ≥ 1 → dropped (test fails, expected `batches.len() == 1`).

**`test_2h_future_boundary_one_ms_over_dropped`:**

- event.timestamp = `T + 2h + 1`
- max_future = `(T + δ) + 2h`
- check: `event.timestamp > max_future` ⇔ `1 > δ`
- δ = 0 → dropped (test passes). δ ≥ 1 → kept (test fails, expected `batches.is_empty()`).

Locally δ is 0 because both calls happen within a few hundred ns and truncate to the same ms. In CI — shared-tenant VMs under load, debug/coverage instrumentation, scheduler pauses — δ can exceed 1 ms and flip the assertion.

Suggested fix: Use dependency injection to make it testable. Make `seal_batches` accept `now_ms` as a parameter (or a `Clock` trait) so tests can pass the same `T` they used to build the event:

```rust
pub(crate) fn seal_batches(
    mut events: Vec<LogEvent>,
    log_group: &str,
    log_stream: &str,
    min_log_level: Option<LogLevel>,
    now_ms: i64,
) -> Vec<SealedBatch> { … }
```

Side benefit: `seal_batches` becomes a pure function — easier to reason about and friendlier to future property/fuzz tests. Production caller passes `SystemTime::now()` itself.

src/uploader/batcher.rs
```diff
@@ -1,4 +1,627 @@
...
+    while pos < bytes.len() {
+        let end = (pos + MAX_EVENT_LENGTH).min(bytes.len());
+        // Find a valid UTF-8 char boundary
+        let mut boundary = end;
+        while boundary > pos && !message.is_char_boundary(boundary) {
+            boundary -= 1;
+        }
+        if boundary == pos {
+            // Safety: can't happen with valid UTF-8 (max 4 bytes < MAX_EVENT_LENGTH),
+            // but guard against infinite loop.
```
aws-kevinrickard, line 202: We should also guard against data loss by informing the customer that something went wrong. At least log an error.

src/uploader/batcher.rs
```diff
@@ -1,4 +1,627 @@
...
+            *earliest = None;
+        }
+    };
+
+    for event in events {
+        // Drop events outside CW allowed range
+        if event.timestamp < cutoff || event.timestamp > max_future {
+            tracing::debug!(
+                timestamp = event.timestamp,
+                "Dropping event outside CW allowed range"
```
aws-kevinrickard, line 98: This would be much more useful to a customer if it showed what the timestamp, cutoff, and max_future values are

src/uploader/batcher.rs
```diff
@@ -1,4 +1,627 @@
...
+            );
+            continue;
+        }
+
+        // minimumLogLevel filtering for GG structured log format
+        if let Some(min_level) = min_log_level {
+            if should_filter_by_level(&event.message, min_level) {
+                continue;
```
aws-kevinrickard, line 110: We should debug log here with why it's being ignored before continuing, so that customers can figure out why their logs aren't being uploaded if they set their config wrong

src/uploader/batcher.rs
```diff
@@ -1,4 +1,627 @@
...
+    if bytes.len() <= MAX_EVENT_LENGTH {
+        return vec![LogEvent {
+            timestamp,
+            message: message.to_string(),
+        }];
+    }
+
+    tracing::trace!(
+        size = bytes.len(),
+        "Chunking oversized event into multiple CW events"
```
aws-kevinrickard, line 188: This should be debug, and ideally also list the number of chunks it was split into.

Reasoning: Breaking up large lines into multiple events is an unfortunate reality that can cause unexpected problems for customers (e.g. A stack trace emitted at error level is only tagged with error in the first event, so their query which filters to error events won't search the next events like they might expect). Having this debug, and also showing the number of chunks it was split into, would help them figure out what's going on when they're inspecting things.

src/uploader/batcher.rs
```diff
@@ -1,4 +1,627 @@
...
+///
+/// Events are sorted by timestamp, filtered by age and log level,
+/// oversized events are chunked, and batches are sealed at CW API limits.
+pub(crate) fn seal_batches(
+    mut events: Vec<LogEvent>,
+    log_group: &str,
+    log_stream: &str,
+    min_log_level: Option<LogLevel>,
+) -> Vec<SealedBatch> {
+    events.sort_by_key(|e| e.timestamp);
```
aws-kevinrickard, line 66: Nit: Might want to put a comment calling out that the sort used here must be stable. There's currently no unit tests checking for a regression like this so it wouldn't be caught if someone swapped this for some other sort that was unstable.

src/uploader/batcher.rs
```diff
@@ -1,4 +1,627 @@
...
+use crate::scanner::LogEvent;
+
+/// CW per-event framing overhead (bytes).
+const EVENT_STORAGE_OVERHEAD: usize = 26;
+/// Timestamp field size (bytes).
+const TIMESTAMP_BYTES: usize = 8;
+/// CW PutLogEvents max payload: 1 MB.
+const MAX_BATCH_SIZE: usize = 1_048_576;
+/// CW max message bytes per event: 256KB - 8 - 26 = 262,110.
+pub(crate) const MAX_EVENT_LENGTH: usize = 256 * 1024 - TIMESTAMP_BYTES - EVENT_STORAGE_OVERHEAD;
```
aws-kevinrickard: Nit: Better to use `MAX_EVENT_BYTES` to remove ambiguity with respect to bytes vs characters

src/scanner/multiline.rs
```diff
@@ -254,7 +238,8 @@ mod tests {
         ];
         let events = assemble_multiline(lines, Some(&pattern));
         assert_eq!(events.len(), 1);
-        assert!(events[0].message.len() <= MAX_EVENT_SIZE);
+        // No truncation — batcher handles chunking
+        assert!(events[0].message.len() > 200_000);
```
aws-kevinrickard: Nit: This doesn't fully assert that no truncation happened, but easily could:

```rust
// Expected total: "START event" (11) + 2 * 200_000 = 400_011
assert_eq!(events[0].message.len(), "START event".len() + 2 * large_line.len());
```

src/uploader/batcher.rs
paabbs: thanks, will add a small tolerance buffer (±5ms) to the boundary tests to account for clock drift between the test and seal_batches

src/uploader/batcher.rs
paabbs: Sure, I can add tracing::error! although I am not sure if that would be reachable

src/uploader/batcher.rs
paabbs: Makes sense. Will include all three values in the debug log.

src/uploader/batcher.rs
paabbs: True, will add that

src/uploader/batcher.rs
paabbs: Sure, I can change to debug and include chunk count.

src/uploader/batcher.rs
paabbs: Will be addressing this as a part of the revision

src/uploader/batcher.rs
paabbs: Will rename that

src/scanner/multiline.rs
paabbs: Will be modifying the test

alter-mage (approved):

src/uploader/batcher.rs
aws-kevinrickard: Thanks for adding the injection rather than an arbitrary buffer, I think it is better

src/scanner/multiline.rs
```diff
@@ -254,7 +238,9 @@ mod tests {
         ];
         let events = assemble_multiline(lines, Some(&pattern));
         assert_eq!(events.len(), 1);
-        assert!(events[0].message.len() <= MAX_EVENT_SIZE);
+        // No truncation — batcher handles chunking
+        // "START event" (11) + "\n" + 200_000 x's + "\n" + 200_000 x's = 400_011
```
aws-kevinrickard: Nit: why are "\n"s here? They are not put in by assemble_multiline, and if they were counted then the count should be 400_013

paabbs: Thanks for the catch! Will fix this

aws-kevinrickard (approved):

PR #250, opened 2026-05-06
feat(uploader): upload orchestration with retry, checkpoint advancement, and disk cleanup

paabbs:
**Issue #, if available:**

N/A — Upload orchestrator for Rust v2 LogManager

**Description of changes:**

Wire the full upload cycle matching Java LogManager's `processLogsAndUpload()`: scan → read → multiline → batch → upload → checkpoint → disk cleanup.

**Key changes:**
- Full pipeline split into `scan_and_filter_files` → `read_file_events` → `upload_and_advance_checkpoints`
- App-level retry with [2,4,8]s backoff for retriable CW errors (LimitExceeded, ResourceNotFound)
- Graceful shutdown via SIGTERM with `SHUTDOWN_REQUESTED` checks between sources (5s Classic grace period)
- Disk space enforcement every cycle regardless of upload success — deletes oldest non-active files when over limit, protecting device from disk exhaustion when offline (per [AWS docs](https://docs.aws.amazon.com/greengrass/v2/developerguide/log-manager-component.html))
- Config schema restructured to match official docs:
  - `logsUploaderConfiguration` wrapper with `componentLogsConfigurationMap` (map) + `systemLogsConfiguration` (object)
  - Supports both nested and flat config formats
  - Legacy `componentLogsConfiguration` field name accepted via serde alias
  - All fields accept both string and native JSON types (matching Java's `Coerce` behavior)
- Recipe: `aws.greengrass.LogManager`, `aarch64` + `armv8l` + Windows manifests, `{configuration:/}` in Setenv, no shutdown lifecycle
- `LogLevel` default = `Info` (matching Java); EMF bypasses filtering regardless (no `level` field)
- `resolve_thing_name()` logs `error!` when `AWS_IOT_THING_NAME` missing/empty, proceeds with "unknown"
- TTL eviction of stale checkpoint entries (24h, matching Java `ProcessingFiles.evictStaleEntries()`)
- `uploadToCloudWatch` check for system logs (disabled by default, warns if enabled without path)
- E2E upload pipeline tests with `StaticReplayClient` mock

Connects all previously-merged modules (scanner, batcher, CW client, disk management, checkpoint) into a working component. Without this, the binary loads config but does nothing.

- 224 unit + integration tests pass (including 2 E2E upload pipeline tests with mocked CW)
- End-to-end on GG Classic container: component deployed via GG CLI, EMF events verified in CloudWatch (us-west-2)
- End-to-end on GG Lite container: component deployed via cloud API to thing group, EMF events verified in CloudWatch (us-east-1)
- Both `x86_64-unknown-linux-musl` and `aarch64-unknown-linux-musl` static builds verified with cargo-zigbuild
- `cargo clippy` clean (zero warnings), `cargo fmt` applied

On GG Lite, TES requires explicit port configuration in the device config ([aws-greengrass-lite#950](https://github.com/aws-greengrass/aws-greengrass-lite/issues/950)):
```yaml
services:
  aws.greengrass.TokenExchangeService:
    configuration:
      port: 8090
```

**Behavioral difference from Java LogManager:**
- Disk enforcement deletes oldest non-active files even if not yet uploaded (Java only deletes after upload). This matches the [documented behavior](https://docs.aws.amazon.com/greengrass/v2/developerguide/log-manager-component.html) and protects IoT devices from disk exhaustion when offline.

**Checklist:**
- [x] Updated or added new end-to-end tests


recipe.yaml
```diff
@@ -4,39 +4,39 @@ ComponentVersion: "{COMPONENT_VERSION}"
...
           Execute: OWNER
     Lifecycle:
-      install: chmod +x {artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-arm64
-      run: "{artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-arm64"
+      install: "chmod +x {artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-arm64"
+      Setenv:
+        GG_LOG_MANAGER_CONFIG: "{configuration:/configJson}"
+      run: "{artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-arm64 --work-dir {work:path}"
       shutdown: SIGTERM
-  - Platform:
```
utsavvyas, line 35: We need to support windows platform

recipe.yaml
```diff
@@ -4,39 +4,39 @@ ComponentVersion: "{COMPONENT_VERSION}"
...
+      runtime: "*"
     Artifacts:
       - URI: s3://gg-components/gg-log-manager-{COMPONENT_VERSION}-linux-amd64
         Permission:
           Execute: OWNER
     Lifecycle:
-      install: chmod +x {artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-amd64
-      run: "{artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-amd64"
+      install: "chmod +x {artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-amd64"
+      Setenv:
```
utsavvyas, line 30: Why do we need this now ?

README.md
```diff
@@ -1,10 +1,64 @@
-# AWS Greengrass Log Manager (Rust)
+## Log Manager (Rust)
```
utsavvyas: AWS GreenGrass Log Manager

README.md
```diff
@@ -1,10 +1,64 @@
...
 
-A Generic Type component for AWS IoT Greengrass that tails log files and EMF JSON files, uploading them to Amazon CloudWatch Logs.
+Rust reimplementation of `aws.greengrass.LogManager` as a Generic Type component for GG Classic and GG Lite.
```
utsavvyas: Not needed.

README.md
```diff
@@ -1,10 +1,64 @@
...
 
-This is a Rust reimplementation of `aws.greengrass.LogManager`, targeting minimal binary size and memory footprint for resource-constrained devices running GG Classic or GG Lite.
+The log manager component collects and uploads logs from Greengrass core devices to Amazon CloudWatch Logs.
+It tails configured log directories, batches events respecting CloudWatch API limits, and manages disk space
+by deleting uploaded files when configured.
 
-## Building
+### Features
```
utsavvyas, line 9: Should we call out that data loss can happen if cloud connection is lost and disk limit is full

README.md
```diff
@@ -1,10 +1,64 @@
...
+
+**Logs Uploader** — Scans configured directories for log files matching a regex pattern, reads new content
+from persisted byte offsets, batches events (1MB / 10K events / 23h span limits), and uploads via PutLogEvents.
+Supports multi-line log assembly, EMF JSON passthrough, and minimumLogLevel filtering for JSON-structured
+Greengrass logs.
+
+**Checkpoint Persistence** — Tracks per-file read offsets using content-hash-based file identity (survives
+log rotation). Atomic checkpoint writes with fsync for crash safety. Compatible with Java LogManager v2.3.1+
```
utsavvyas: Does previous version of logmanager has different kind of checkpoint logic ?

src/config/schema.rs
```diff
@@ -44,8 +44,11 @@ impl From<std::io::Error> for ConfigError {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Default)]
 #[serde(rename_all = "UPPERCASE")]
 pub enum LogLevel {
-    Debug,
+    /// Default is Debug (= no filtering). Java only filters JSON-structured GG logs,
+    /// not text or EMF. Since EMF is our primary use case and all EMF sources use DEBUG,
```
utsavvyas: This seems wrong behaviour. We should be able to identify the emf logs. User may want emf logs along with info logs upload to the cloud and this behaviour seems weird.

src/credentials/mod.rs
```diff
@@ -1,6 +1,37 @@
...
+//! Credentials are handled by the AWS SDK default chain (`aws_config::load_defaults`),
+//! which picks up TES-provided credentials automatically when running as a GG component.
+//! Both Classic and Lite set `AWS_IOT_THING_NAME` for components.
 
-mod tes;
+/// Resolve ThingName from `$AWS_IOT_THING_NAME` environment variable.
+/// Both GG Classic and GG Lite set this for component processes.
+pub fn resolve_thing_name() -> Option<String> {
+    let name = std::env::var("AWS_IOT_THING_NAME").ok();
+    tracing::info!(thing_name = ?name, "Resolved thing name");
```
utsavvyas: If the name is empty we should atleast log error

tests/log_manager_integration_test.rs
```diff
@@ -0,0 +1,705 @@
...
+    // Read from checkpoint offset — only gets new content
+    let (events2, _) = read_file_from_offset(&offsets2[0].0, offsets2[0].1, now_ms()).unwrap();
+    assert_eq!(events2.len(), 1);
+    assert!(events2[0].message.contains("line3"));
+}
+
+// --- T3: Completed file detection ---
+
+#[test]
+fn test_completed_file_detection() {
```
utsavvyas, line 187: This integ test is confusing. We want to test E2E behaviour right ?

tests/orchestrator_integration_test.rs
```diff
@@ -0,0 +1,705 @@
...
+    // With override
+    let interval = effective_interval_secs(Some(60), 300);
+    assert!((60..=65).contains(&interval));
+
+    // Without override — uses global
+    let interval = effective_interval_secs(None, 300);
+    assert!((300..=305).contains(&interval));
+}
+
+// --- T10: Checkpoint persistence rate-limiting ---
```
utsavvyas: Why do we need rate limit on checkpoint ?

tests/orchestrator_integration_test.rs
```diff
@@ -0,0 +1,705 @@
...
+}
+
+// --- T14: Multi-line assembly end-to-end ---
+
+#[test]
+fn test_multiline_assembly_end_to_end() {
+    let dir = TempDir::new().unwrap();
+
+    // Write multi-line Java stack trace
+    write_emf_file(
```
utsavvyas: This is not emf file rigth ?

tests/log_manager_integration_test.rs
```diff
@@ -0,0 +1,705 @@
...
+    for line in lines {
+        writeln!(f, "{}", line).unwrap();
+    }
+    path
+}
+
+// --- T1: Full cycle scan → read → batch → checkpoint advance ---
+
+#[test]
+fn test_full_cycle_scan_read_batch_checkpoint() {
```
utsavvyas, line 53: Can we have logs along with emf into one example ?

src/uploader/mod.rs
```diff
@@ -1,34 +1,264 @@
...
+const BACKOFF_DELAYS: [u64; 3] = [2, 4, 8];
+
+/// Upload a single batch with app-level retry for Retriable errors.
+/// Returns Ok(true) on success, Ok(false) if retries exhausted, Err on non-retriable.
+#[cfg(feature = "aws-sdk")]
+pub async fn upload_with_retry(
+    client: &mut CwLogsClient,
+    batch: &SealedBatch,
+) -> Result<bool, String> {
+    for (attempt, delay) in BACKOFF_DELAYS.iter().enumerate() {
```
utsavvyas: Why do we need this ? I thought we do rely on the sdk level retry. These do seem overkill

src/main.rs
```diff
@@ -37,118 +54,425 @@ async fn main() {
...
-    // Main loop placeholder - wait for shutdown
-    shutdown_notify.notified().await;
+    // Main upload loop
+    let mut last_persist = Instant::now();
+    let persist_interval = Duration::from_secs(config.periodic_upload_interval_sec);
 
-    // Graceful shutdown sequence
-    graceful_shutdown().await;
-}
+    loop {
```
utsavvyas, line 111: Any reason for using the loop instead of periodic scheduler ?

src/main.rs
```diff
@@ -37,118 +54,425 @@ async fn main() {
...
+    info!("Shutdown complete");
 }
 
-fn initiate_shutdown(notify: &Notify) {
-    if !SHUTDOWN_REQUESTED.swap(true, Ordering::SeqCst) {
-        info!("Received shutdown signal");
-        notify.notify_one();
+/// Process all configured log sources in one cycle.
+#[cfg(not(tarpaulin_include))]
+async fn process_all_sources(
```
utsavvyas, line 148: Can we reduce duplication ?

src/main.rs
```diff
@@ -37,118 +54,425 @@ async fn main() {
...
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap_or_default()
+        .as_millis() as i64;
+
+    let mut file_events: Vec<(PathBuf, Vec<gg_log_manager::scanner::LogEvent>, u64, String)> =
+        Vec::with_capacity(offsets.len());
+
+    for (path, offset) in &offsets {
+        // Skip files already fully read and not active (rotation edge case)
+        // Java: startPosition == file.length() && !isActiveFile → delete + update timestamp
```
utsavvyas: nit: Not needed

src/main.rs
```diff
@@ -37,118 +54,425 @@ async fn main() {
...
+            .map(|s| s.to_string())
+            .unwrap_or_else(|| derive_log_group_name(&comp.component_name, None));
+        let pattern = match Regex::new(&comp.source.log_file_regex) {
+            Ok(r) => r,
+            Err(e) => {
+                error!(component = comp.component_name, error = %e, "Invalid file regex, skipping");
+                continue;
+            }
+        };
+        process_source(
```
utsavvyas: Do java parallelize this execution ? If one component has very large logs we might cause impact on other component as well.

src/main.rs
```diff
@@ -37,118 +54,425 @@ async fn main() {
...
 
+/// Process a single log source: scan → read → batch → upload → checkpoint → disk.
+// Note: Uses std::fs (blocking I/O) inside async. This is safe because we run on
+// tokio current_thread runtime — there are no other tasks to block. If the runtime
+// is ever changed to multi-thread, these calls must be replaced with tokio::fs.
 #[cfg(not(tarpaulin_include))]
-async fn graceful_shutdown() {
-    info!("Step 1/4: Cancelling pending upload futures");
-    cancel_pending_uploads();
+async fn process_source(
```
utsavvyas, line 205: Having single method for 200+ line doesnt make sense.

src/main.rs
```diff
@@ -37,118 +54,425 @@ async fn main() {
...
+                    warn!(path = %path.display(), error = %e, "Failed to delete completed file");
+                } else {
+                    info!(path = %path.display(), "Deleted completed file after upload");
+                }
+            }
+        }
+
+        // Disk space enforcement — pass ALL fully-uploaded non-active files, not just this cycle's
+        if let Some(limit_str) = source.disk_space_limit.as_deref() {
+            if let Ok(limit_val) = limit_str.parse::<u64>() {
```
utsavvyas, line 431: What happens if the disk size is over the limit and upload is not completed ?

src/main.rs
```diff
@@ -37,118 +54,425 @@ async fn main() {
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    // TTL eviction of stale checkpoint entries
+    evict_stale_entries(store, log_group, TTL_24H_MS);
```
utsavvyas, line 471: Why this is needed ?

recipe.yaml
paabbs: Will add that after testing for windows instances

recipe.yaml
aws-kevinrickard: Should be aws.greengrass.LogManager ?

recipe.yaml
```diff
@@ -4,39 +4,39 @@ ComponentVersion: "{COMPONENT_VERSION}"
...
+ComponentDependencies:
+  aws.greengrass.TokenExchangeService:
+    VersionRequirement: ">=0.0.0"
+    DependencyType: HARD
 ComponentConfiguration:
   DefaultConfiguration:
-    periodicUploadIntervalSec: 300
-    componentLogsConfiguration: []
-    systemLogsConfiguration: []
+    configJson: '{"periodicUploadIntervalSec":300,"componentLogsConfiguration":[],"systemLogsConfiguration":[]}'
```
aws-kevinrickard: 1.  systemLogsConfiguration  should be logsUploaderConfiguration. systemLogsConfiguration  (nested) , and should be a map instead of a list

2.  componentLogsConfiguration should be componentLogsConfigurationMap , and should be a map instead of a list

3. Why is made as json in a string under key configJson?

recipe.yaml
```diff
@@ -4,39 +4,39 @@ ComponentVersion: "{COMPONENT_VERSION}"
...
         Permission:
           Execute: OWNER
     Lifecycle:
-      install: chmod +x {artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-amd64
-      run: "{artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-amd64"
+      install: "chmod +x {artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-amd64"
+      Setenv:
+        GG_LOG_MANAGER_CONFIG: "{configuration:/configJson}"
+      run: "{artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-amd64 --work-dir {work:path}"
       shutdown: SIGTERM
```
aws-kevinrickard: When the shutdown lifecycle is defined as a string, it represents the command to be run. "SIGTERM" doesn't make sense here. See https://docs.aws.amazon.com/greengrass/v2/developerguide/component-recipe-reference.html#recipe-format

The run lifecycle will be sent sigterm on shutdown by default already; If that's what you want, this can be removed

recipe.yaml
```diff
@@ -4,39 +4,39 @@ ComponentVersion: "{COMPONENT_VERSION}"
...
       architecture: amd64
+      runtime: "*"
     Artifacts:
       - URI: s3://gg-components/gg-log-manager-{COMPONENT_VERSION}-linux-amd64
         Permission:
           Execute: OWNER
     Lifecycle:
-      install: chmod +x {artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-amd64
-      run: "{artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-amd64"
+      install: "chmod +x {artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-amd64"
```
aws-kevinrickard: The artifacts already have execute permission for the owner set on line 22, this should be redundant

recipe.yaml
```diff
@@ -4,39 +4,39 @@ ComponentVersion: "{COMPONENT_VERSION}"
...
+      Setenv:
+        GG_LOG_MANAGER_CONFIG: "{configuration:/configJson}"
+      run: "{artifacts:path}/gg-log-manager-{COMPONENT_VERSION}-linux-amd64 --work-dir {work:path}"
       shutdown: SIGTERM
   - Platform:
       os: linux
       architecture: arm64
```
aws-kevinrickard: This should be:
```suggestion
      architecture: aarch64
```

Additionally, if you want to support armv8l, you'll need to add `"architecture.detail": "armv8l",` as well. See the aws.greengrass.SecureTunneling 2.0.0 component recipe as an example , and https://docs.aws.amazon.com/greengrass/v2/developerguide/component-recipe-reference.html#recipe-format for more info

src/config/schema.rs
aws-kevinrickard, line 428: This should be componentLogsConfigurationMap as well, see https://docs.aws.amazon.com/greengrass/v2/developerguide/log-manager-component.html#log-manager-component-configuration for existing configuration schema

src/config/schema.rs
aws-kevinrickard: The current default is Info, I'd rather not make a change that could break people

src/main.rs
```diff
@@ -21,7 +35,10 @@ async fn main() {
     tracing_subscriber::fmt::init();
 
     let args: Vec<String> = std::env::args().collect();
-    let config_arg = parse_args(&args);
+    let config_arg = parse_arg(&args, "--config")
+        .or_else(|| std::env::var("GG_LOG_MANAGER_CONFIG").ok())
```
aws-kevinrickard, line 39: I would suggest using the component sdk to read configs, it's usually cleaner and more flexible. But injecting them into args or env vars is also an acceptable tradeoff if you're okay with the component restarting on any config changes, and wiring everything up in the recipe.

src/main.rs
```diff
@@ -37,118 +54,425 @@ async fn main() {
...
+        components = config.component_logs_configuration.len(),
+        system = config.system_logs_configuration.len(),
+        "Configuration loaded"
     );
 
-    let shutdown_notify = Arc::new(Notify::new());
-    let shutdown_notify_clone = shutdown_notify.clone();
+    // ThingName resolution
+    let thing_name = resolve_thing_name().unwrap_or_else(|| {
+        warn!("AWS_IOT_THING_NAME not set, using 'unknown'");
```
aws-kevinrickard: If we ever can't resolve a thing name, something has gone very wrong, I think this should be error.

Proceeding with an unknown string is acceptable though

recipe.yaml
paabbs: Needed for GG Lite compatibility. Per the [recipe reference docs](https://docs.aws.amazon.com/greengrass/v2/developerguide/component-recipe-reference.html), omitting runtime means the manifest only matches Classic devices. runtime: "*" ensures deployment to both GG Classic and GG Nucleus Lite

README.md
paabbs: Will revert

README.md
paabbs: Thanks for this! I did a deep dive into this after you posed this question and observed that `diskSpaceLimit` is only enforced after successful upload. 

If the device is offline, log files accumulate past the limit with no cleanup — this is the current behavior  in the Java LogManager as well. Classic does have  a mechanism to clean up the logs for `/greengrass/v2/logs/` but that's just nucleus sepicifc logs. AS of now neither classic nor Lite has any other mechanism to clean up the directories LogManager monitors (e.g., `/var/log/gg-metrics/`).

paabbs: I'll fix this by enforcing diskSpaceLimit every scan cycle regardless of upload status, deleting oldest non-active files when over the limit. Any other thoughts are appreciated as well!

README.md
paabbs: Yeah, Java <2.3.1 used a flat single-file-per-component checkpoint format (currentComponentFileProcessingInformation). v2.3.1 introduced hash-keyed multi-file tracking (currentComponentFileProcessingInformationV2)

src/config/schema.rs
paabbs: Makes sense will reert back to `Info` this was a extra safe guard rail, but introduces breaking change. But yeah, Users who want DEBUG-level structured logs can explicitly set "minimumLogLevel": "DEBUG" in the recipe

src/credentials/mod.rs
paabbs: Will fix. Changing to `error!` when missing/empty, and proceeding with "unknown" as fallback.

tests/log_manager_integration_test.rs
paabbs: Will convert to a full E2E test using `aws-smithy-mocks` (AWS's official SDK mock framework) that exercises the complete scan→upload→checkpoint→delete cycle with a mocked CW client

tests/orchestrator_integration_test.rs
paabbs: No explicit rate-limitingneed though,  checkpoint saves once per upload cycle (end of loop iteration). Makes sense to remove and fold checkpoint durability verification into the E2E rewrite

tests/orchestrator_integration_test.rs
paabbs: Will rename and add it to the E2E tests

tests/log_manager_integration_test.rs
paabbs: Will add, expanding T1 to include both EMF files (*.emf.json) and component log files (*.log) processed through the pipeline, showing the full cycle works for both source type

src/uploader/mod.rs
paabbs: SDK retry handles HTTP-level transient failures (timeouts, 500s). This app-level retry handles errors that come back as successful HTTP responses but are logically retriable LimitExceededException on group/stream creation, ResourceNotFoundException when a group/stream is deleted externally between cache check and upload. Saw the same with Java implementation, can drop it if deemed unnecessary

src/main.rs
paabbs: loop + tokio::select! on sleep gives scheduleWithFixedDelay semantics, the interval starts after processing completes, preventing overlapping cycles when uploads take longer than expected. Also enables instant shutdown via the select branch. Java uses `ScheduledExecutorService.scheduleWithFixedDelay` which is the same pattern.

src/main.rs
paabbs: Sure will refactor

src/main.rs
paabbs: Java processes all components sequentially as well, single for loop to collect metadata, then componentMetadata.forEach(unit -> processLogFiles + upload). Took the same approach ehre as well. If one component has large logs, others wait until next cycle. This avoids CW rate limit issues and keeps checkpoint management simple. Would you want me to defer and optimize it?

src/main.rs
paabbs: Fair point, will refactor into smaller functions: `scan_and_filter_files`,` read_file_events`, and `upload_and_advance_checkpoints`. Cna make it similar to Java

src/main.rs
paabbs: Currently nothing, disk enforcement is gated behind if !result.succeeded.is_empty(), so it only runs when at least one upload succeeds. Same behavior as Java. I can raise it as a bug and send out a fix by moving disk enforcement outside this gate so it runs every cycle, deleting oldest non-active files when over the limit regardless of upload status (as noted in the README comment).

src/main.rs
paabbs: Prevents unbounded checkpoint growth. When log files are rotated/deleted, their checkpoint entries become stale (never accessed again). Without eviction, the checkpoint JSON grows indefinitely. `last_accessed` is refreshed every cycle for active files (touch-on-read in `recover_offsets`), so only entries for files that no longer exist go stale. Java has the same mechanism: `ProcessingFiles.evictStaleEntries()` with `MAX_CACHE_INACTIVE_TIME_SECONDS = 86400` (24h)

src/config/schema.rs
paabbs: Will restructure to match the docs: wrap in `logsUploaderConfiguration`, change `componentLogsConfiguration` (list) to `componentLogsConfigurationMap` (map keyed by component name), and make `systemLogsConfiguration` a single object. Will also support the legacy list format for backward compat with <v2.2.0 configs.

recipe.yaml
paabbs: Will fix the schema by restructuring `DefaultConfiguration` to match the official format: `logsUploaderConfiguration` → `componentLogsConfigurationMap` (map) + `systemLogsConfiguration` (object).

 > "why configJson as string?"

recipe interpolation on an object node already serializes it to JSON, so the manual stringification was unnecessary. Will change to `{configuration:/logsUploaderConfiguration}` in Setenv, which gives proper merge/reset support for fleet deployments while the binary still receives JSON via env var.

recipe.yaml
paabbs: Good catch, will remove

recipe.yaml
paabbs: Will remove `shutdown: SIGTERM`,  it's actively harmful on Classic (tries to execute "SIGTERM" as a shell command, wasting time before the real signal). On Lite it's ignored entirely.

One concern: Classic gives only **5 seconds** between SIGTERM and SIGKILL (hardcoded in `UnixExec.gracefulShutdownTimeout`). Will add `SHUTDOWN_REQUESTED` checks between sources so the binary exits promptly. Checkpoint save should be <1s, so 5s is sufficient if we don't block on a slow upload.

recipe.yaml
paabbs: Will fix to aarch64 and add armv8l manifest entry. Using zigbuild for cross-compilation, all targets (linux amd64/aarch64, windows amd64) build from one machine.

src/main.rs
paabbs: Agreed, env vars are an acceptable tradeoff for now. The component restarts on config changes (version bump currently, auto-restart once [GG Lite #769](https://github.com/aws-greengrass/aws-greengrass-lite/issues/769) is fixed). We can have the components SDK integration as the followups

src/main.rs
paabbs: Yep, will change to error!. Checked that both Classic and Lite always set `AWS_IOT_THING_NAME` for component processes, so its absence means the deployment environment is broken. Proceeding with "unknown" to avoid crashing.

README.md
utsavvyas: Ideally we should not change the behaviour. I do agree that this behaviour can be helpful but I would not change the behaviour.

README.md
utsavvyas: If I move from Java <2.3.1 to this log manager it will break right ? I think checkpoint consumption will not happen but even checkpoint read will fail and create panic right ?

src/main.rs
utsavvyas: Isnt scheduler at fix rate makes more sense here ?

src/uploader/mod.rs
```diff
@@ -1,34 +1,264 @@
...
+        match client.upload_batch(batch).await {
```
utsavvyas: Seems like this should move to the cw client itself ?

src/uploader/mod.rs
```diff
@@ -1,34 +1,264 @@
...
+
+    for (path, new_offset, content_hash) in succeeded {
+        // Find the scanned file to check is_active and file length
+        let scanned = scanned_files
+            .iter()
+            .find(|f| &f.content_hash == content_hash);
+        let is_active = scanned.is_some_and(|f| f.is_active);
+        let file_len = std::fs::metadata(path).map(|m| m.len()).unwrap_or(0);
+
+        if !is_active && file_len == *new_offset && file_len > 0 {
```
utsavvyas: nit : maybe creating method will make it more readable.

src/main.rs
```diff
@@ -37,118 +54,467 @@ async fn main() {
     }
 
     info!(
-        "Loaded {} component configs, {} system configs",
-        config.component_logs_configuration.len(),
-        config.system_logs_configuration.len()
+        components = config
```
utsavvyas: What is the point of the log ? Length doesnt provide enough information

src/main.rs
utsavvyas: Is it intentional ? If you need longer to upload will you move your second upload delayed as well.

src/main.rs
```diff
@@ -37,118 +54,467 @@ async fn main() {
...
+            .map(|f| f.content_hash.clone())
+            .unwrap_or_default();
+
+        file_events.push((path.clone(), assembled, new_offset, content_hash));
+    }
+
+    file_events
 }
 
+/// Phase 3: Upload events, advance checkpoints, enforce disk limits.
```
utsavvyas: This comment is confusing.

src/main.rs
```diff
@@ -37,118 +54,467 @@ async fn main() {
...
 #[cfg(not(tarpaulin_include))]
-async fn flush_pending_uploads() {
-    // Flush in-progress PutLogEvents - implementation in uploader track
+#[allow(clippy::too_many_arguments)]
+async fn upload_and_advance_checkpoints(
```
utsavvyas, line 388: We should make it modular. It seems like it should exist within the scanner

README.md
paabbs: Sounds good, the current Rust implementation is more aggressive than Java in two ways:

**Timing**: Java calls `freeDiskSpace()` only [after successful upload](https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/main/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java#L541). Rust runs it every cycle.

**Scope**: Java's [`getProcessedLogFiles()`](https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/main/src/main/java/com/aws/greengrass/logmanager/model/LogFileGroup.java#L44-L50) only returns files where `lastModified ≤ lastProcessed` (already uploaded). Rust currently deletes ANY non-active file and that includes un-uploaded ones as well.

Where as Java gets away with after-upload-only because Nucleus's Logback ([`PersistenceConfig.java#L353-L363`](https://github.com/aws-greengrass/aws-greengrass-logging-java/blob/main/src/main/java/com/aws/greengrass/logging/impl/config/PersistenceConfig.java#L353-L363)) provides primary disk protection (`totalSizeCap=10MB`) for stdout/stderr logs. But on GG Lite there's no Logback — components log via [systemd/journald only](https://github.com/aws-greengrass/aws-greengrass-lite/blob/main/modules/recipe2unit/src/unit_file_generator.c#L389-L393), and custom `logFileDirectoryPath` files have zero platform-level protection.

Will revert to the Java behavior for now

README.md
paabbs: Will fix to match exact java behavior in the coming revision

src/main.rs
paabbs: Yes it was intentional, today we delay the following uploads.`fixedDelay` semantics (sleep after work completes). This matches the current Java implementation which uses `while { processLogsAndUpload(); sleepFractionalSeconds(periodicUpdateIntervalSec); }` ([LogManagerService.java#L831](https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/main/src/main/java/com/aws/greengrass/logmanager/LogManagerService.java#L831)).

src/uploader/mod.rs
paabbs: Mkaes sense, will refactor

src/main.rs
paabbs: This was a miss, thanks for the catch!

src/main.rs
paabbs: Replaced those comments with better descriptions, thanks for the callout!

src/main.rs
paabbs: Moving it to the scanner module would create circular dependencies (it orchestrates upload + checkpoint + disk, not scanning), but the modularity concern is valid.

One suggestion is to refactor to a LogManager struct that owns client, store, thing_name, and checkpoint_path. Methods access fields directly. Rust's borrow checker allows disjoint field borrows within &mut self methods. This eliminates the parameter passing entirely. But this should be a separate PR

src/config/mod.rs
```diff
@@ -28,23 +29,8 @@ pub fn load_config(config_arg: &str) -> Result<LogManagerConfig, ConfigError> {
 
 #[must_use = "config result must be handled"]
 pub fn parse_config(json: &str) -> Result<LogManagerConfig, ConfigError> {
-    let mut config: LogManagerConfig =
+    let config: LogManagerConfig =
         serde_json::from_str(json).map_err(|e| ConfigError::Parse(format!("{e}")))?;
-    // Deduplicate by component name, last-wins.
```
utsavvyas, line 33: The only reason for change is to match the Java LogManager input ?

src/disk/mod.rs
```diff
@@ -161,7 +161,12 @@ mod tests {
         set_mtime(&processed, 100);
 
         // total=400, limit=250 → need 150 freed, but only processed is deletable
-        let deleted = free_disk_space(tmp.path(), &log_pattern(), 250, &[processed.clone()]);
+        let deleted = free_disk_space(
```
utsavvyas, line 164: nit: Why there are format changes ?

src/scanner/checkpoint.rs
```diff
@@ -46,29 +46,66 @@ pub struct LastFileProcessedTimestamp {
     pub last_file_processed_time_stamp: u64,
 }
 
+/// Deprecated format: per-file checkpoint (flat: one entry per component, not nested by hash).
+/// Compatible with Java LogManager versions ≤2.3.0. Disable via deprecatedVersionSupport=false.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
```
utsavvyas, line 51: Can we move this to later PRs ?

utsavvyas: Now this is changes almost whole log manager pipeline.

src/uploader/cw_client.rs
```diff
@@ -70,6 +85,38 @@ impl CwLogsClient {
...
+    /// # Errors
+    /// Returns `Err(CwUploadError::AuthError)` if credentials are invalid (non-retriable).
+    pub async fn upload_batch_with_retry(
+        &mut self,
+        batch: &SealedBatch,
+    ) -> Result<UploadOutcome, CwUploadError> {
+        for (attempt, delay) in BACKOFF_DELAYS.iter().enumerate() {
+            match self.upload_batch(batch).await {
+                Ok(()) => return Ok(UploadOutcome::Success),
+                Err(CwUploadError::AuthError) => {
```
utsavvyas, line 100: Can we simply have retryable and non retryable to make it simpler

README.md
```diff
@@ -1,10 +1,66 @@
...
+Supports `deleteLogFileAfterCloudUpload` for automatic cleanup of successfully uploaded files.
+
+### Sample Configuration
+
+```json
+{
+  "logsUploaderConfiguration": {
+    "systemLogsConfiguration": {
+      "uploadToCloudWatch": "true",
+      "logFileDirectoryPath": "/greengrass/v2/logs",
```
aws-kevinrickard, line 31: This sample tells users to put `logFileDirectoryPath` and `logFileRegex` directly inside `systemLogsConfiguration`. **Java's `aws.greengrass.LogManager` does not accept these fields here** — `LogManagerService.processConfiguration` only reads `uploadToCloudWatch`, `multiLineStartPattern`, `minimumLogLevel`, `diskSpaceLimit`, `diskSpaceLimitUnit`, and `deleteLogFileAfterCloudUpload` from the system block. The directory and regex are auto-resolved from the Nucleus's own logging configuration (`services.<NucleusName>.configuration.logging.outputDirectory` and the hardcoded `greengrass.log` filename) at the moment LogManager processes its config.

Two consequences of teaching this shape to users:

1. **It diverges from the [public AWS docs](https://docs.aws.amazon.com/greengrass/v2/developerguide/log-manager-component.html) and from any existing Java-LogManager config in the field.** Operators copy-pasting from those sources, or migrating from Java, get a config that parses cleanly and silently drifts from what they expect.
2. **It introduces a cross-component drift hazard.** If Nucleus's `logging.outputDirectory` is ever changed (deployment, override, future default change), there's no longer a single source of truth — operators have to remember to keep `systemLogsConfiguration.logFileDirectoryPath` in sync by hand. Java's design avoided this by construction.

The right behavior is to drop `logFileDirectoryPath` and `logFileRegex` from `systemLogsConfiguration` (matching Java's schema) and resolve those values at runtime via Greengrass IPC `GetConfiguration`:

- **Classic:** read nucleus component's configuration `logging.outputDirectory`; default to nucleus `<rootPath>/logs` when unset; regex defaults to `greengrass\w*\.log` (Java's `DEFAULT_FILE_REGEX`).
- **Lite:** the Nucleus-component config tree may not exist or may be shaped differently. The IPC call is expected to be absent/fail; the implementation should detect that and fall back to a Lite-appropriate default — or skip cleanly with a clear log explaining "no Nucleus log directory discoverable on this runtime."

Once the schema matches Java's, this sample becomes:

json
"systemLogsConfiguration": {
 "uploadToCloudWatch": "true",
 "minimumLogLevel": "INFO",
 "diskSpaceLimit": "25",
 "diskSpaceLimitUnit": "MB"
}

— which is what the public docs already show, and which works on any standard install without the user having to know where Nucleus is currently writing logs.

src/config/schema.rs
```diff
@@ -73,52 +74,196 @@ impl DiskSpaceLimitUnit {
...
+    #[serde(default, deserialize_with = "deserialize_optional_string_or_number")]
     pub disk_space_limit: Option<String>,
     #[serde(default)]
     pub disk_space_limit_unit: DiskSpaceLimitUnit,
-    #[serde(default)]
+    #[serde(default, deserialize_with = "deserialize_bool_or_string")]
     pub delete_log_file_after_cloud_upload: bool,
     pub multi_line_start_pattern: Option<String>,
+    /// Per-source upload interval override. Not yet wired — all sources use the global
+    /// periodicUploadIntervalSec. Per-source timing requires tracking last-upload-time per source.
```
aws-kevinrickard, line 91: what is the plan for this? Currently it's not wired so if a customer tries to use it they will get silent behavior drift

In the mean time we should at least log a warning when this is set and put a TODO to finish implementation

README.md
```diff
@@ -1,10 +1,66 @@
...
+
+**Checkpoint Persistence** — Tracks per-file read offsets using content-hash-based file identity (survives
+log rotation). Atomic checkpoint writes with fsync for crash safety. Supports both current and deprecated
+checkpoint formats (controlled by `deprecatedVersionSupport` configuration).
+
+**Disk Space Management** — Enforces per-source disk limits by deleting oldest fully-uploaded files
+when the configured `diskSpaceLimit` is exceeded after a successful upload cycle.
+Supports `deleteLogFileAfterCloudUpload` for automatic cleanup of successfully uploaded files.
+
+### Sample Configuration
```
aws-kevinrickard, line 24: Will we support updateToTlogIntervalSec ? I don't see it in this PR

README.md
```diff
@@ -1,10 +1,66 @@
...
+      }
+    }
+  },
+  "periodicUploadIntervalSec": "300"
+}
+```
+
+### GG Lite Note
+
+On GG Lite devices, TES requires an explicit port configuration in `/etc/greengrass/config.yaml`:
```
aws-kevinrickard, line 54: Why is this required? It should not be, this configuration is an output rather than an input on lite, so setting it to something should do nothing because it is overwritten when the port is determined. On lite, the port can not be chosen by the customer

src/config/mod.rs
paabbs: Yeah for the backward compatibility

README.md
paabbs: Thanks for the catch, I can track it as a fast followup for this PR

aws-kevinrickard: Please add TODOs in the code for anything which is being saved for later, so that we don't lose track of it

paabbs:
Closing in favor of a smaller, individually reviewable stack on `rust/v2`. First PR: #251 (config schema alignment). The remaining changes — upload subsystem + typed retry, checkpoint backward-compat, orchestration + disk parity, full E2E suite, and IPC config + recipe + deployment docs — will follow as separate PRs. All review feedback here is being carried into the corresponding stacked PRs; this discussion remains the reference for it.

PR #251, opened 2026-06-16
feat(config): convert componentLogsConfiguration list to keyed map

paabbs:
## Summary

Restructure component configuration to the official `aws.greengrass.LogManager`
schema so configs use the documented shape and existing deployments continue to
parse unchanged.

## What changed

- Nest configuration under `logsUploaderConfiguration` wrapper
- `componentLogsConfiguration` (list) → `componentLogsConfigurationMap` (map keyed
  by component name); legacy list still accepted via serde alias
- `systemLogsConfiguration`: list → single object
- `ConfigError`: derive `thiserror::Error` + `#[non_exhaustive]`; add
  `Json(#[from] serde_json::Error)` preserving the error chain
- `deserialize_bool_or_string`: trim whitespace, warn on unrecognized values,
  default to `false` (tolerant — a generic-type component cannot safely reject
  config on non-critical fields)
- `periodicUploadIntervalSec` warns and falls back to 300s on an invalid or
  non-positive value (matching `aws.greengrass.LogManager`)
- Structured startup log with component names, system-config presence, and interval
- `logFileDirectoryPath`/`logFileRegex` optional (empty = "not provided"; runtime
  derivation lands with the scan pipeline)

## How tested

- `cargo build` · `cargo fmt --check` · `cargo test` · `cargo check --no-default-features` — all green
- Pre-existing clippy findings in `src/uploader/` on `rust/v2` are unrelated

## Deferred

- Wire per-source `uploadIntervalSec` (plain `// TODO`)
- Runtime derivation of empty path/regex defaults (scan pipeline)


tests/config_integration_test.rs
```diff
@@ -1,233 +0,0 @@
-// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
-// SPDX-License-Identifier: Apache-2.0
-
-//! Integration tests for config parsing with full LLD J.5 config
-
-use gg_log_manager::config::{
```
utsavvyas, line 6: Why are we removing the integ tests ?

tests/config_test.rs
```diff
@@ -31,40 +29,51 @@ const TEST_CONFIG_JSON: &str = r#"{
...
-    ]
+    },
+    "systemLogsConfiguration": {
+        "logFileDirectoryPath": "/tmp",
+        "logFileRegex": "syslog.*",
+        "logGroupName": "/aws/greengrass/system/syslog",
+        "diskSpaceLimit": "50",
+        "diskSpaceLimitUnit": "GB",
+        "minimumLogLevel": "WARN",
+        "uploadIntervalSec": 120
```
utsavvyas: Why do we have uploadIntervalSec here ?

tests/config_test.rs
```diff
@@ -31,40 +29,51 @@ const TEST_CONFIG_JSON: &str = r#"{
...
 }"#;
 
 #[test]
 fn test_parse_config_fields() {
     let config = parse_config(TEST_CONFIG_JSON).expect("Should parse valid config");
 
     assert_eq!(config.periodic_upload_interval_sec, 300);
-    assert_eq!(config.component_logs_configuration.len(), 3);
-    assert_eq!(config.system_logs_configuration.len(), 1);
+    assert_eq!(
```
utsavvyas: Can we not change the formatting in this PR ?

src/config/schema.rs
```diff
@@ -208,55 +367,110 @@ mod tests {
         assert_eq!(unit, DiskSpaceLimitUnit::GB);
     }
 
-    /// Verify that the JSON wire format is unchanged — existing Java configs must parse.
+    /// Verify that the new map format parses correctly.
```
utsavvyas: Can we not use new map ?

src/config/schema.rs
```diff
@@ -73,52 +74,189 @@ impl DiskSpaceLimitUnit {
...
+/// Accepts a JSON boolean (`true`) or a string (`"true"`/`"false"`). The Greengrass
+/// Nucleus passes config values as strings, so users may supply either form.
+fn deserialize_bool_or_string<'de, D>(deserializer: D) -> Result<bool, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let value = serde_json::Value::deserialize(deserializer)?;
+    match &value {
+        serde_json::Value::Bool(b) => Ok(*b),
+        serde_json::Value::String(s) => Ok(s.eq_ignore_ascii_case("true")),
```
utsavvyas: **[non-blocking]** `deserialize_bool_or_string` maps any unrecognized string and any non-string/non-bool value to `false` silently. `"true"`/`"false"` work as intended, but `"yes"`, `"1"`, a numeric `1`, or `" true"` with stray whitespace all become `false` with no signal.

Since this controls `deleteLogFileAfterCloudUpload` (and `uploadToCloudWatch`), a silent wrong-default could mean logs are unexpectedly retained (disk fills) or deleted. Suggest trimming the string and emitting a `tracing::warn!` when the value isn't a recognized boolean form, so operators can spot the misconfig.

Hard-erroring is an option but may reject inputs that parse today on the Java/v1 side — a warn is safer for backward compat.

src/config/schema.rs
```diff
@@ -73,52 +74,189 @@ impl DiskSpaceLimitUnit {
...
 }
 
-fn default_periodic_interval() -> u64 {
-    DEFAULT_UPLOAD_INTERVAL_SEC
+impl<'de> Deserialize<'de> for LogManagerConfig {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let value = serde_json::Value::deserialize(deserializer)?;
```
alter-mage: Is there a reason the custom `Deserialize` impl uses `serde_json::Value` as an intermediate? This couples the deserialization to JSON specifically — would this ever need to work with a non-JSON format (e.g., TOML from a config file), or is JSON-only intentional here?

src/config/schema.rs
```diff
@@ -73,52 +74,189 @@ impl DiskSpaceLimitUnit {
...
+                    .or_else(|| v.as_str().and_then(|s| s.parse().ok()))
+            })
+            .unwrap_or(DEFAULT_UPLOAD_INTERVAL_SEC);
+
+        // Accept the nested `logsUploaderConfiguration` wrapper (official schema) or a
+        // flat top-level object (recipe interpolation flattens the subtree).
+        let uploader_config = if let Some(inner) = obj.get("logsUploaderConfiguration") {
+            serde_json::from_value(inner.clone()).map_err(serde::de::Error::custom)?
+        } else {
+            serde_json::from_value(value.clone()).map_err(serde::de::Error::custom)?
```
alter-mage: Would it be worth avoiding the `.clone()` here by using `obj.remove("logsUploaderConfiguration")` and consuming the `Value` directly into `from_value`? Not a performance concern at startup, just wondering if there was a reason ownership isn't transferred.

src/config/schema.rs
```diff
@@ -73,52 +74,189 @@ impl DiskSpaceLimitUnit {
...
+/// Nucleus passes config values as strings, so users may supply either form.
+fn deserialize_bool_or_string<'de, D>(deserializer: D) -> Result<bool, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let value = serde_json::Value::deserialize(deserializer)?;
+    match &value {
+        serde_json::Value::Bool(b) => Ok(*b),
+        serde_json::Value::String(s) => Ok(s.eq_ignore_ascii_case("true")),
+        _ => Ok(false),
```
alter-mage: What should happen when `deleteLogFileAfterCloudUpload` is set to an unexpected type like a number or array? Currently the `_ => Ok(false)` arm silently treats it as `false` — is that intentional GG Nucleus tolerance, or would returning a deserialization error be safer here?

src/config/schema.rs
```diff
@@ -73,52 +74,189 @@ impl DiskSpaceLimitUnit {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct LogSourceConfig {
+    #[serde(default)]
     pub log_file_directory_path: String,
```
alter-mage, line 61: With `#[serde(default)]` on `log_file_directory_path` and `log_file_regex`, a config entry like `"MyApp": {}` now parses with both set to `""`. An empty regex matches everything, and an empty directory path would scan cwd.

The docs show `logFileDirectoryPath` as optional (Java derives it from the Nucleus log config at runtime), so the default makes sense for deserialization — but is there a plan to fill in the Nucleus log directory as a fallback before scanning? If so, a brief comment here would help. If not, might be worth rejecting empty strings in `validate_config()`.

src/config/schema.rs
```diff
@@ -73,52 +74,189 @@ impl DiskSpaceLimitUnit {
...
+                    },
+                );
+            }
+            Ok(map)
+        }
+    }
+}
+
+// Keep these type aliases so modules that reference the prior names do not churn.
+pub type ComponentLogConfig = ComponentSourceConfig;
```
alter-mage: nit: These aliases don't appear to have any consumers in-tree. If no parallel PRs reference the old names, consider removing them to keep the public API surface clean. Happy to leave if there's a reason to keep them.

src/config/schema.rs
```diff
@@ -73,52 +74,189 @@ impl DiskSpaceLimitUnit {
...
 }
 
+/// Component source config — the value in the `componentLogsConfigurationMap`.
+/// Does NOT contain `componentName`; that is supplied by the map key.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct ComponentLogConfig {
-    pub component_name: String,
+pub struct ComponentSourceConfig {
     #[serde(flatten)]
```
alter-mage, line 158: Are you aware of the serde `flatten` + `deny_unknown_fields` interaction? If strict config validation is ever added, `flatten` can cause unexpected "unknown field" errors for fields that actually belong to the flattened struct. Just flagging in case that is a future plan.

src/config/mod.rs
```diff
@@ -58,19 +54,33 @@ fn validate_log_source(source: &LogSourceConfig) -> Result<(), ConfigError> {
...
-        .iter()
-        .try_for_each(|c| {
-            tracing::debug!(component = %c.component_name, dir = %c.source.log_file_directory_path, "Validating component config");
-            validate_log_source(&c.source)
-        })?;
-    config.system_logs_configuration.iter().try_for_each(|s| {
-        tracing::debug!(dir = %s.source.log_file_directory_path, "Validating system log config");
-        validate_log_source(&s.source)
-    })?;
+    for (name, comp) in &config
```
alter-mage, line 69: Since `HashMap` has non-deterministic iteration order, is there a concern about which component's validation error gets surfaced first when multiple components have invalid configs? Would a `BTreeMap` make error reporting more predictable, or is first-error-wins acceptable here?

alter-mage (commented):
Review covering config schema refactoring and Rust code quality. All comments are questions — happy to discuss.

**General question on `ConfigError`:** Since `thiserror` is already in `Cargo.toml`, is there a reason `ConfigError` uses a manual `impl Display` + `impl Error` rather than deriving with `#[derive(thiserror::Error)]`? The `Parse` variant currently stringifies the serde error — would wrapping `serde_json::Error` directly (with `#[from]`) preserve the error chain for callers?

src/main.rs
```diff
@@ -38,8 +38,16 @@ async fn main() {
 
     info!(
         "Loaded {} component configs, {} system configs",
-        config.component_logs_configuration.len(),
-        config.system_logs_configuration.len()
+        config
+            .logs_uploader_configuration
+            .component_logs_configuration_map
+            .len(),
+        usize::from(
```
alter-mage: Minor: is there a readability preference between `usize::from(x.is_some())` vs `if x.is_some() { 1 } else { 0 }` or just using `{}` with the bool directly in the log? The `usize::from` conversion isn't immediately obvious at a glance.

tests/config_integration_test.rs
paabbs: This integration test was written against the previous **list-shaped** API — it parsed `componentLogsConfiguration: [ … ]` and indexed `config.component_logs_configuration[0]` / `system_logs_configuration[0]`. After realigning to the documented `componentLogsConfigurationMap` schema, those list-indexed assertions no longer compile. The coverage it provided (multi-component parsing, per-field assertions, defaults, the per-source upload-interval override, system-log config, and derived log-group names) is preserved in the rewritten `tests/config_test.rs`, which is itself an integration-level test crate under `tests/`. Happy to add a dedicated map-shaped end-to-end config fixture if you'd like one separate from those cases

tests/config_test.rs
paabbs: Thanks for the catch, will remove them from this version

tests/config_test.rs
paabbs: The wrapping is just `rustfmt`, the new `componentLogsConfigurationMap` path is long enough that it auto-wraps (a one-liner fails `cargo fmt --check`). I'll bind it to a local so the asserts stay on one line:
```
 let uploader = &config.logs_uploader_configuration;
 assert_eq!(uploader.component_logs_configuration_map.len(), 3);
 assert!(uploader.system_logs_configuration.is_some());
 ```

File: src/config/schema.rs (hunk 4 above)
Comment (paabbs) [reply src/config/schema.rs]: Will change the doc string

File: src/config/schema.rs (hunk 5 above)
Comment (paabbs) [reply src/config/schema.rs]: Good catch, I'll trim the value and warn! on anything that isn't a recognized boolean, still defaulting to false. Note it can only ever resolve to false (never silently true), so it can't trigger unexpected deletion; the real benefit is catching a typo'd uploadToCloudWatch that would otherwise silently stop logs from shipping.

**Comment by paabbs** [comment]
> Review covering config schema refactoring and Rust code quality. All comments are questions — happy to discuss.
> 
> **General question on `ConfigError`:** Since `thiserror` is already in `Cargo.toml`, is there a reason `ConfigError` uses a manual `impl Display` + `impl Error` rather than deriving with `#[derive(thiserror::Error)]`? The `Parse` variant currently stringifies the serde error — would wrapping `serde_json::Error` directly (with `#[from]`) preserve the error chain for callers?

No good reason to keep it manual, I'll switch ConfigError to #[derive(thiserror::Error)] + #[non_exhaustive] (consistent with the uploader's error type). And yes, #[from] preserves source(): since Parse is used both for serde errors and our own diskSpaceLimit messages, it makes sense to add a dedicated Json(#[from] serde_json::Error) variant for the parse path and keep Parse/Validation for our messages.

File: src/config/schema.rs (hunk 6 above)
Comment (paabbs) [reply src/config/schema.rs]: Yes — JSON-only is intentional. A Greengrass component's configuration always arrives as JSON (inline via `--config`, a JSON file, or the component SDK's `GetConfiguration` IPC); there's no TOML/YAML path. The `serde_json::Value` intermediate is what lets us accept both the nested `logsUploaderConfiguration` wrapper and a flat top-level form, and coerce string-or-native scalars like `periodicUploadIntervalSec` (`"300"` or `300`). A plain derive wouldn't handle those two cases cleanly.

File: src/config/schema.rs (hunk 7 above)
Comment (paabbs) [reply src/config/schema.rs]: That's useful. It's only because `obj` is an immutable borrow of `value` (via `as_object()`), so `from_value` (which takes ownership) needed an owned copy. I'll switch to `value.as_object_mut()` + `remove("logsUploaderConfiguration")` to consume the nested subtree directly, and move `value` in the flat-form branch — dropping both clones (this one and the `value.clone()` in the `else`), while keeping the "expected object" guard.

File: src/config/schema.rs (hunk 8 above)
Comment (paabbs) [reply src/config/schema.rs]: Intentional tolerance — and `_ => false` actually matches the Java component for this field: it coerces `deleteLogFileAfterCloudUpload` by stringifying the value first, so a number/array/etc. resolves to `false` there too. Java never errors on an unexpected type (it always coerces to a default of `false`), so hard-erroring would reject configs that parse today. I'll add a `warn!` on unrecognized values for visibility but keep the `false` default — no native-number special-casing needed here.

File: src/config/schema.rs (hunk 9 above)
Comment (paabbs) [reply src/config/schema.rs:61]: There is a plan, so I'll go with the comment rather than rejecting. The Java component treats both fields as optional and *derives* them when unset — regex → `^<componentName>\w*.log`, directory → the component's logger dir, else the root log dir — and treats empty as "not provided" rather than rejecting. The docs also list both as optional (not needed for components that log to stdout/stderr), so rejecting empty in `validate_config()` would break those cases. Our `#[serde(default)]` empty string is that "not provided" placeholder; the derivation is applied when the scan pipeline is wired into the run loop, so it's out of scope for this config-parsing PR. I'll add a comment here so the empty default isn't read as "match everything."

File: src/config/schema.rs (hunk 10 above)
Comment (paabbs) [reply src/config/schema.rs]: Makes sense, they have no in-tree consumers; they were left over from the struct rename. I'll remove both aliases and their re-export.

File: src/config/schema.rs (hunk 11 above)
Comment (paabbs) [reply src/config/schema.rs:158]: Aware of it, thanks. We don't use `deny_unknown_fields` and have no plan for strict config validation — unknown keys are tolerated on purpose. This component is a generic type (standalone process) targeting both GG Classic and Lite; on both runtimes the config arrives as a JSON blob via `--config` or IPC, and neither runtime enforces a closed schema. Tolerating unknown keys keeps forward-compatibility (a newer deployment can add config keys an older binary ignores gracefully). If strict validation is ever needed, we'll account for the flatten interaction then.

File: src/config/mod.rs (hunk 12 above)
Comment (paabbs) [reply src/config/mod.rs:69]: First-error-wins is acceptable — validation fails fast and the component is rejected regardless. The Java component also uses an unordered map (`ConcurrentHashMap`) with no ordering contract. Happy to switch to `BTreeMap` for reproducible diagnostics if you feel strongly, but it wouldn't change behavior.

File: src/main.rs (hunk 13 above)
Comment (paabbs) [reply src/main.rs]: Agreed, it's not clear. I'll switch to structured tracing fields — log the component names directly and the system-config presence as a bool, rather than coercing to a count.

**Review by utsavvyas — APPROVED** [review]

File: src/config/schema.rs [hunk 14]
```diff
@@ -73,50 +54,204 @@ impl DiskSpaceLimitUnit {
...
+            .as_object()
+            .ok_or_else(|| serde::de::Error::custom("expected object"))?;
+
+        let periodic = obj
+            .get("periodicUploadIntervalSec")
+            .and_then(|v| {
+                v.as_u64()
+                    .or_else(|| v.as_str().and_then(|s| s.parse().ok()))
+            })
+            .unwrap_or(DEFAULT_UPLOAD_INTERVAL_SEC);
```
alter-mage: Should this warn when `periodicUploadIntervalSec` is present but unparseable? Currently `"periodicUploadIntervalSec": "30o"` (typo) silently falls back to 300s — the user wouldn't know their value was ignored. The bool deserializer now warns on unrecognized values; this could get the same treatment for consistency.

paabbs: Makes sense. I'll mirror the functionality of the Java Log manager: trim the string and warn! (with the offending value) on an unparseable or non-positive interval, keeping the 300s default. That also closes a gap where a literal 0 currently parses through instead of falling back.

alter-mage (approved):

PR #252, opened 2026-07-01
feat(uploader): add typed upload subsystem and thing-name resolution

paabbs:
## Summary
Adds the CloudWatch Logs upload subsystem and thing-name resolution: typed failure classification and retry, the upload-pipeline API the orchestrator calls, and credential/thing-name handling. Groundwork for the orchestration change that stacks on top of this.

## What changed
- **Typed upload results** — classify CloudWatch failures as `Auth` / `Retryable` / `Fatal` (`CwUploadError`, `#[non_exhaustive]`); add `#[must_use]` `UploadOutcome { Success, RetriesExhausted }`. The AWS SDK already retries transient faults, so the app layer re-attempts **exactly once** for a missing log group/stream, with no app-level backoff sleep.
- **Upload-pipeline API** — `upload_source_events`, `advance_checkpoints`, `evict_stale_entries` (24h TTL), jittered `effective_interval_secs`, `mtime_to_ms`. `upload_source_events` returns `Result` so **auth failures surface as `Err(Auth)`** for the caller to refresh credentials before retrying; other failures stay per-source, and a source's checkpoints advance only when **every** batch for it succeeds.
- **Checkpoint advancement** — `advance_checkpoints` resolves each file's state (active-flag, mtime) **by path**, so it always matches the file whose bytes were read; this avoids mis-attributing state to a different file that shares a first-line content hash. The checkpoint map stays keyed by `content_hash`. `advance`/`evict` take an injected `now` for deterministic tests.
- **Thing-name resolution** — via an injected env lookup, logging an `error!` when unset/empty; removes the empty TES credential stub.
- Misc: `div_ceil` chunk sizing; `seal_batches`/`MAX_EVENT_BYTES` scoped to `pub(crate)` with batcher internals gated behind the `aws-sdk` feature; `config/schema.rs` test comment put back on its own line.

## How tested
- `cargo test` — 209 passed, 0 failed
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo fmt --check` — clean
- `cargo check --no-default-features --all-targets` — clean

## Notes
- `CwUploadError` is `pub` so the follow-up orchestration PR (which wires `main.rs`) can match on `Err(Auth)` to stop the cycle and refresh credentials; other uploader internals are `pub(crate)`. The orchestration PR stacks directly on this branch.
- First of the upload split; the orchestration PR follows. Base: `rust/v2`.

### DIFF
--- a/src/config/schema.rs
+++ b/src/config/schema.rs
@@ -502,6 +502,7 @@ mod tests {
         assert!(!f(r#""yes""#));
         assert!(!f(r#""1""#));
         assert!(!f(r#""""#)); // empty → false
+
         // Unexpected types → false (with warn)
         assert!(!f("1"));
         assert!(!f("null"));
--- a/src/credentials/mod.rs
+++ b/src/credentials/mod.rs
@@ -1,6 +1,67 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Credential provider and environment detection for Greengrass Nucleus and Nucleus Lite
+//! ThingName resolution for Greengrass.
+//!
+//! Credentials are handled by the AWS SDK default chain (`aws_config::load_defaults`),
+//! which picks up TES-provided credentials automatically when running as a Greengrass
+//! component. Both Classic and Lite set `AWS_IOT_THING_NAME` for component processes.
 
-mod tes;
+/// Resolve the thing name from the `AWS_IOT_THING_NAME` environment variable.
+///
+/// Returns `None` (after logging an error) when the variable is unset or empty, so the
+/// caller can fall back to a placeholder. Both GG Classic and GG Lite set this for
+/// component processes.
+#[must_use]
+pub fn resolve_thing_name() -> Option<String> {
+    resolve_thing_name_from(|key| std::env::var(key))
+}
+
+/// Resolve the thing name using an injected environment lookup.
+///
+/// Split out from [`resolve_thing_name`] so tests can supply a closure instead of
+/// mutating process-global environment state.
+fn resolve_thing_name_from(
+    get_var: impl Fn(&str) -> Result<String, std::env::VarError>,
+) -> Option<String> {
+    match get_var("AWS_IOT_THING_NAME") {
+        Ok(name) if !name.is_empty() => {
+            tracing::info!(thing_name = %name, "Resolved thing name");
+            Some(name)
+        }
+        Ok(_) => {
+            tracing::error!("AWS_IOT_THING_NAME is set but empty");
+            None
+        }
+        Err(_) => {
+            tracing::error!("AWS_IOT_THING_NAME not set");
+            None
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_resolve_thing_name_present() {
+        assert_eq!(
+            resolve_thing_name_from(|_: &str| Ok("test-device-01".to_string())),
+            Some("test-device-01".to_string())
+        );
+    }
+
+    #[test]
+    fn test_resolve_thing_name_missing() {
+        assert_eq!(
+            resolve_thing_name_from(|_: &str| Err(std::env::VarError::NotPresent)),
+            None
+        );
+    }
+
+    #[test]
+    fn test_resolve_thing_name_empty() {
+        assert_eq!(resolve_thing_name_from(|_: &str| Ok(String::new())), None);
+    }
+}
--- a/src/disk/mod.rs
+++ b/src/disk/mod.rs
@@ -33,7 +33,7 @@ fn dir_size(dir: &Path, pattern: &Regex) -> u64 {
 /// (fully uploaded, not actively written to). This function trusts the list it receives.
 ///
 /// Returns paths of successfully deleted files.
-pub(crate) fn free_disk_space(
+pub fn free_disk_space(
     dir: &Path,
     pattern: &Regex,
     limit_bytes: u64,
@@ -93,7 +93,9 @@ pub(crate) fn free_disk_space(
     deleted
 }
 
+// Single-arg &[x.clone()] kept per review; allow for clippy 1.94.1.
 #[cfg(test)]
+#[allow(clippy::cloned_ref_to_slice_refs)]
 mod tests {
     use super::*;
     use std::fs::File;
--- a/src/scanner/mod.rs
+++ b/src/scanner/mod.rs
@@ -9,7 +9,7 @@ mod reader;
 
 pub use checkpoint::{
     load_checkpoint, recover_offsets, remove_component, save_checkpoint, trim_stale_on_load,
-    CheckpointStore, FileCheckpoint,
+    CheckpointStore, FileCheckpoint, LastFileProcessedTimestamp,
 };
 pub use multiline::assemble_multiline;
 pub use reader::{compute_content_hash, read_file_from_offset, LogEvent};
--- a/src/uploader/batcher.rs
+++ b/src/uploader/batcher.rs
@@ -17,7 +17,7 @@
 //! **Dropped events don't advance checkpoint.** Java returns consumed byte count for
 //! dropped events (14-day filter) so the caller advances the file offset past them.
 //! This batcher silently drops them. The orchestrator must filter old events before
-//! batching, or accept re-reading them each cycle. Tracked under Asana 1.8.
+//! batching, or accept re-reading them each cycle.
 //!
 //! **Log level filtering matches Java behavior.** JSON-structured `GreengrassLogMessage`
 //! lines are deserialized to extract the `level` field. Text-format lines pass through
@@ -181,7 +181,7 @@ fn chunk_message(message: &str, timestamp: i64) -> Vec<LogEvent> {
         }];
     }
 
-    let num_chunks = (bytes.len() + MAX_EVENT_BYTES - 1) / MAX_EVENT_BYTES;
+    let num_chunks = bytes.len().div_ceil(MAX_EVENT_BYTES);
     tracing::debug!(
         size = bytes.len(),
         chunks = num_chunks,
@@ -518,8 +518,6 @@ mod tests {
         ));
         assert!(!should_filter_by_level("short", LogLevel::Error));
     }
-
-    // --- Boundary tests (exact-at-limit) ---
 
     #[test]
     fn test_14_day_boundary_exact_kept() {
--- a/src/uploader/cw_client.rs
+++ b/src/uploader/cw_client.rs
@@ -9,18 +9,47 @@ use aws_sdk_cloudwatchlogs::{
 };
 use std::collections::HashSet;
 
+/// Classification of a CloudWatch upload failure.
+///
+/// The AWS SDK already retries transient faults (throttling, 5xx, timeouts)
+/// internally via [`RetryConfig`], so the only fault worth a manual retry
+/// is a missing log group/stream — everything else is either fatal for this cycle or
+/// an auth problem the same credentials cannot fix.
 #[derive(Debug, thiserror::Error)]
-pub(crate) enum CwUploadError {
+#[non_exhaustive]
+pub enum CwUploadError {
+    /// Authentication/authorization failure. Not retryable with the same credentials —
+    /// kept distinct so the caller can trigger a TES credential refresh.
     #[error("authentication error")]
-    AuthError,
-    #[error("retriable: {0}")]
-    Retriable(String),
+    Auth,
+    /// A fault the SDK does not retry but a single immediate re-attempt may clear —
+    /// notably `ResourceNotFoundException` (recreate the group/stream) or a raw
+    /// transport error surfacing after the SDK's own retries are exhausted.
+    #[error("retryable: {0}")]
+    Retryable(String),
+    /// A non-retryable failure for this cycle: invalid input, a hard resource/account
+    /// quota such as `LimitExceeded` that retrying cannot clear, or a transient class
+    /// the SDK already retried and exhausted (e.g. throttling/5xx).
     #[error("{0}")]
-    Other(String),
+    Fatal(String),
 }
 
-pub(crate) struct CwLogsClient {
+/// Outcome of an upload attempt.
+#[derive(Debug)]
+#[must_use = "upload outcome must be handled"]
+pub(crate) enum UploadOutcome {
+    /// All events uploaded successfully.
+    Success,
+    /// A retryable fault persisted after the single re-attempt — the batch will be
+    /// retried on the next upload cycle.
+    RetriesExhausted,
+}
+
+pub struct CwLogsClient {
     client: Client,
+    // Only read by `recreate_client` (the credential-refresh hook), which has no
+    // production caller in this crate yet — exercised by unit tests.
+    #[allow(dead_code)]
     config: aws_sdk_cloudwatchlogs::Config,
     created_groups: HashSet<String>,
     created_streams: HashSet<String>,
@@ -56,6 +85,9 @@ impl CwLogsClient {
     /// Recreate the SDK client (e.g., after persistent network errors).
     /// Caches are preserved — if a resource was deleted externally, the next
     /// put_log_events will get ResourceNotFoundException which clears the cache.
+    // No production caller yet; exercised by unit tests. Retained as the client's
+    // credential-refresh hook for the upload orchestration.
+    #[allow(dead_code)]
     pub(crate) fn recreate_client(&mut self) {
         self.client = Client::from_conf(self.config.clone());
     }
@@ -68,6 +100,47 @@ impl CwLogsClient {
         self.ensure_log_stream(&batch.log_group, &batch.log_stream)
             .await?;
         self.put_log_events(batch).await
+    }
+
+    /// Upload a batch, retrying exactly once for a retryable fault.
+    ///
+    /// The AWS SDK already retries throttling/5xx/timeouts internally
+    /// ([`RetryConfig`] with `max_attempts(5)`), so there is deliberately no app-level
+    /// sleep/backoff loop here. The single re-attempt exists for the one case the SDK
+    /// cannot handle: a missing log group/stream. On `ResourceNotFoundException`,
+    /// [`Self::put_log_events`] clears the resource cache, so re-running `upload_batch`
+    /// recreates the group/stream and re-sends once when the target was deleted
+    /// between the cache check and the upload. Any other persistent failure is
+    /// deferred to the next cycle.
+    ///
+    /// # Errors
+    /// Returns [`CwUploadError::Auth`] when credentials are invalid (non-retryable);
+    /// the caller stops processing the source so credentials can be refreshed.
+    pub(crate) async fn upload_batch_with_retry(
+        &mut self,
+        batch: &SealedBatch,
+    ) -> Result<UploadOutcome, CwUploadError> {
+        match self.upload_batch(batch).await {
+            Ok(()) => return Ok(UploadOutcome::Success),
+            Err(CwUploadError::Auth) => return Err(CwUploadError::Auth),
+            Err(CwUploadError::Fatal(msg)) => {
+                tracing::error!(log_group = %batch.log_group, error = %msg, "Non-retryable upload error");
+                return Ok(UploadOutcome::RetriesExhausted);
+            }
+            Err(CwUploadError::Retryable(msg)) => {
+                tracing::warn!(log_group = %batch.log_group, error = %msg, "Retryable error, recreating resources and retrying once");
+            }
+        }
+
+        // Single re-attempt, no sleep. Persistent failures wait for the next upload cycle.
+        match self.upload_batch(batch).await {
+            Ok(()) => Ok(UploadOutcome::Success),
+            Err(CwUploadError::Auth) => Err(CwUploadError::Auth),
+            Err(e) => {
+                tracing::error!(log_group = %batch.log_group, error = %e, "Upload failed after recreate retry");
+                Ok(UploadOutcome::RetriesExhausted)
+            }
+        }
     }
 
     async fn ensure_log_group(&mut self, log_group: &str) -> Result<(), CwUploadError> {
@@ -91,12 +164,10 @@ impl CwLogsClient {
                 Ok(())
             }
             Err(SdkError::ServiceError(e)) if is_limit_exceeded_group(e.err()) => Err(
-                CwUploadError::Retriable("LimitExceededException on create_log_group".to_string()),
+                CwUploadError::Fatal("LimitExceededException on create_log_group".to_string()),
             ),
-            Err(SdkError::ServiceError(e)) if is_auth_error(e.err()) => {
-                Err(CwUploadError::AuthError)
-            }
-            Err(e) => Err(CwUploadError::Retriable(e.to_string())),
+            Err(SdkError::ServiceError(e)) if is_auth_error(e.err()) => Err(CwUploadError::Auth),
+            Err(e) => Err(CwUploadError::Retryable(e.to_string())),
         }
     }
 
@@ -126,13 +197,8 @@ impl CwLogsClient {
                 self.created_streams.insert(key);
                 Ok(())
             }
-            Err(SdkError::ServiceError(e)) if is_limit_exceeded_stream(e.err()) => Err(
-                CwUploadError::Retriable("LimitExceededException on create_log_stream".to_string()),
-            ),
-            Err(SdkError::ServiceError(e)) if is_auth_error(e.err()) => {
-                Err(CwUploadError::AuthError)
-            }
-            Err(e) => Err(CwUploadError::Retriable(e.to_string())),
+            Err(SdkError::ServiceError(e)) if is_auth_error(e.err()) => Err(CwUploadError::Auth),
+            Err(e) => Err(CwUploadError::Retryable(e.to_string())),
         }
     }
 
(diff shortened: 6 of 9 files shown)


src/disk/mod.rs
```diff
@@ -93,7 +93,9 @@ pub(crate) fn free_disk_space(
     deleted
 }
 
+// Single-arg &[x.clone()] kept per review; allow for clippy 1.94.1.
 #[cfg(test)]
+#[allow(clippy::cloned_ref_to_slice_refs)]
```
utsavvyas, line 98: why do you need this change ?

src/uploader/cw_client.rs
```diff
@@ -56,6 +82,9 @@ impl CwLogsClient {
     /// Recreate the SDK client (e.g., after persistent network errors).
     /// Caches are preserved — if a resource was deleted externally, the next
     /// put_log_events will get ResourceNotFoundException which clears the cache.
+    // No production caller yet; exercised by unit tests. Retained as the client's
+    // credential-refresh hook for the upload orchestration.
+    #[allow(dead_code)]
```
utsavvyas, line 90: This comment is adding no value

src/uploader/cw_client.rs
```diff
@@ -90,13 +160,13 @@ impl CwLogsClient {
                 self.created_groups.insert(log_group.to_string());
                 Ok(())
             }
+            // LimitExceeded is the SDK's job to retry; if it surfaces here the SDK has
+            // already exhausted its attempts, so treat it as fatal for this cycle.
```
utsavvyas: Can we simply omit comments which are providing what information.

src/uploader/cw_client.rs
```diff
@@ -90,13 +160,13 @@ impl CwLogsClient {
...
             Err(SdkError::ServiceError(e)) if is_limit_exceeded_group(e.err()) => Err(
-                CwUploadError::Retriable("LimitExceededException on create_log_group".to_string()),
+                CwUploadError::Fatal("LimitExceededException on create_log_group".to_string()),
             ),
-            Err(SdkError::ServiceError(e)) if is_auth_error(e.err()) => {
-                Err(CwUploadError::AuthError)
-            }
-            Err(e) => Err(CwUploadError::Retriable(e.to_string())),
+            Err(SdkError::ServiceError(e)) if is_auth_error(e.err()) => Err(CwUploadError::Auth),
```
utsavvyas, line 169: I dont want to see the formatting changes.

src/uploader/cw_client.rs
```diff
@@ -127,12 +197,10 @@ impl CwLogsClient {
                 Ok(())
             }
             Err(SdkError::ServiceError(e)) if is_limit_exceeded_stream(e.err()) => Err(
-                CwUploadError::Retriable("LimitExceededException on create_log_stream".to_string()),
+                CwUploadError::Fatal("LimitExceededException on create_log_stream".to_string()),
```
utsavvyas: What was the reason for non retryable limit exceeded exception for the create log stream ? When this can happen and would it be auto recoverable ?

src/uploader/cw_client.rs
```diff
@@ -9,18 +9,44 @@ use aws_sdk_cloudwatchlogs::{
...
+    /// notably `ResourceNotFoundException` (recreate the group/stream) or a raw
+    /// transport error surfacing after the SDK's own retries are exhausted.
+    #[error("retryable: {0}")]
+    Retryable(String),
+    /// A non-retryable failure for this cycle (invalid input, or a transient class the
+    /// SDK already retried and exhausted, e.g. throttling/`LimitExceeded`/5xx).
     #[error("{0}")]
-    Other(String),
+    Fatal(String),
 }
```
utsavvyas, line 35: Do you have strong reason to change the CWUploadError ? I think making this update forces you to raise large PR. Maybe have separate PR for just name update if you really want to do it later.

src/scanner/mod.rs
```diff
@@ -9,7 +9,7 @@ mod reader;
 
 pub use checkpoint::{
     load_checkpoint, recover_offsets, remove_component, save_checkpoint, trim_stale_on_load,
-    CheckpointStore, FileCheckpoint,
+    CheckpointStore, FileCheckpoint, LastFileProcessedTimestamp,
```
utsavvyas, line 12: Why this is needed ?

src/uploader/batcher.rs
```diff
@@ -34,7 +34,7 @@ const TIMESTAMP_BYTES: usize = 8;
 /// CW PutLogEvents max payload: 1 MB.
 const MAX_BATCH_SIZE: usize = 1_048_576;
 /// CW max message bytes per event: 256KB - 8 - 26 = 262,110.
-pub(crate) const MAX_EVENT_BYTES: usize = 256 * 1024 - TIMESTAMP_BYTES - EVENT_STORAGE_OVERHEAD;
+pub const MAX_EVENT_BYTES: usize = 256 * 1024 - TIMESTAMP_BYTES - EVENT_STORAGE_OVERHEAD;
```
utsavvyas: Should it be pub ?

src/uploader/mod.rs
```diff
@@ -1,34 +1,237 @@
...
+        }
+    } else {
+        UploadResult {
+            succeeded: vec![],
+            failed: file_info.into_iter().map(|(p, _, _)| p).collect(),
+        }
+    }
+}
+
+// --- Checkpoint advancement ---
```
utsavvyas: This comment seems weird. If we have to add comment there is an opportunity to refactor it or this comment adds no value.

src/uploader/mod.rs
```diff
@@ -1,34 +1,237 @@
...
+
+    if all_ok {
+        UploadResult {
+            succeeded: file_info,
+            failed: vec![],
+        }
+    } else {
+        UploadResult {
+            succeeded: vec![],
+            failed: file_info.into_iter().map(|(p, _, _)| p).collect(),
```
utsavvyas, line 133: What will happen if X number of lines are uploaded successfully but not all for any particular file.

src/uploader/mod.rs
```diff
@@ -1,34 +1,237 @@
...
 
+/// Upload events from multiple files for a single log source.
+///
+/// `seal_batches` merges events across files (sorted by timestamp), so per-file success
+/// cannot be tracked at batch granularity. We therefore use an all-or-nothing model for
+/// the source: a file's checkpoint advances only if EVERY batch succeeds; if any batch
+/// fails, all files for this source are marked failed and retried on the next cycle.
 #[cfg(feature = "aws-sdk")]
-pub(crate) use cw_client::{CwLogsClient, CwUploadError};
+pub async fn upload_source_events(
```
utsavvyas, line 83: `upload_source_events` is the most complex function added in this PR but has no direct test.

The leaf helpers (`effective_interval_secs`, `advance_checkpoints`, `is_file_fully_uploaded`, `evict_stale_entries`) and the `cw_client` retry paths are covered, but the source-level orchestration branches here are not exercised:
- empty-events early return
- all-batches-succeed path (populating `succeeded`)
- partial failure: a later batch returns `RetriesExhausted`/`Auth` after an earlier batch already succeeded

Can we add a test that drives a multi-batch source through the replay client where batch N fails? That's the path where the all-or-nothing offset behavior actually matters — and where events in already-succeeded batches get re-read and re-sent on the next cycle.

src/disk/mod.rs
paabbs: This is just for linting purposes, can be removed if deemed strictly unnecessary

src/uploader/cw_client.rs
paabbs: Makes sense. More of a reviewer note to be honest, but I could have just added to the PR notes

src/uploader/cw_client.rs
paabbs: Sure, I can remove the obvious "what"s

src/uploader/cw_client.rs
paabbs: We have this pipeline established before publishing a PR to remove that's where these are coming from 
```
cargo test 
cargo clippy --all-targets -- -D warnings 
cargo fmt --check 
cargo check --no-default-features --all-targets 
```

src/uploader/cw_client.rs
paabbs: `LimitExceededException` is a **hard resource/account quota**, not a rate limit — CloudWatch Logs rate limiting surfaces as `ThrottlingException` (which the SDK retries), so a quota can't be cleared by retrying → non-retryable.

For `create_log_stream` specifically it effectively **can't happen**. Per the [CreateLogStream API reference](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_CreateLogStream.html):

> There is no limit on the number of log streams that you can create for a log group. There is a limit of 50 TPS on `CreateLogStream` operations, after which transactions are throttled.

and its documented errors are only `InvalidParameterException`, `ResourceAlreadyExistsException`, `ResourceNotFoundException`, and `ServiceUnavailableException` — **no `LimitExceededException`**. That is also why the SDK exposes no typed `LimitExceededException` variant for `CreateLogStreamError` (the code string-matches it defensively).

**Action:** I'll remove this branch for `create_log_stream`. The genuine quota case exists only on `create_log_group` (which has a typed `LimitExceededException` variant), where it stays `Fatal` / non-retryable — a quota needs an increase or cleanup, not a retry.

src/uploader/cw_client.rs
utsavvyas: If that is the case just one PR which has all the formatting changes

src/uploader/cw_client.rs
paabbs: It's not a cosmetic rename, the variants encode the retry contract this PR introduces, so the old names would misdescribe the new behavior:

- `Retriable` → `Retryable(String)`: the contract narrowed. It no longer implies an app-level backoff loop; it now marks the single case worth **one immediate re-attempt** — `ResourceNotFoundException`, where we recreate the log group/stream and retry once.
- `Other` → `Fatal(String)`: now an explicit "non-retryable this cycle" class that surfaces as `UploadOutcome::RetriesExhausted` and defers to the next upload cycle. It covers hard quotas that retrying can't clear and faults the SDK already exhausted its own retry budget on. `Other` didn't convey that role.
- `AuthError` → `Auth`: this one is purely cosmetic (drops the redundant `Error` suffix).

`Retryable`/`Fatal` are tied to the behavioral reclassification that is the substance of this PR so splitting them out would leave the new retry logic referring to `Retriable`/`Other`, or worse, ship a variant named `Fatal` while the code still retries it in a backoff loop. The names and the control flow are one atomic change.

src/scanner/mod.rs
paabbs: The upload subsystem (`src/uploader/mod.rs`) needs to construct `LastFileProcessedTimestamp` values when updating the checkpoint store after a successful upload. The type was already `pub` in `scanner/checkpoint.rs` but wasn't re-exported from the `scanner` module, so `uploader/mod.rs` couldn't access it through the public `crate::scanner::` path without reaching into the internal module structure. This adds it to the existing re-export block for consistency with the other checkpoint types already there (`CheckpointStore`, `FileCheckpoint`).

src/uploader/cw_client.rs
paabbs: Will follow from the next one

src/uploader/batcher.rs
paabbs: That was a miss, will revert this. Thanks for the catch

src/uploader/mod.rs
paabbs: Added them just for the grouping. I can remove them if deemed unnecessary

src/uploader/mod.rs
paabbs: This is an all-or-nothing model per source — a "source" being one configured log source (a component's logs, or the system logs) mapping to one CloudWatch log group. Events from all of a source's files are merged into timestamp-ordered batches for upload, so a single batch can span multiple files and one file can span multiple batches — there's no per-file batch boundary to record a partial offset against.

If any batch fails, the whole source is marked failed and no file's checkpoint advances (checkpoints advance only over the succeeded set, which is empty on failure). The next cycle re-reads each file from its last committed offset and re-sends. So events CloudWatch already accepted before the failure may be re-sent — at-least-once: no data loss, with possible duplicates for the pre-failure portion. (No server-side dedup: per the [PutLogEvents reference](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html), sequence tokens are ignored and `DataAlreadyAcceptedException` is never returned.)

The current design doesn't track partial per-file offsets, since the merged batching gives no per-file "made it this far" point. Finer per-file checkpointing to reduce duplicates on partial failure is possible, but it's a behavior change beyond this PR's scope, better tracked as a follow-up.

src/uploader/mod.rs
paabbs: Makes sense.  The source-level orchestration branches aren't exercised. I'll add a test using the same `StaticReplayClient` harness the existing `cw_client` tests use, covering:

1. **Empty events** — early return (all files reported succeeded, nothing uploaded)
2. **All batches succeed** — `succeeded` populated with all files, `failed` empty
3. **Multi-batch partial failure** — enough events to seal 2+ batches (>10,000 events hits the `MAX_NUM_OF_LOG_EVENTS` split), replay client returns 200 for batch 1 then a failure for batch 2 → whole source marked failed (`succeeded` empty, `failed` = all files)

That third case is the one that pins the all-or-nothing offset behavior from my reply above, already-sent events in batch 1 get re-read and re-sent next cycle.

alter-mage (commented):
Overall this is solid, well-documented code. The retry model is thoughtfully designed and the tests exercise the important failure paths. A few suggestions below.

src/uploader/mod.rs
alter-mage, line 83: Right now, `upload_source_events` returns `UploadResult` which only has `succeeded` and `failed` file lists. If the upload failed because of an auth error (e.g. expired TES credentials), the caller gets the same result as if it failed because CloudWatch was temporarily unavailable.

This matters because the orchestrator will loop over multiple log sources. If credentials are expired, source 1 will fail, then source 2 will fail, then source 3 will fail — all with the same expired token. The orchestrator has no way to know "stop trying, your credentials are bad" vs. "that one source had a blip, keep going with the others."

Suggestion: change the return type to `Result<UploadResult, CwUploadError>` so that auth failures bubble up as `Err(Auth)` and the orchestrator can stop immediately and refresh credentials.

src/uploader/mod.rs
```diff
@@ -1,34 +1,237 @@
...
+        .subsec_nanos() as u64
+        % 6;
+    base + jitter
+}
+
+// --- Upload result tracking ---
+
+/// Result of uploading events for a single log source.
+#[derive(Debug)]
+pub struct UploadResult {
```
alter-mage, line 61: The 3- and 4-element tuples in `UploadResult` and the `file_events` parameter are hard to read at call sites. For example:

```rust
let succeeded = vec![(file_path.clone(), 11u64, "hash1".to_string())];
```

It's not obvious which field is which without checking the type definition. Consider introducing named structs:

```rust
pub struct FileUploadInput {
    pub path: PathBuf,
    pub events: Vec<LogEvent>,
    pub new_offset: u64,
    pub content_hash: String,
}

pub struct FileUploadSuccess {
    pub path: PathBuf,
    pub new_offset: u64,
    pub content_hash: String,
}
```

This also makes it safer to add fields later without touching every call site, and prevents accidental position swaps that the compiler can't catch when adjacent tuple elements share a type.

src/uploader/mod.rs
```diff
@@ -1,34 +1,237 @@
...
+
+    let file_map = store
+        .file_processing_info
+        .entry(log_group_key.to_string())
+        .or_default();
+
+    for (path, new_offset, content_hash) in succeeded {
+        let scanned = scanned_files
+            .iter()
+            .find(|f| &f.content_hash == content_hash);
```
alter-mage: Nit: `scanned_files.iter().find()` inside the loop is O(n×m). Fine at current scale (a source typically has <10 files), but if file counts grow, consider pre-building a `HashMap<&str, &ScannedFile>` keyed by `content_hash` before the loop for O(1) lookups.

src/uploader/mod.rs
```diff
@@ -1,34 +1,237 @@
...
+        % 6;
+    base + jitter
+}
+
+// --- Upload result tracking ---
+
+/// Result of uploading events for a single log source.
+#[derive(Debug)]
+pub struct UploadResult {
+    /// Files whose ALL batches succeeded: (path, new_offset, content_hash).
```
alter-mage, line 62: Consider adding `#[must_use]` to `UploadResult` and `UploadOutcome`. Both represent outcomes that the caller must inspect to advance checkpoints or handle errors. Without it, a forgotten `.await;` (without binding the result) compiles silently.

```rust
#[must_use = "upload results must be checked to advance checkpoints"]
pub struct UploadResult { ... }
```

src/uploader/mod.rs
```diff
@@ -1,34 +1,237 @@
...
+/// A completed file (`!is_active && file_length == new_offset && len > 0`) is removed
+/// from the checkpoint and its component `lastFileProcessedTimeStamp` is advanced
+/// monotonically. A partial or active file simply has its `start_position` updated.
+pub fn advance_checkpoints(
+    store: &mut CheckpointStore,
+    log_group_key: &str,
+    succeeded: &[(PathBuf, u64, String)], // (path, new_offset, content_hash)
+    scanned_files: &[ScannedFile],
+) -> Vec<PathBuf> {
+    let now = now_ms();
```
alter-mage: `advance_checkpoints` and `evict_stale_entries` each call `now_ms()` independently, but they're always invoked back-to-back in the same upload cycle. There's no reason they need different timestamps. Consider having the caller pass `now: u64` into both so they share one consistent value. This also makes tests fully deterministic — no need for "25 hours ago" arithmetic against real wall time.

src/credentials/mod.rs
```diff
@@ -1,6 +1,60 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serial_test::serial;
+
+    #[test]
+    #[serial]
+    fn test_resolve_thing_name_present() {
+        std::env::set_var("AWS_IOT_THING_NAME", "test-device-01");
```
alter-mage: `std::env::set_var` and `remove_var` are deprecated as unsafe since Rust 1.78 (they mutate shared process state without synchronization). `#[serial]` prevents the data race but won't suppress the deprecation warning on newer toolchains.

Two options:

1. Wrap in `unsafe` with a safety comment:
```rust
// SAFETY: #[serial] ensures no concurrent test execution
unsafe { std::env::set_var("AWS_IOT_THING_NAME", "test-device-01") };
```

2. Inject the lookup so tests don't touch process state:
```rust
fn resolve_thing_name_from(
    get_var: impl Fn(&str) -> Result<String, std::env::VarError>
) -> Option<String> { ... }
```

Option 2 is more idiomatic and makes tests deterministic without needing `#[serial]`.

src/config/schema.rs
```diff
@@ -502,7 +502,7 @@ mod tests {
         assert!(!f(r#""yes""#));
         assert!(!f(r#""1""#));
         assert!(!f(r#""""#)); // empty → false
-        // Unexpected types → false (with warn)
+                              // Unexpected types → false (with warn)
```
alter-mage: Nit: this looks like an accidental whitespace change — the comment moved from its own line to being aligned with the inline comment above. Might want to revert so git blame stays clean.

src/uploader/mod.rs
paabbs: Good call — I'll make auth failures distinguishable. I'll change `upload_source_events` to return `Result<UploadResult, CwUploadError>` so an auth failure surfaces as `Err(Auth)`, while transient/per-source failures stay in `UploadResult.failed`. But a point to note is that this is a minor parity although a net new improvement from the Java log manager version.

The classification already exists one layer down: `upload_batch_with_retry` only ever returns `Err` for the auth case — throttling, 5xx, and transient transport errors are retried by the SDK and, once exhausted, come back as a non-fatal per-source outcome to retry on the next cycle. So this change just propagates the auth signal instead of collapsing it into a generic failure. 

The orchestrator can then stop the current cycle on `Err(Auth)` — the credentials are bad for every source and a retry with the same token won't help (it needs a refresh), so there's no point attempting the remaining sources with the same expired credentials. Every other failure type stays per-source, so one source's transient blip or bad input doesn't halt the others. I'll add a test covering the auth-bubble-up path alongside the multi-batch test.

src/uploader/mod.rs
paabbs: Agreed the positional tuples are hard to read — `(PathBuf, u64, String)` and `(PathBuf, Vec<LogEvent>, u64, String)` don't self-document at call sites. Named structs like `FileUploadSuccess` / `FileUploadInput` would read better and make future field additions safer.

One note: for the current tuples every field is a distinct type, so the compiler does catch positional swaps today — the win here is readability and future-proofing rather than swap-safety (that risk would only appear once two same-typed fields are adjacent).

I'd prefer to land this as a focused readability follow-up rather than in this PR: it's a non-behavioral rename that touches the `file_events` input, which is produced in `main.rs` and consumed here — so it's cleaner as one atomic, compiler-checked pass over the full pipeline in a single change, and it keeps this PR scoped to the upload subsystem. I'll leave a `TODO` at `UploadResult` capturing the intent.

src/uploader/mod.rs
paabbs: Ack, I'll pre-build a `HashMap<&str, &ScannedFile>` keyed by `content_hash` and `get()` per file instead of the linear `find`, turning the per-iteration scan into O(n + m).

One detail: I'll build it first-wins (`entry().or_insert()`), so it preserves the current first-match selection in the rare case where two files share a `content_hash` (e.g. empty files, or files with an identical first line). That keeps it a pure performance change with no behavioral difference.

src/uploader/mod.rs
paabbs: Makes sense, will add `#[must_use]` to both `UploadResult` and `UploadOutcome`, matching the convention already used elsewhere in the crate. It guards against a caller silently dropping an outcome it needs to act on (advancing checkpoints / handling retries). No current caller ignores them, so it's purely a future-proofing lint.

src/uploader/mod.rs
paabbs: Makes sense,  will thread a single `now: u64` from the caller into both `advance_checkpoints` and `evict_stale_entries` instead of each calling the clock internally.

src/credentials/mod.rs
paabbs: Good point. On our current edition (2021) `set_var`/`remove_var` are still safe fns — they only become `unsafe` in edition 2024 — so there's no active deprecation warning today; this is forward-looking hygiene.

I'll go with your option 2 and inject the env lookup: a `resolve_thing_name_from(get_var: impl Fn(&str) -> Result<String, VarError>)` helper, with `resolve_thing_name()` delegating to it via `std::env::var`.

src/config/schema.rs
paabbs: This was from the format checks that we run before the PR:
```
cargo test 
cargo clippy --all-targets -- -D warnings 
cargo fmt --check 
cargo check --no-default-features --all-targets 
```

src/uploader/mod.rs
```diff
@@ -1,34 +1,259 @@
...
+
+    let file_map = store
+        .file_processing_info
+        .entry(log_group_key.to_string())
+        .or_default();
+
+    for (path, new_offset, content_hash) in succeeded {
+        let scanned = scanned_by_hash.get(content_hash.as_str()).copied();
+        let is_active = scanned.is_some_and(|f| f.is_active);
+        let file_len = std::fs::metadata(path).map(|m| m.len()).unwrap_or(0);
```
mihirpujara-amazon: Following the first-wins thread with @alter-mage (and your note that collisions happen for empty / identical-first-line files) — I think there's a *separate* provenance issue here, independent of which entry first-wins picks:

```rust
let scanned   = scanned_by_hash.get(content_hash.as_str()).copied(); // hash-matched entry
let is_active = scanned.is_some_and(|f| f.is_active);                 // from matched entry
let file_len  = std::fs::metadata(path).map(|m| m.len()).unwrap_or(0);// from succeeded `path`
```

`file_len` comes from `succeeded[].path`, but `is_active` (and `mtime` on L202) come from the hash-matched `ScannedFile`, which on a collision can be a different file than `path`. Since `compute_content_hash` hashes only the first line and `scan_directory` marks only the newest file active:

- older rotated file `A` and active file `B` share a first line → same hash;
- `succeeded` has both `(A, .., H)` and `(B, .., H)`;
- processing `(B, offB, H)`: first-wins returns `A`, so `is_active = false`, but `file_len = metadata(B)` — if `B` was fully read this cycle, `is_file_fully_uploaded(false, len(B), offB)` is true and the *active* file `B` gets marked complete and dropped from the checkpoint.

Am I reading that right? If so, resolving `file_len` / `is_active` / `mtime` all from the same `ScannedFile` (e.g. `metadata(&scanned.path)`) would keep them consistent. Non-blocking, but it's an edge the first-wins change doesn't cover.

paabbs: Makes sense. `advance_checkpoints` takes `is_active`/`mtime` from the hash-matched `ScannedFile` but `file_len` from `metadata(path)`. On a first-line hash collision (empty / identical-first-line files), a rotated file and the active file share a hash, so the active file's entry resolves `is_active` to the *other* file (`false`) while `file_len` is its own, and if it was fully read that cycle it's wrongly marked complete (and deleted under `deleteLogFileAfterCloudUpload`).

It slipped because `succeeded` is keyed by path while file state is resolved by hash — the two only diverge on a collision, which no test covers.

I'll fix it in this PR: resolve `is_active`/`file_len`/`mtime` for each file from a single consistent source so per-file state can't be mixed, plus a regression test for the collision case. Thanks for the detailed trace.

utsavvyas (approved):

mihirpujara-amazon (approved):

PR #253, opened 2026-07-08
feat(orchestrator): wire the periodic log upload pipeline

paabbs:
## Summary
Wires the upload orchestration on top of the merged upload-subsystem (#252): a fixed-delay loop driving scan → read → upload → checkpoint → disk-enforce, with ordered config-source resolution, thing-name handling, and graceful shutdown. This is the consumer that turns the subsystem library into a running component.

## What changed (`src/main.rs`, plus shared helpers in `src/uploader/`)
- **Fixed-delay loop** (`loop` + `tokio::select!`) over `process_all_sources` → `process_source` → `scan_and_filter_files` / `read_file_events` / `upload_and_advance_checkpoints`; sleeps the upload interval after each pass so a slow cycle never overlaps the next, and a shutdown request wakes it immediately.
- **Auth early-stop** via `enum CycleOutcome { Continue, StopForAuth }` — `Err(Auth)` stops the cycle; other outcomes continue to the next source.
- **Single `now` per cycle** threaded into `advance_checkpoints` / `evict_stale_entries`.
- **Ordered config-source resolver** (CLI flag → environment → default), source-agnostic; thing-name → `"unknown"` fallback with `error!`.

## Addressed in review
- **AWS SDK is a hard dependency** — removed the `aws-sdk` cargo feature, the no-sdk stub `main`, the per-function `#[cfg(feature)]` gates, and the blanket `allow(dead_code)`.
- **Shared `complete_file`** — the checkpoint-completion core (remove entry + monotonic timestamp advance) is now one helper called by both the `read_file_events` rotation branch and `advance_checkpoints`, so they can't drift. Completion triggers (`>=` vs `==`) and file deletes stay at each call site.
- **Removed unused hooks** — `CwLogsClient::recreate_client` (+ its config field) and `is_different_date`, both only reachable from tests.
- **Tests** — tempfile tests for `read_file_events` (completion/rotation, active-fully-read, partial, empty-skip, multiline) and `scan_and_filter_files`.

## How tested
- `cargo test` — 222 passed, 0 failed
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo fmt --check` — clean

## Notes for reviewers
- **Cross-boundary** — this PR also touches the merged subsystem lib (`src/uploader/mod.rs`, `src/uploader/cw_client.rs`) for the `complete_file` extraction, the feature removal, and the dead-hook cleanup.
- **Auth early-stop** — an efficiency choice to avoid hammering CloudWatch with known-bad credentials for the rest of a cycle. Stop-only by design: the SDK credential provider refreshes on expiry next cycle, so there is no explicit client refresh.
- **Thing-name resolution** — `resolve_thing_name` returns `Option<String>`; the caller applies the `"unknown"` fallback and logs `error!` when unset/empty.
- **Single `now` per cycle** — one timestamp threaded into `advance_checkpoints`/`evict_stale_entries` for deterministic tests. `read_file_events` intentionally keeps its own per-log-line fallback timestamp, distinct from the cycle bookkeeping `now`.
- **Disk enforcement** — runs only after a successful upload, deletes only already-uploaded, non-active files, oldest-first, and stops once under the limit.
- **Deferred (plain TODOs / follow-ups):** per-source `uploadIntervalSec` cadence, IPC `GetConfiguration` config source, system-source directory/regex runtime derivation, a `LogManager` struct refactor, and per-file granularity on batch failure. The disk-enforcement removal/delete path and the `process_*` orchestrators remain untested here (they need the replay-client seam).

## Base
Targets `rust/v2` (the upload-subsystem it builds on, #252, is merged).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,14 +6,41 @@
 //! Rust implementation of aws.greengrass.LogManager for GG Classic and GG Lite.
 //! Tails log files and EMF JSON files, uploads to CloudWatch Logs.
 
-use gg_log_manager::config::{load_config, validate_config};
+use gg_log_manager::config::{
+    derive_log_group_name, load_config, validate_config, ConfigError, LogLevel, LogManagerConfig,
+    LogSourceConfig,
+};
+use gg_log_manager::credentials::resolve_thing_name;
+use gg_log_manager::scanner::{
+    assemble_multiline, load_checkpoint, read_file_from_offset, recover_offsets, save_checkpoint,
+    scan_directory, trim_stale_on_load, CheckpointStore, LogEvent, ScannedFile,
+};
+use gg_log_manager::uploader::{
+    advance_checkpoints, complete_file, effective_interval_secs, evict_stale_entries,
+    format_log_stream_name, mtime_to_ms, upload_source_events, CwLogsClient, CwUploadError,
+    TTL_24H_MS,
+};
+use regex::Regex;
+use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
+use std::time::Instant;
 use tokio::sync::Notify;
-use tokio::time::{timeout, Duration};
-use tracing::{error, info};
+use tokio::time::Duration;
+use tracing::{error, info, warn};
 
 static SHUTDOWN_REQUESTED: AtomicBool = AtomicBool::new(false);
+
+/// Result of processing a source (or a whole cycle). `StopForAuth` signals a systemic auth
+/// failure that should halt the rest of the cycle so the SDK credential provider can refresh
+/// before the next pass; `Continue` means carry on normally. Self-documenting alternative to
+/// a bare `bool` whose `true`/`false` meaning would not be obvious at the call sites.
+#[must_use]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum CycleOutcome {
+    Continue,
+    StopForAuth,
+}
 
 #[cfg(not(tarpaulin_include))]
 #[tokio::main(flavor = "current_thread")]
@@ -21,9 +48,10 @@ async fn main() {
     tracing_subscriber::fmt::init();
 
     let args: Vec<String> = std::env::args().collect();
-    let config_arg = parse_args(&args);
+    let raw_config = resolve_raw_config(&args);
+    let work_dir = parse_arg(&args, "--work-dir").unwrap_or_else(|| ".".to_string());
 
-    let config = match load_config(&config_arg) {
+    let config = match load_config(&raw_config) {
         Ok(c) => c,
         Err(e) => {
             error!("Config load error: {e}");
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -6,24 +6,80 @@
 //! Rust implementation of aws.greengrass.LogManager for GG Classic and GG Lite.
 //! Tails log files and EMF JSON files, uploads to CloudWatch Logs.
 
-use gg_log_manager::config::{load_config, validate_config};
+// Feature-agnostic helpers are unused in the no-sdk stub build; allow dead_code only there.
```
utsavvyas: I dont understand why dead code allow is required ?

src/main.rs
```diff
@@ -6,24 +6,80 @@
...
+/// Result of processing a source (or a whole cycle). `StopForAuth` signals a systemic auth
+/// failure that should halt the rest of the cycle so the SDK credential provider can refresh
+/// before the next pass; `Continue` means carry on normally. Self-documenting alternative to
+/// a bare `bool` whose `true`/`false` meaning would not be obvious at the call sites.
+#[cfg(feature = "aws-sdk")]
+#[must_use]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum CycleOutcome {
+    Continue,
+    StopForAuth,
```
utsavvyas, line 42: Should we make it generic retryable and non retryable ?

src/main.rs
```diff
@@ -50,113 +106,571 @@ async fn main() {
...
-    info!("Step 4/4: Shutdown complete, exiting");
+    Some(scanned_files)
 }
 
-#[cfg(not(tarpaulin_include))]
-fn cancel_pending_uploads() {
-    // Cancel pending uploads - implementation in uploader track
+/// Read new content from each scanned file at its checkpointed offset; assemble multi-line entries.
+#[cfg(feature = "aws-sdk")]
+fn read_file_events(
```
mihirpujara-amazon, line 341: The core loop logic here (`read_file_events`) and `scan_and_filter_files` have no direct tests, and `process_all_sources` / `process_source` / `upload_and_advance_checkpoints` / `main` are all `#[cfg(not(tarpaulin_include))]` — so the new upload path is largely uncovered (only `parse_arg` / config resolution are tested).

`read_file_events` has the branching most worth pinning down:
- the `offset >= file_len && !is_active` completion path (checkpoint removal + monotonic timestamp advance + optional delete),
- multiline assembly, and
- the empty-events skip.

This is the same gap @utsavvyas raised on #252's `upload_source_events`, which you covered with `StaticReplayClient` tests — could we add tempfile-based tests here for at least the `read_file_events` completion/rotation branch? That's the path where a wrong `is_active`/offset decision can drop or duplicate data.

mihirpujara-amazon (commented):
Cross-PR cleanup: in #252, `CwLogsClient::recreate_client` (and the `config` field it uses) were kept with `#[allow(dead_code)]`, justified as the credential-refresh hook "for the upload orchestration." This PR *is* that orchestration, and it relies on the SDK credential provider's auto-refresh instead (per the reviewer notes) — it never calls `recreate_client`. So it stays dead, exercised only by `test_client_can_be_recreated`.

Suggest dropping `recreate_client` + the `config` field + that test now that the consumer exists and doesn't need them — or wiring an explicit refresh if we do want it. Leaning remove, since the SDK auto-refresh is the documented design here.

src/main.rs
```diff
@@ -6,24 +6,80 @@
...
+/// a bare `bool` whose `true`/`false` meaning would not be obvious at the call sites.
+#[cfg(feature = "aws-sdk")]
+#[must_use]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum CycleOutcome {
+    Continue,
+    StopForAuth,
+}
+
+/// No-sdk stub entry point: no CloudWatch client, so log an error and exit non-zero.
```
utsavvyas: Why do that happen ?

src/main.rs
```diff
@@ -50,113 +106,571 @@ async fn main() {
...
+            .map(str::to_string)
+            .unwrap_or_else(|| derive_log_group_name(name, None));
+        sources.push((&comp.source, log_group));
+    }
+
+    if let Some(ref sys) = config.logs_uploader_configuration.system_logs_configuration {
+        let enabled = sys.upload_to_cloud_watch;
+        // TODO: when logFileDirectoryPath/logFileRegex are unset on the system source,
+        // runtime-resolve them from the Nucleus logging config (via component SDK
+        // GetConfiguration); until then an unconfigured system source is skipped.
```
utsavvyas: This should be agnostic from where you get the information.

src/main.rs
```diff
@@ -50,113 +106,571 @@ async fn main() {
...
+                    }
+                    if source.delete_log_file_after_cloud_upload {
+                        if let Err(e) = std::fs::remove_file(path) {
+                            warn!(path = %path.display(), error = %e, "Failed to delete completed file");
+                        } else {
+                            info!(path = %path.display(), "Deleted completed file (rotation edge case)");
+                        }
+                    }
+                }
+                continue;
```
utsavvyas, line 397: This file-completion logic is duplicated. The fully-read / inactive-file completion here (remove the entry from the checkpoint, advance `lastFileProcessedTimeStamp` monotonically, and optionally delete the file) is the same completion `advance_checkpoints` performs via `is_file_fully_uploaded`. Two independent copies that each delete files and move timestamps will drift over time — and this is exactly the path where a divergence silently drops or double-deletes log data.

Suggest extracting a single `complete_file(...)` helper that both this rotation branch and `advance_checkpoints` call, so completion semantics (checkpoint removal + monotonic timestamp advance + conditional delete) live in one place.

src/main.rs
paabbs: True, the `allow(dead_code)` isn't actually needed — it was only masking that the no-sdk build is a non-functional stub. This component's sole purpose is uploading to CloudWatch, so the AWS SDK is a hard dependency: I'm making `aws-config`/`aws-sdk-cloudwatchlogs` non-optional and removing the `aws-sdk` feature entirely, which drops the stub `main`, the per-fn `#[cfg(feature="aws-sdk")]` gates, and this blanket allow

src/main.rs
paabbs: `CycleOutcome` is cycle-flow control — *should the whole upload cycle stop, or continue to the next source?* — a different axis from error retryability, which is already modeled by `CwUploadError { Retryable, Fatal, Auth }`. Only a systemic auth failure stops the cycle (`StopForAuth`); every other outcome (success, or a per-source retryable/fatal/exhausted error) returns `Continue`. Naming it "retryable/non-retryable" would conflate the two — a non-retryable *per-source* failure still continues. Happy to rename `StopForAuth` → `Stop` if you'd prefer a generic stop reason.

paabbs:
> Cross-PR cleanup: in #252, `CwLogsClient::recreate_client` (and the `config` field it uses) were kept with `#[allow(dead_code)]`, justified as the credential-refresh hook "for the upload orchestration." This PR _is_ that orchestration, and it relies on the SDK credential provider's auto-refresh instead (per the reviewer notes) — it never calls `recreate_client`. So it stays dead, exercised only by `test_client_can_be_recreated`.
> 
> Suggest dropping `recreate_client` + the `config` field + that test now that the consumer exists and doesn't need them — or wiring an explicit refresh if we do want it. Leaning remove, since the SDK auto-refresh is the documented design here.

True, recreate_client and the config field it uses were kept as a credential-refresh hook for the orchestration, but this orchestration relies on the SDK credential provider's automatic refresh (the cycle stops on an auth failure and the provider refreshes before the next pass), so it never calls recreate_client. Now that the consumer exists and doesn't need it, I'll remove recreate_client, the config field, and test_client_can_be_recreated in this PR.

src/main.rs
paabbs: The stub existed only so `cargo check --no-default-features` compiled, but nothing consumes that build and the binary is non-functional without the SDK. Since the SDK is a hard dependency here, I'm dropping the `aws-sdk` feature (the SDK deps become non-optional) and removing this stub along with it.

src/main.rs
paabbs: The config resolver (`resolve_raw_config`) is already source-agnostic — it returns the raw config regardless of origin and everything downstream just parses it. This is a wording issue in the TODO here; I'll reword it (and the sibling one) to not name a specific source, since the derived values flow through the same path no matter where they originate.

src/main.rs
paabbs: You're right — this duplicates the completion `advance_checkpoints` performs (remove the checkpoint entry + advance `lastFileProcessedTimeStamp` monotonically), and two independent copies would drift on exactly the path that can drop or double-delete data. I'll extract a single `complete_file(...)` that both this rotation branch and `advance_checkpoints` call, so the checkpoint-removal + monotonic-timestamp-advance core lives in one place.

I'll keep two things at the call sites rather than inside the helper, because they genuinely differ between the two paths: the **completion trigger** (this branch tolerates a past-EOF offset — `offset >= len` — while `advance_checkpoints` uses strict equality) and the **file delete** (here it's inline in the rotation branch, whereas `advance_checkpoints` returns the completed set and the delete is driven by the caller after the upload step). Folding the delete into the helper would change where/when those deletes happen, so I'll centralize only the checkpoint+timestamp core and leave the trigger and delete at each site. I'll also add a test that fails if the timestamp advance regresses. Happy to fully centralize as a part of this PR

src/main.rs
paabbs: Agreed — `read_file_events` and `scan_and_filter_files` take no CloudWatch client, so they're directly tempfile-testable. I've added tests for the completion/rotation branch (offset ≥ len && !active → checkpoint removal + monotonic timestamp advance + optional delete — the path where a wrong is_active/offset decision drops or duplicates data), the fully-read-but-active case, the empty-events skip, and multiline assembly, plus the scan filter (kept/dropped by last-processed timestamp). The `process_*` orchestrators need the replay-client seam, so I prioritized these two client-free helpers.

utsavvyas (approved):

mihirpujara-amazon (approved):

mihirpujara-amazon (approved):
Approving — both my comments are addressed in `cb33c59`:

- `recreate_client` (and the `config` field / `test_client_can_be_recreated`) removed, and the `aws-sdk` feature dropped so the stub `main` and per-fn gates go with it.
- `read_file_events` and `scan_and_filter_files` now have direct tempfile tests (completion/rotation, active-fully-read, partial read, empty-events, multiline, scan filter), and completion logic is centralized in the shared `complete_file()` — nice fix using `entry().or_default()` so the timestamp advance is unconditional when the group is absent.

Verified locally: 222 tests pass, `clippy --all-targets -- -D warnings` and `cargo fmt --check` clean. Also a nice cleanup dropping the other dead helper (`is_different_date`). LGTM.

PR #255, opened 2026-07-11
feat(config): IPC config source with live reload (Greengrass nucleus + nucleus lite)

paabbs:
## Summary
Adds an IPC-based configuration source so the component reads its own configuration from the Greengrass runtime — at startup (a fallback after CLI/env) and live via a configuration-update subscription, so config changes apply without a component restart. Works on both Greengrass nucleus and nucleus lite. Behind an opt-in `gg-ipc` feature so the default build stays portable.

## Why
The component's recipe carries no `Setenv`/`--config`, so on-device it would otherwise start on `"{}"` defaults on both runtimes. Reading its own deployment-merged config via IPC `GetConfiguration` gives it configuration at runtime, and subscribing to configuration updates lets it apply later changes live — without the service interruption of restarting the component on a config-only change. (Greengrass Classic does not restart a component for config-only changes; components consume such updates via the IPC configuration-update subscription.) CLI/env remain the primary sources for local/test.

## What changed
- `Cargo.toml`: optional `gg_sdk` (`aws-greengrass-component-sdk` `=1.0.4`) + a `gg-ipc` feature, not in `default`.
- `src/ipc_config.rs`: `connect_ipc()` (single connect) and `read_config()` (own-config `GetConfiguration`), behind `#[cfg(feature = "gg-ipc")]`; an empty module when the feature is off.
- `src/main.rs`:
  - `config_from_ipc()` wired as the last source in `resolve_raw_config` (`CLI → env → IPC → default`).
  - Config held in a shared cell the upload loop reads at the top of each cycle.
  - A configuration-update subscription whose callback signals the loop; the loop (which owns the connection) re-reads via `GetConfiguration`, validates, and swaps the cell — keeping the previous config on a bad update. No restart.

## Behavior notes
- Config changes take effect on the next cycle; a configuration-update notification also triggers an immediate processing cycle.
- If a component's `logGroupName` changes, its previous checkpoint entries are left behind and the new group starts from the beginning (re-uploading already-sent data for that group) — inherent to re-keying on a config change.
- The startup read is the first-load path on both runtimes; the subscription provides restart-free updates.

## How tested
- Default build (no feature): `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` — all green; `gg_sdk` is not compiled on the default path.
- Unit tests: resolution order (CLI/env/IPC/default, incl. empty-env fallthrough); `apply_config_update` swaps on a valid update and keeps the previous config on an invalid one.
- Not unit-tested (needs a device or mock socket): the subscription callback firing, and the end-to-end deploy → reload path.

## Notes for reviewers
- Opt-in feature: `gg_sdk`'s build depends on libclang + a C toolchain (edition 2024), so `gg-ipc` is kept out of `default` to keep `cargo build/test/clippy` portable. A dedicated `--features gg-ipc` CI job is recommended.
- Deferred (separate follow-ups): pruning checkpoint entries for components removed from config (pre-existing, metadata-only); resolving `systemLogsConfiguration` `logFileDirectoryPath`/`logFileRegex` from the runtime logging configuration; honoring per-source `uploadIntervalSec`.

## Base
Based directly on `rust/v2` (rebased over the merged orchestrator and checkpoint PRs). The rebase surfaced one interaction: the shutdown checkpoint flush now reads `deprecatedVersionSupport` from the live config cell (not the startup value), so a hot-reloaded setting governs the final write.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,6 +83,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "aws-greengrass-component-sdk"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d34fc6201e58c3f0b31726ad9ca9478f7333bb557e5dbd93e7df9f68a9b53df2"
+dependencies = [
+ "bindgen",
+ "cc",
+]
+
+[[package]]
 name = "aws-runtime"
 version = "1.7.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -482,6 +492,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "bindgen"
+version = "0.72.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "993776b509cfb49c750f11b8f07a46fa23e0a1386ffc01fb1e7d343efc387895"
+dependencies = [
+ "bitflags",
+ "cexpr",
+ "clang-sys",
+ "itertools",
+ "log",
+ "prettyplease",
+ "proc-macro2",
+ "quote",
+ "regex",
+ "rustc-hash",
+ "shlex",
+ "syn",
+]
+
+[[package]]
 name = "bitflags"
 version = "2.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -566,6 +596,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "cexpr"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6fac387a98bb7c37292057cffc56d62ecb629900026402633ae9160df93a8766"
+dependencies = [
+ "nom",
+]
+
+[[package]]
 name = "cfg-if"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


utsavvyas (approved):

paabbs:
Verification status for the `gg-ipc` path (it needs Linux — the SDK won't build on macOS):

- **Builds on Linux with `--features gg-ipc`**, including the static musl release target: `cargo zigbuild --release --features gg-ipc --target aarch64-unknown-linux-musl` → statically linked binary. Two cross-build notes: `bindgen` needs C headers for the target — set `BINDGEN_EXTRA_CLANG_ARGS_aarch64_unknown_linux_musl` to musl include paths (e.g. zig's bundled musl headers); native gnu-target builds need glibc >= 2.30 (`gettid`).
- **Ran the static binary on a Linux host without a Greengrass socket:** the IPC source fast-fails cleanly (no hang/panic) and config resolution falls through to CLI/env/default; full startup → shutdown cycle works, including the final checkpoint flush.
- **Not yet exercised:** the on-device path — `GetConfiguration` returning the deployment-merged config, and the update subscription firing. That needs a real device (or a mock socket); I'd like to validate it there before this leaves review. A dedicated `--features gg-ipc` CI job is recommended so the build stays verified going forward.

mihirpujara-amazon (approved):
Approving. IPC config source + live-reload design looks sound — per-cycle snapshot of `Arc<RwLock<Arc<Config>>>`, the subscription callback only signals a `Notify` (guarded by `catch_unwind` against unwinding through the extern "C" trampoline), and the IPC re-read runs on the loop thread that owns the handle (never the callback thread). Poison-tolerant locks, single `Sdk::init()`, keep-previous-config on a bad update, and `apply_config_update` is a clean tested seam.

Verified:
- Rebased onto current `rust/v2` (#254) — diff is now just the IPC layer.
- Default build green locally: 242 tests pass, `clippy --all-targets -- -D warnings` and `cargo fmt --check` clean.
- `--features gg-ipc` compiles for `aarch64-unknown-linux-musl` (confirmed by author via cargo-zigbuild; can't build on macOS since `gg_sdk` is epoll/Linux-only).

One caveat for follow-up (not blocking): the live IPC path is compile-verified but not runtime-verified — the deploy→reload path and the subscription callback firing still need a device/mock-socket test (as the PR notes call out). Worth a dedicated Linux `--features gg-ipc` CI job so the IPC code stays type-checked going forward. LGTM.

PR #257, opened 2026-07-21
feat(disk): enforce disk limits every scan cycle with safe reclamation

paabbs:
## Summary
Enforces each source's `diskSpaceLimit` on **every scan cycle** (not only after a successful upload), so already-uploaded files are reclaimed even during an upload outage. A source without its own limit can be bounded by an optional, customer-provided component-level `defaultDiskSpaceLimit`; when neither is set the source is not bounded. By default only fully-uploaded files are deleted; an opt-in `deleteUnuploadedFilesOnDiskPressure` (default `false`) additionally sheds the oldest un-uploaded files when a source is still over its limit.

## Why
Today the disk limit is only enforced after a successful upload, so during an upload outage or on idle cycles a source's directory can grow past its configured limit. On runtimes without a nucleus-managed log store (e.g. nucleus lite, where the component tails application-written `.emf`/log files that nothing else rotates), this component may be the only thing bounding those directories — an upload outage must not be able to fill the disk.

## Relationship to the Java component
The **default deletion policy is unchanged from `aws.greengrass.LogManager`**: only fully-uploaded, non-active files are ever deleted by default (`DiskSpaceManagementService` semantics — never un-uploaded data, never the active file). Three things intentionally go further, on runtime merits rather than parity: enforcement runs every cycle instead of only after a successful upload; operators can set a single component-level `defaultDiskSpaceLimit` that applies to sources without their own limit (unset everywhere = unbounded, as in the Java component); and the opt-in flag allows shedding un-uploaded files under pressure (off by default = Java behavior).

## What changed
- `src/disk/mod.rs`: new `enforce_disk_limit` — enumerates the directory itself, classifies each file (active / uploaded-safe / un-uploaded), deletes uploaded-safe oldest-first, then (opt-in) un-uploaded oldest-first, each logged at WARN. The active (newest) file is never deleted; every file is re-stat'd immediately before unlink. Replaces the list-driven `free_disk_space`.
- `src/main.rs`: enforcement runs unconditionally at the end of every `process_source` cycle — including cycles halted for credential refresh (it is a local filesystem operation). Cycles whose directory scan **failed** skip enforcement (no classification data). The upload cycle outcome is preserved.
- Files skipped by the scanner as hash-duplicates of a newer file are classified **un-uploaded** (they may not have been uploaded) and preserved by default; `scan_directory` now returns those skipped paths for this classification.
- `src/config/`: `deleteUnuploadedFilesOnDiskPressure` (accepts bool or string, like the other flags), the optional component-level `defaultDiskSpaceLimit`/`defaultDiskSpaceLimitUnit`, and `effective_disk_limit_bytes` resolving source limit → component default → unbounded (parse errors never invent a bound).
- README: disk space management section.

## Behavior notes
- **`defaultDiskSpaceLimit` applies per source** (each source without its own limit is bounded to that value individually). Leaving both unset keeps a source unbounded — matching the Java component.
- **Known edge (documented):** a directory whose files all share a first line (identical banner) dedups to one processed file; the skipped duplicates are preserved in default mode, so such a source is not bounded by default — it WARNs each cycle, and `deleteUnuploadedFilesOnDiskPressure` restores a hard bound. Recommended for high-volume or collision-prone producers.
- **Known edge (documented):** with `deleteLogFileAfterCloudUpload`, once the newer twin of a hash-duplicate is uploaded and deleted, the remaining duplicate is treated as completed on the next cycle and becomes reclaimable without having been uploaded itself — inherent to first-line content identity; strengthening that identity is tracked as follow-up work.
- Config changes to the limit or flag (including via live reload) take effect on the next cycle.
- The guarantee is best-effort: only files known to be uploaded are reclaimed by default, and enforcement pauses on cycles where the scan itself fails.

## How tested
- Unit suite for `enforce_disk_limit` (ordering, opt-in shedding, active protection, tracked/in-flight protection, pattern filter, undeletable files, missing dir); idle-cycle reclamation; the `defaultDiskSpaceLimit` resolution (component default applies when a source sets no limit, a source limit overrides it, neither set = unbounded, string and number forms); multi-cycle collision-workload tests (default preserves / flag sheds); the all-completed-scan case; scan-error skip; and integration tests driving the real scan→enforce path.
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` — all green.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -79,6 +79,9 @@ pub fn validate_config(config: &LogManagerConfig) -> Result<(), ConfigError> {
         warn_if_upload_interval_unwired(&sys.source, "systemLogsConfiguration");
         validate_log_source(&sys.source)?;
     }
+    if let Some(ref limit) = config.logs_uploader_configuration.default_disk_space_limit {
+        validate_disk_limit(limit)?;
+    }
     tracing::info!("Configuration validation passed");
     Ok(())
 }
@@ -130,6 +133,38 @@ pub fn parse_disk_space_limit(limit: Option<&str>) -> Result<Option<u64>, Config
             Ok(parsed)
         })
         .transpose()
+}
+
+/// Effective disk limit in bytes for a source, or `None` when the source is unbounded.
+///
+/// Resolution: the source's own `diskSpaceLimit`, else the component-level
+/// `defaultDiskSpaceLimit`, else `None` — only ever customer-provided values. A value that
+/// fails to parse is treated as unset (WARN); `validate_config` rejects invalid limits at
+/// load, so the `Err` arms are defensive.
+#[must_use]
+pub fn effective_disk_limit_bytes(
+    source: &LogSourceConfig,
+    uploader_config: &LogsUploaderConfig,
+) -> Option<u64> {
+    match parse_disk_space_limit(source.disk_space_limit.as_deref()) {
+        Ok(Some(limit)) => return Some(source.disk_space_limit_unit.to_bytes(limit)),
+        Ok(None) => {} // unset — fall through to the component default
+        Err(e) => {
+            tracing::warn!(error = %e, "Invalid source diskSpaceLimit; ignoring it and trying the component default");
+        }
+    }
+    match parse_disk_space_limit(uploader_config.default_disk_space_limit.as_deref()) {
+        Ok(Some(limit)) => Some(
+            uploader_config
+                .default_disk_space_limit_unit
+                .to_bytes(limit),
+        ),
+        Ok(None) => None, // neither set — the source is unbounded
+        Err(e) => {
+            tracing::warn!(error = %e, "Invalid defaultDiskSpaceLimit; source is left unbounded");
+            None
+        }
+    }
 }
 
 /// Derive log group name: /aws/greengrass/{componentType}/{region}/{componentName}
@@ -265,6 +300,79 @@ mod tests {
         assert!(parse_disk_space_limit(Some("abc")).is_err());
         assert_eq!(parse_disk_space_limit(None).unwrap(), None);
         assert_eq!(parse_disk_space_limit(Some("")).unwrap(), None);
+    }
+
+    #[test]
+    fn test_effective_disk_limit_none_when_neither_set() {
+        let comp: ComponentSourceConfig =
+            serde_json::from_str(r#"{"logFileDirectoryPath":"/tmp","logFileRegex":".*"}"#).unwrap();
+        let uploader = LogsUploaderConfig::default();
+        assert_eq!(effective_disk_limit_bytes(&comp.source, &uploader), None);
+    }
+
+    #[test]
+    fn test_effective_disk_limit_uses_configured_source_value() {
+        let comp: ComponentSourceConfig = serde_json::from_str(
+            r#"{"logFileDirectoryPath":"/tmp","logFileRegex":".*","diskSpaceLimit":"50","diskSpaceLimitUnit":"MB"}"#,
+        )
+        .unwrap();
+        let uploader = LogsUploaderConfig::default();
+        assert_eq!(
+            effective_disk_limit_bytes(&comp.source, &uploader),
+            Some(50 * 1024 * 1024)
+        );
+    }
+
+    #[test]
+    fn test_effective_disk_limit_falls_back_to_component_default() {
+        let comp: ComponentSourceConfig =
+            serde_json::from_str(r#"{"logFileDirectoryPath":"/tmp","logFileRegex":".*"}"#).unwrap();
+        let uploader = LogsUploaderConfig {
+            default_disk_space_limit: Some("10".into()),
+            default_disk_space_limit_unit: DiskSpaceLimitUnit::MB,
+            ..Default::default()
+        };
+        assert_eq!(
+            effective_disk_limit_bytes(&comp.source, &uploader),
+            Some(10 * 1024 * 1024)
+        );
+    }
+
+    #[test]
+    fn test_effective_disk_limit_source_overrides_component_default() {
+        let comp: ComponentSourceConfig = serde_json::from_str(
+            r#"{"logFileDirectoryPath":"/tmp","logFileRegex":".*","diskSpaceLimit":"5","diskSpaceLimitUnit":"MB"}"#,
+        )
+        .unwrap();
+        let uploader = LogsUploaderConfig {
+            default_disk_space_limit: Some("99".into()),
+            default_disk_space_limit_unit: DiskSpaceLimitUnit::GB,
+            ..Default::default()
+        };
+        // Source limit wins over the component default.
+        assert_eq!(
+            effective_disk_limit_bytes(&comp.source, &uploader),
+            Some(5 * 1024 * 1024)
+        );
+    }
+
+    #[test]
+    fn test_effective_disk_limit_bad_source_falls_to_component_default() {
+        // An unparseable source limit is treated as unset and falls through to the default;
+        // a bound is never invented from a bad value.
+        let comp: ComponentSourceConfig = serde_json::from_str(
+            r#"{"logFileDirectoryPath":"/tmp","logFileRegex":".*","diskSpaceLimit":"not-a-number"}"#,
+        )
+        .unwrap();
+        let uploader = LogsUploaderConfig {
+            default_disk_space_limit: Some("2".into()),
+            default_disk_space_limit_unit: DiskSpaceLimitUnit::MB,
+            ..Default::default()
+        };
+        assert_eq!(
+            effective_disk_limit_bytes(&comp.source, &uploader),
+            Some(2 * 1024 * 1024)
+        );
     }
 
     #[test]
--- a/src/config/schema.rs
+++ b/src/config/schema.rs
@@ -69,6 +69,11 @@ pub struct LogSourceConfig {
     pub disk_space_limit_unit: DiskSpaceLimitUnit,
     #[serde(default, deserialize_with = "deserialize_bool_or_string")]
     pub delete_log_file_after_cloud_upload: bool,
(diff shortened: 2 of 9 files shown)


src/config/mod.rs
```diff
@@ -132,6 +133,21 @@ pub fn parse_disk_space_limit(limit: Option<&str>) -> Result<Option<u64>, Config
...
+/// fail-safe when unset or unparseable. Enforcement therefore always has a bound; a
+/// source deployed without an explicit `diskSpaceLimit` is bounded by the default
+/// rather than left unbounded.
+#[must_use]
+pub fn effective_disk_limit_bytes(source: &LogSourceConfig) -> u64 {
+    match parse_disk_space_limit(source.disk_space_limit.as_deref()) {
+        Ok(Some(limit)) => source.disk_space_limit_unit.to_bytes(limit),
+        // Ok(None) = unset. Err is unreachable after validate_config (which rejects invalid
+        // limits); defaulting defensively rather than propagating.
+        _ => DEFAULT_DISK_SPACE_LIMIT_BYTES,
```
utsavvyas: We should not use the default disk limit if there are errors.

src/config/schema.rs
```diff
@@ -8,6 +8,11 @@ use std::collections::HashMap;
 
 pub const DEFAULT_UPLOAD_INTERVAL_SEC: u64 = 300;
 
+/// Fail-safe per-source disk limit applied ONLY when a source omits `diskSpaceLimit`.
+/// A conservative 256 MiB backstop so an unconfigured source cannot grow unbounded on
+/// constrained edge devices; an explicit per-source `diskSpaceLimit` always takes precedence.
+pub const DEFAULT_DISK_SPACE_LIMIT_BYTES: u64 = 256 * 1024 * 1024;
```
utsavvyas: Lets not use any default value for this.

src/disk/mod.rs
```diff
@@ -1,101 +1,201 @@
...
+            None
         }
     }
 }
 
-/// Free disk space by deleting oldest processed files until directory size ≤ limit.
-///
-/// Only deletes files in `processed_files` — caller must ensure these are safe to delete
-/// (fully uploaded, not actively written to). This function trusts the list it receives.
+// TODO(disk-pressure): Today this is a per-source QUOTA on each source's own directory;
```
utsavvyas: I dont think this is TODO. I would simply say it is not supported.

src/main.rs
```diff
@@ -395,56 +398,114 @@ async fn process_source(
...
+                &scanned_files,
+                file_events,
+                now,
+            )
+            .await;
+        }
     }
 
-    upload_and_advance_checkpoints(
+    // Enforce the disk limit LAST, every cycle, unconditionally — including on the StopForAuth
```
utsavvyas: It seems like we ae writing same/similar comments multiple places.

src/config/mod.rs
paabbs: Sure will address this as a part of the correction in the design choice of making the limit configurable. A `diskSpaceLimit` that fails to parse is now treated as unset (logged at `WARN`) and falls through to the component-level default if one is configured, otherwise the source is not bounded. No value is ever invented on an error. (`validate_config` still rejects invalid limits at load; this hardens the helper for consistency.)

src/config/schema.rs
paabbs: Will replace with an optional component-level `defaultDiskSpaceLimit`(`+defaultDiskSpaceLimitUnit`), so the fallback is customer-provided rather than baked in: a source's own `diskSpaceLimit` wins, else the component default applies, and when neither is set the source is not bounded (same as the Java component). One knob for operators who want every source bounded — useful on nucleus lite where nothing else rotates the tailed files — without the component choosing a number.

src/disk/mod.rs
paabbs: Ack, will reword

src/main.rs
paabbs: Makes sense will shorten them

utsavvyas (approved):

src/disk/mod.rs
```diff
@@ -1,101 +1,193 @@
...
+
+/// Enforce `limit_bytes` on `dir` by deleting the oldest eligible files.
 ///
-/// Returns paths of successfully deleted files.
-pub fn free_disk_space(
+/// Enumerates `dir` directly and classifies each matching file: newest mtime = **active**
+/// (never deleted); `mtime <= last_ts_ms && !is_tracked` = **uploaded-safe**; the rest =
+/// **un-uploaded**. Deletes uploaded-safe oldest-first until under limit, then (only if
+/// `allow_unuploaded`) un-uploaded oldest-first, WARN each. Files are re-stat'd before unlink.
+pub fn enforce_disk_limit(
```
alter-mage, line 104: Nit: `limit_bytes` and `last_ts_ms` are adjacent bare `u64`s — a call site that accidentally swaps them would compile silently. Have you considered a params struct or newtypes (`Bytes(u64)`, `EpochMs(u64)`) here, or is the single-call-site nature of this function enough mitigation?

src/disk/mod.rs
```diff
@@ -1,101 +1,193 @@
...
+#[derive(Debug, Default, PartialEq, Eq)]
+pub struct EnforceOutcome {
+    /// Fully-uploaded ("uploaded-safe") files that were deleted.
+    pub uploaded_deleted: Vec<PathBuf>,
+    /// Un-uploaded files deleted under disk pressure (only when `allow_unuploaded`).
+    pub unuploaded_deleted: Vec<PathBuf>,
+}
+
+/// A matching regular file found during directory enumeration.
+struct DiskEntry {
```
alter-mage, line 30: Minor: every other struct in this crate derives `Debug` — was omitting it here intentional, or just an oversight?

src/main.rs
```diff
@@ -383,68 +395,124 @@ async fn process_all_sources(
...
+                now,
+            )
+            .await;
+        }
     }
 
-    upload_and_advance_checkpoints(
+    // Enforce the disk limit last, every cycle — see `enforce_source_disk_limit` for the
+    // classification/ordering rationale (it runs on the StopForAuth path and idle cycles too).
+    enforce_source_disk_limit(
```
alter-mage: The return value is discarded here (intentionally, since the side effects happen inside). But `UploadResult` uses `#[must_use]` for a similar pattern. Should `EnforceOutcome` or `enforce_disk_limit` get `#[must_use]` for consistency, with an explicit `let _ =` at this call site to signal intent?

src/disk/mod.rs
```diff
@@ -1,101 +1,193 @@
...
+            Some(DiskEntry {
+                path: e.path(),
+                size: meta.len(),
+                mtime: meta.modified().unwrap_or(SystemTime::UNIX_EPOCH),
+            })
+        })
+        .collect()
+}
+
+fn mtime_ms(t: SystemTime) -> u64 {
```
alter-mage, line 65: This is the same logic as `uploader::mtime_to_ms`. I see the `disk` module is intentionally dependency-free (no `use crate::*`), so importing from `uploader` would introduce a coupling edge. Is a shared `util::epoch_ms` function on the roadmap, or are you keeping these intentionally duplicated for module isolation?

src/disk/mod.rs
paabbs: Fair callout — today it's mitigated rather than type-enforced: this function has a single production call site (`enforce_source_disk_limit`), which passes named locals `limit_bytes` and `last_ts` in order, and the unit tests use byte-scale limits vs epoch-ms timestamps, so a transposition flips `total <= limit_bytes` and fails the deletion assertions rather than passing silently.

 I'd rather not introduce the newtypes as a one-off here, because the crate passes bare `u64`s across every internal boundary today (bytes: `file_len`/`bytes_read`/`start_position`; epoch-ms: `mtime_ms`/`now`/`last_modified`; durations: `ttl_ms` vs `periodic_upload_interval_sec`) — a `Bytes` that exists only in `disk` would just add `.0` conversion churn at each module edge while the same hazard stays everywhere else. Ironically this call site is the best-mitigated instance in the crate; `is_file_fully_uploaded(is_active, file_len, bytes_read)` has two adjacent byte counts where a swap wouldn't be caught by magnitude, and the ms-vs-sec duration pair is the riskier confusion.

paabbs: Addd a TODO,   tracking a crate-wide follow-up that applies `Bytes`/`EpochMs` (and probably `Duration` for the sec/ms values) consistently. One constraint that makes it a deliberate change rather than a quick one: the epoch-ms fields live in the checkpoint structs that serialize to the on-disk format made backward-compatible in #254, so the newtypes need `#[serde(transparent)]` plus a wire-format round-trip test against the legacy checkpoint before touching those.

src/disk/mod.rs
paabbs: Not intentional — added `#[derive(Debug)]`. All three fields are `Debug`, so it's free and matches the module's other structs.

src/main.rs
paabbs: Good consistency catch — done. I put `#[must_use]` on the `EnforceOutcome` type (mirroring `UploadResult`) rather than on `enforce_disk_limit`: the inner function's result is already consumed for checkpoint cleanup inside `enforce_source_disk_limit`, so the value dropped at this call site is the wrapper's return. Marking the type flags that discard, and I've made it explicit with `let _ = …` here and in the tests that don't assert on the outcome.

src/disk/mod.rs
paabbs: Intentional, for module isolation — `disk` is deliberately dependency-free (only `std` + `regex`, no `crate::` paths), so it doesn't reach into `uploader` for a three-line `SystemTime → epoch-ms` conversion; importing `uploader::mtime_to_ms` would couple `disk` to `uploader` for a three-line conversion.

alter-mage (approved):

PR #258, opened 2026-07-22
fix: prevent unnecessary config.tlog writes from Integer/Long type mismatch

alter-mage:
## Problem

LogManager's `convertToMapOfObjects()` puts `long` fields (`startPosition`, `lastModifiedTime`, `lastAccessedTime`) into a `Map<String, Object>`. Java autoboxes these as `Long`.

When Nucleus deserializes config.tlog, Jackson's default `ObjectMapper` produces `Integer` for values that fit in 32 bits. The duplicate-write guard in `ConfigUtil.updateFromMapWhenChanged` uses `Objects.equals(node.getOnce(), value)` — which returns `false` for `Integer(4096)` vs `Long(4096)`, causing a tlog write on every processing cycle even when the value hasn't changed.

On devices with frequent reboots, this causes unbounded config.tlog growth.

## Fix

Added `asDeserializedNumber(long)` helper that returns `Integer` when the value fits without loss (`value == (int) value`), `Long` otherwise. This matches Jackson's default deserialization behavior, so `Objects.equals` succeeds when the numeric value is unchanged.

## Testing

Replicated with Docker using the actual Nucleus config store (Nucleus mainline without `USE_LONG_FOR_INTS`):

| Scenario | Before fix | After fix |
|---|---|---|
| 1 component, 5 reboots, unchanged values | +10 tlog lines | 0 |
| 36 components, 1 reboot | +108 tlog lines | +36 (lastAccessedTime only) |
| Large position (>Integer.MAX_VALUE), unchanged | +1 line | 0 |
| Value genuinely changes (4096 → 3000000000) | Write ✓ | Write ✓ |
| Large value persisted as Long, same value on reload | 0 | 0 |

### DIFF
--- a/src/main/java/com/aws/greengrass/logmanager/util/ConfigUtil.java
+++ b/src/main/java/com/aws/greengrass/logmanager/util/ConfigUtil.java
@@ -74,7 +74,7 @@ public final class ConfigUtil {
             try {
                 if (existingChild == null || existingChild instanceof Topic) {
                     Topic node = t.createLeafChild(key.toString());
-                    if (!Objects.equals(node.getOnce(), value)) {
+                    if (!valuesEqual(node.getOnce(), value)) {
                         node.withValueChecked(childMergeBehavior.getTimestampToUse(), value);
                     }
                 } else {
@@ -86,5 +86,20 @@ public final class ConfigUtil {
                 logger.error("Should never fail in updateChild", e);
             }
         }
+    }
+
+    /**
+     * Compares two values for equality, normalizing integral numeric types to long before comparison.
+     * This prevents false negatives from Integer/Long type mismatch when the config store
+     * deserializes numbers as Integer (Jackson default) but the caller provides Long (Java autoboxing of long).
+     * Floating-point numbers fall through to Objects.equals to avoid truncation.
+     */
+    private static boolean valuesEqual(Object existing, Object incoming) {
+        if (existing instanceof Number && incoming instanceof Number
+                && !(existing instanceof Double || existing instanceof Float
+                || incoming instanceof Double || incoming instanceof Float)) {
+            return ((Number) existing).longValue() == ((Number) incoming).longValue();
+        }
+        return Objects.equals(existing, incoming);
     }
 }


src/main/java/com/aws/greengrass/logmanager/LogManagerService.java
```diff
@@ -879,15 +879,29 @@ public Map<String, Object> convertToMapOfObjects() {
...
             currentProcessingFileInformationMap.put(PERSISTED_CURRENT_PROCESSING_FILE_LAST_ACCESSED_TIME,
-                    lastAccessedTime);
+                    asDeserializedNumber(lastAccessedTime));
             currentProcessingFileInformationMap.put(PERSISTED_CURRENT_PROCESSING_FILE_HASH, fileHash);
             return currentProcessingFileInformationMap;
         }
 
+        /**
+         * Returns the value as Integer if it fits without loss, Long otherwise.
+         * Matches Jackson's default deserialization behavior for config.tlog, ensuring
```
aws-kevinrickard: Rather than trying to match how the object was de-serialized, why not convert the de-serialized number to our known memory type (long)? This would remove the branch

src/main/java/com/aws/greengrass/logmanager/util/ConfigUtil.java
```diff
@@ -87,4 +87,16 @@ private static void updateChild(Topics t, CaseInsensitiveString key, Object valu
...
+
+    /**
+     * Compares two values for equality, normalizing numeric types to Long before comparison.
+     * This prevents false negatives from Integer/Long type mismatch when the config store
+     * deserializes numbers as Integer (Jackson default) but the caller provides Long (Java autoboxing of long).
+     */
+    private static boolean valuesEqual(Object existing, Object incoming) {
+        if (existing instanceof Number && incoming instanceof Number) {
+            return ((Number) existing).longValue() == ((Number) incoming).longValue();
+        }
```
aws-kevinrickard: nit: if a floating point number is returned from config (which shouldn't happen, and wouldn't make sense either), .longValue() would truncate and could cause a false equality result.

To be defensive we could handle that case too, but not blocking

aws-kevinrickard (approved):

saranyailla (approved):

